PR #305, opened 2014-11-27
Dockerfile for err

gbin:
This build an err image will all the dependencies for all the backends.
The only thing you have to do is to mount a volume from the host for the
data and put there a valid configuration.


gbin:
not that interesting, maybe in the contrib directory someday

PR #308, opened 2014-12-04
backend/irc: prefix '{nick}: ' for messages directed at requester

sijis:
Its a working override of the send() method. If this works, we should be able to do the same for the hipchat backend.

This should only apply to messages initiated in a groupchat with
'in_reply_to' specified in the send() method. Messages initiated
from a private message (chat) should not be affected.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,4 @@
+include COPYING
 include *.rst
 include errbot/builtins/*.plug
 include errbot/*.svg
--- a/errbot/backends/base.py
+++ b/errbot/backends/base.py
@@ -1149,9 +1149,20 @@ class Backend(object):
         bottom = self.bottom_of_help_message()
         return ''.join(filter(None, [top, description, usage, bottom]))
 
-    def send(self, user, text, in_reply_to=None, message_type='chat'):
+    def send(self, user, text, in_reply_to=None, message_type='chat', groupchat_nick_reply=False):
         """Sends a simple message to the specified user."""
-        mess = self.build_message(text)
+
+        nick_reply = self.bot_config.GROUPCHAT_NICK_PREFIXED
+
+        if (message_type == 'groupchat'
+                and in_reply_to
+                and nick_reply
+                and groupchat_nick_reply):
+            reply_text = self.groupchat_reply_format().format(in_reply_to.nick, text)
+        else:
+            reply_text = text
+
+        mess = self.build_message(reply_text)
         if hasattr(user, 'stripped'):
             mess.to = user.stripped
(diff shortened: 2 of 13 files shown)


sijis:
I've made a breakthrough! I forgot to add the parameter to botplugin.py. Anyhow, I should be able to clean this up in the next few days or so.

sijis:
Question:

If i have this line in irc.py
```python
'{}: {}'.format(in_reply_to.nick, text)
```
if i don't pass in_reply_to in a plugin, i get an error. How can i get the requester name without needing in_reply_to to be passed?

Sijis

sijis:
Hmmm. Maybe I just have to pass msg.nick instead of msg.from in the plugin.

With msg.frm I get #channel
Msg.nick I should see 'sijis'.

I'll try this again tomorrow.

sijis:
In this in the right direction?

In a plugin:
```python
    def ping(self, msg, args):
        """ this command responds 'pong' """
        self.send(msg.frm,
                  'pong (native)',
                  message_type=msg.type)
        self.send(msg.frm,
                  'pong! (no in_reply_to)',
                  message_type=msg.type,
                  groupchat_nick_reply=True)
        self.send(msg.frm,
                  'pong@ (all set)',
                  message_type=msg.type,
                  in_reply_to=msg,
                  groupchat_nick_reply=True)
```

The output:
```
22:43:15 @saviles | .ping                                                               │
22:43:15   errbot | pong (native)                                                       │
22:43:16   errbot | pong! (no in_reply_to)                                              │
22:43:17   errbot | saviles: pong@ (all set)    
```

In backends/irc.py (or any other backend)
```python
    def send(self, user, text, in_reply_to=None, message_type='chat', groupchat_nick_reply=False):
        """Sends a simple message to the specified user."""

        nick_reply = self.bot_config.GROUPCHAT_NICK_PREFIXED

        # nick_reply must be set in config and enabled via plugin
        # otherwise, response is sent normally
        if ( message_type == 'groupchat'
             and in_reply_to
             and nick_reply
             and groupchat_nick_reply ):
            reply_text = '{}: {}'.format(in_reply_to.nick, text)
        else:
            reply_text = text

        super().send(user, reply_text, in_reply_to, message_type, groupchat_nick_reply)
```

gbin:
LTGM

sijis:
I added the bulk of the logic to base.py, as I was implementing for each backend I noticed i was just reusing the same code method. I thought this would be a cleaner/simpler solution to the code dupe and then allow each backed to define a 'format' for how groupchat messages are sent.

gbin:
thx

PR #320, opened 2015-01-06
Support for case-insensitive file system (e.g. OS X)

gugu:
+ fixed some bug

### DIFF
--- a/errbot/__init__.py
+++ b/errbot/__init__.py
@@ -6,9 +6,9 @@ PY2 = not PY3
 __all__ = ['BotPlugin', 'webhook', 'webroute', 'webview']
 
 # Import some stuff here so that it's easier to access them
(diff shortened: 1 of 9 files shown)


gbin:
relative imports are nice especially if they fix something O o . Can you fix the PEP8 brackage so we can merge it ? thx !

gugu:
Yes, you have errBot.py in errbot directory.

So when code `from errbot import PY2` runs in errbot/decorators.py:

* in linux, it imports PY2 from errbot/__init__.py
* in MacOS it tries to import PY2 from errbot/errBot.py

Sure, will fix PEP8 tomorrow morning

gbin:
Ha ok, thx for the explanation !

gugu:
fixed :)

PR #333, opened 2015-03-08
Autoinstaller for missing plugins dependencies

gbin:
This adds an option so err can autoinstall the missing requirements from plugins requirements.txt

It also makes the preloaded (either builtins and extra) plugins loading more consistent with the rest (it will go through the subdirectory and properly detect all the plugin roots).

### DIFF
--- a/errbot/config-template.py
+++ b/errbot/config-template.py
@@ -31,7 +31,16 @@ BOT_DATA_DIR = '/var/lib/err'
 # locally before publishing it. Note that you can specify only a single
 # directory, however you are free to create subdirectories with multiple
 # plugins inside this directory.
-BOT_EXTRA_PLUGIN_DIR = None 
+BOT_EXTRA_PLUGIN_DIR = None
+
+# If True, when you err wants to start a plugin and a dependency from
+# requirements.txt of the plugin is missing, it will automatically invoke
+# pip install --user for the missing dependencies.
+# It will detect if the bot is running in a virtualenv and if yes, it
+# will install it in the virtualenv instead of on the user.
+#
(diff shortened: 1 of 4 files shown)


gbin:
@zoni Does it look like ok for you ?

errbot/config-template.py
```diff
@@ -31,7 +31,16 @@
...
+BOT_EXTRA_PLUGIN_DIR = None
+
+# If True, when you err wants to start a plugin and a dependency from
+# requirements.txt of the plugin is missing, it will automatically invoke
+# pip install --user for the missing dependencies.
+# It will detect if the bot is running in a virtualenv and if yes, it
+# will install it in the virtualenv instead of on the user.
+#
+# Note: be sure to give the proper autorizations to the user running the bot.
+AUTOINSTALL_DEPS = True
```
zoni: I suggest changing this to the following text:

```
# Should plugin dependencies be installed automatically? If this is true
# then Err will use pip to install any missing dependencies automatically.
#
# If you have installed Err in a virtualenv, this will run the equivalent
# of `pip install`. If no virtualenv is detected, the equivalent of `pip
# install --user` is used to ensure the package(s) is/are only installed for
# the user running Err.
```

zoni: Maybe also put a `#` in front of `AUTOINSTALL_DEPS` to comment it out, so that people upgrading can continue using their old `config.py` and do not have to add *AUTOINSTALL_DEPS* before their bot starts again (Err checks for missing configuration items and refuses to start if any are missing).

zoni:
Code looks good to me. :+1:

PR #344, opened 2015-05-15
Hipchat keepalive fixes

zoni:
This PR fixes a long-standing bug where `XMPP_KEEPALIVE_INTERVAL` wasn't being honored, and, if the user hasn't specified it in their config, uses a better default for HipChat (because the SleekXMPP default of 3600 works fine just about anywhere except for on HipChat).

### DIFF
--- a/errbot/backends/hipchat.py
+++ b/errbot/backends/hipchat.py
@@ -316,6 +316,12 @@ class HipchatBackend(XMPPBackend):
         super(HipchatBackend, self).__init__(config)
 
     def create_connection(self):
+        # HipChat connections time out with the default keepalive interval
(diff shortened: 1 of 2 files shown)


zoni:
Tests are [passing](https://travis-ci.org/gbin/err/builds/62701880) fine, despite what GitHub is claiming.

zoni:
@gbin can we merge this and make a 2.2.1 with it or something? :) Without this the HipChat back-end is unusable because it'll keep getting disconnected by HipChat every 5 minutes or so.

gbin:
argh ok sorry gonna do that

gbin:
thx

PR #347, opened 2015-05-16
chore: Merge branch arg_botcmd from TomNeyland/err

TomNeyland:
This pull request:
* Adds a new decorator for creating argparse style bot commands, ```arg_botcmd```
* Adds tests for ```arg_botcmd```
* Adds documentation for ```arg_botcmd```


TomNeyland:
Here is the 30 second look at what this PR lets you do:

```py
@arg_botcmd('first_name', type=str)
@arg_botcmd('--last-name', dest='last_name', type='str', default='')
@arg_botcmd('--favorite', dest='favorite_number', type=int, default=42)
def hello(self, mess, first_name=None, last_name=None, favorite_number=None):
    # if you send it !hello Err --last-name Bot
    # first_name will be 'Err'
    # last_name will be 'Bot'
    # favorite_number will be 42
    return "Hello %s %s your favorite number is %s" % (first_name, last_name, favorite_number)

```

zoni:
Regarding some discussion on gitter:

While the code with the nested decorator looks a bit confusing at first, I actually find this a very elegant solution. From an end-user perspective, I find the using one-decorator-per-argument very nice looking, definitely nicer than passing some kind of dict to a unique decorator, and it has the extra benefit of mapping over exactly to how people are used to using `parser.add_argument()` in CLI scripts.

I'm personally a big proponent of making APIs as human friendly as possible, even if it makes the underlying code a bit more complex sometimes, so I vote :+1: on this implementation.

TomNeyland:
Interestingly enough this commit built correctly for my repo's travis runner, I wonder if it was a one-off issue.

gbin:
I guess I like it on a second review, I gonna merge it, thx a lot for this interesting PR !

PR #351, opened 2015-05-18
Further slack improvements

zoni:
This further improves the Slack back-end. Most notably, it prevents Err from crashing entirely on certain exceptions, and, biggest of all, it completely changes how Identifiers are constructed (see b622005ed17068d0556725253f77e643d9f1b4f0).

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -8,6 +8,10 @@ from errbot.errBot import ErrBot
 from errbot.utils import deprecated
 
 try:
+    from functools import lru_cache
+except ImportError:
+    from backports.functools_lru_cache import lru_cache
+try:
     from slackclient import SlackClient
 except ImportError:
     logging.exception("Could not start the Slack back-end")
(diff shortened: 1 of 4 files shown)


zoni:
To clarify further, the `node` part of an Identifier is now always the username of the person sending/receiving a message, the `domain` part is always the name of the channel/group.

errbot/backends/slack.py
```diff
@@ -90,22 +112,58 @@ def _handle_slack_event(self, event):
...
+                to_humanreadable = mess.to.domain
+                to_id = self.channelname_to_channelid(to_humanreadable)
+            else:
+                to_humanreadable = mess.to.node
+                api_data = api_resp(
+                    self.sc.api_call(
+                        'im.open',
+                        user=self.username_to_userid(to_humanreadable)
+                    )
+                )
```
zoni: There's a TODO here. Since we never close the IM sessions, we should cache the result of this API call so we only have to do it once for each user, thereby saving a whole bunch of API calls (that count towards our rate limits).

PR #352, opened 2015-05-20
added a server parameter for XMPP

gbin:
It is needed if you want to be able to force an XMPP server to connect to instead of relying on SRV records or direct resolution.

### DIFF
--- a/errbot/backends/xmpp.py
+++ b/errbot/backends/xmpp.py
@@ -261,10 +261,11 @@ class XMPPMUCOccupant(MUCOccupant):
 
 
 class XMPPConnection(object):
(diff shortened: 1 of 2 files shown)


gbin:
@zoni done

zoni:
Looks solid to me!

errbot/backends/xmpp.py
```diff
@@ -368,6 +372,7 @@ def __init__(self, config):
 
         self.jid = Identifier(identity['username'])
         self.password = identity['password']
+        self.server = identity.get(['server'], None)
```
zoni: I overlooked this when reviewing & merging :( This should be `self.server = identity.get('server', None)` instead. Going to push a fix to master.

PR #354, opened 2015-05-22
Added modules as logger names.

gbin:
This allow a clearer view in the log and a better fine grained filtering
of the verbosity.


zoni:
A number of files didn't have `log` defined, but I believe I fixed them all so I'm going to go ahead and merge this.

PR #367, opened 2015-05-28
Completely revamped backup feature

gbin:
Now a single chat command creates an online backup:
!backup

Then you can store safely the backup.py script created.

When you lost your bot, you can restore it from an empty instance with: 
   err -r [FILE]
or
   err --restore [FILE]

It will reinstall all the previously installed plugin, restore their settings then restore their shelf data.


zoni:
Going to take a look at this some time this weekend

sijis:
This is great!

gbin:
I actually miss a unit test on this, maybe just a simple cycle of dump/restore

PR #376, opened 2015-06-01
[slack] Check username against BOT_ADMINS/ACLs

### LINKED ISSUE #374: is_admin and security/access control not working with slack backend
ive tried all combinations of user id, dm id, resource, node, etc and canot get admin access or auth to work with master

zoni:
Fixes #374.

### DIFF
--- a/errbot/backends/base.py
+++ b/errbot/backends/base.py
@@ -1032,8 +1032,11 @@ class Backend(object):
 
         Raises ACLViolation() if the command may not be executed in the given context
         """
-        usr = str(get_jid_from_message(mess))
+        usr = str(self.get_jid_from_message(mess))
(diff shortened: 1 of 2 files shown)


Cobra800089:
This python 2.7 build failed because running setup.py more than once injects a "u" character in front of any single quote or double quote in multiple files. I recommend you pull down a fresh copy of the branch and rerun setup.pu

gbin:
The recent identifier refactoring should fix this: it is always checked against "frm.person" so the most precise identifier pointing to a person (not a nick, not a name)
For Slack this should be the U023BECGF form.
The new !whoami can give you yours easily.

If it still doesn't work please reopen a bug/PR.

PR #388, opened 2015-07-03
Removes campfire support.

gbin:
The backend is not compatible with py3, we have seen no activity from
campfire users (either bugs / support etc..).

We can assume it is unused.

If somebody is really using it, please ping us !


gbin:
As an alternative we can put it in another repo with the pending backends as plugins feature that will land in 2.3

gbin:
We will simply put it outside of the main tree to give the opportunity for a maintainer to take it over if it is still useful.

PR #389, opened 2015-07-04
Ported backends as plugins.

gbin:
This allows third party backends to be run by err.

backends have the same .plug as a normal err plugin.
The backend has to be a subclass of ErrBot.

The port had a ripple effect across the test codebase that has to be cleaned up as we were doing dodgy things like globals (yapsy loads plugins in a virtual modules).

err can now be started with a backend name with -b or --backend.

Another new command line parameter -l to list the available backends
will be added in another PR.

### DIFF
--- a/errbot/backend_manager.py
+++ b/errbot/backend_manager.py
@@ -1,0 +1,75 @@
+import os
+import logging
+import sys
+
+from yapsy.PluginManager import PluginManager
+from yapsy.PluginFileLocator import PluginFileLocator, PluginFileAnalyzerWithInfoFile
+from errbot.errBot import ErrBot
+from .utils import find_roots_with_extra
+
+CORE_BACKENDS = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'backends')
+log = logging.getLogger(__name__)
+
+
+class SpecificBackendLocator(PluginFileAnalyzerWithInfoFile):
+    def __init__(self, name_to_find):
+        super().__init__('SpecificBackendLocator', 'plug')
+        self._name_to_find = name_to_find
+
+    def getInfosDictFromPlugin(self, dirpath, filename):
+        plugin_info_dict, config_parser = super().getInfosDictFromPlugin(dirpath, filename)
+        if plugin_info_dict['name'] != self._name_to_find:
+            # reject
+            return None, config_parser
+        return plugin_info_dict, config_parser
+
+
+class BackendManager(PluginManager):
+    """ BackendManager is a customized plugin manager to enumerate backends
+        and load only one.
+    """
+    def __init__(self, config):
+        self._config = config
+        # set a locator that gets every possible backends
+        self._locator = PluginFileLocator(analyzers=[PluginFileAnalyzerWithInfoFile('AllBackendLocator', 'plug')])
+        super().__init__(plugin_locator=self._locator)
+        self.setCategoriesFilter({'backend': ErrBot})
+        if hasattr(config, 'BOT_EXTRA_BACKEND_DIR'):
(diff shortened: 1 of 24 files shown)


errbot/backends/test.py
```diff
@@ -11,19 +11,6 @@
...
-tempdir = mkdtemp()
-config.BOT_DATA_DIR = tempdir
-config.BOT_LOG_FILE = tempdir + sep + 'log.txt'
-config.BOT_EXTRA_PLUGIN_DIR = []
-config.BOT_LOG_LEVEL = logging.DEBUG
-config.CHATROOM_PRESENCE = ('testroom',)  # we are testing with simple identfiers
-config.BOT_IDENTITY = {'username': 'err'}  # we are testing with simple identfiers
-
 # Errbot machinery must not be imported before this point
 # because of the import hackery above.
```
zoni: It looks like this comment is no longer true now, so it can go away and the import statements below can go up with the other import statements (so then we can also remove the `noqa` statements behind them)

errbot/backends/test.py
```diff
@@ -359,85 +345,10 @@ def tearDown(self):
...
+        self.assertIn(response, self.bot.pop_message(), timeout)
 
     def assertCommandFound(self, command, timeout=5):
         """Assert the given command does not exist"""
-        push_message(command)
-        self.assertNotIn('not found', popMessage(), timeout)
-
-
-@pytest.fixture
-def testbot(request):
```
zoni: It looks like this is removed here but not re-introduced anywhere else, but I would like to keep it so plugin authors can continue to write tests like this:

```python
        # myplugin_test.py
        from errbot.backends.test import testbot, push_message, pop_message

        def test_about(testbot):
            push_message('!about')
            assert "Err version" in pop_message()
```

errbot/utils.py
```diff
@@ -285,3 +286,24 @@ def repeatfunc(func, times=None, *args):  # from the itertools receipes
...
+    """Collects all the paths from path recursively that contains files of type file_sig"""
+    roots = set()  # you can have several .plug per directory.
+    for root, dirnames, filenames in os.walk(path):
+        for filename in fnmatch.filter(filenames, file_sig):
+            roots.add(os.path.dirname(os.path.join(root, filename)))
+    return roots
+
+
+def find_roots_with_extra(base, extra, file_sig='*.plug'):
+    # adds the extra plugin dir from the setup for developers convenience
```
zoni: This function is missing a proper docstring

errbot/backend_manager.py
```diff
@@ -0,0 +1,75 @@
...
+from yapsy.PluginManager import PluginManager
+from yapsy.PluginFileLocator import PluginFileLocator, PluginFileAnalyzerWithInfoFile
+from errbot.errBot import ErrBot
+from .utils import find_roots_with_extra
+
+CORE_BACKENDS = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'backends')
+log = logging.getLogger(__name__)
+
+
+class SpecificBackendLocator(PluginFileAnalyzerWithInfoFile):
```
zoni: This class could use a docstring

errbot/backends/test.py
gbin: indeed this is way cleaner now.

errbot/utils.py
gbin: done

errbot/backend_manager.py
gbin: done

errbot/backends/test.py
gbin: done, with the only nuance that the global push_message was a no-go.
People can use testbot.push_message now which I think is fine and kind of logical.

PR #391, opened 2015-07-10
Fixes related to groupchat replies

sijis:
Fixed issue where replies in groupchat returned the userid instead of the username in slackbackend.
It also uses the new muc identifiers.

I also update the variables to be more descriptive to each backend.

Sijis


gbin:
I went ahead I patched it in there : 2b2531b242f7e85b138c38de623d7b86e5f48834

Thanks for the fix !

PR #394, opened 2015-07-12
Slack: Add new users to the local user cache.

gbin:


zoni:
I think we should fix this upstream by letting SlackClient [process it here](https://github.com/slackhq/python-slackclient/blob/master/slackclient/_client.py#L42-L49), instead of working around it ourselves. That way it just works by default for anyone using that library.

zoni:
It might also be good to handle the [user_change](https://api.slack.com/events/user_change) event at the same time as well.

gbin:
@zoni completely agree, this is weird that the binding doesn't do it by itself. I gonna put a bug on their bindings.

gbin:
@zoni reported it https://github.com/slackhq/python-slackclient/issues/33

gbin:
No reaction from them, I'll merge this one for now and revisit once they improved their implementation.

PR #398, opened 2015-07-12
Remove the XMPP warning from the doc.

gbin:
We are past that now. :)

### DIFF
--- a/docs/user_guide/plugin_development/intro.rst
+++ b/docs/user_guide/plugin_development/intro.rst
@@ -6,40 +6,22 @@ necessary requirements installed and give you an idea of which
 knowledge you should already possess in order to follow along
 without difficulty.
 
-XMPP versus IRC
----------------
-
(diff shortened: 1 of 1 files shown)


docs/user_guide/plugin_development/intro.rst
```diff
@@ -6,40 +6,22 @@ necessary requirements installed and give you an idea of which
...
 
-You should also have some basic familiarity with the `Extensible
-Messaging and Presence Protocol (XMPP)
-<http://xmpp.org/about-xmpp/>`_. Err provides high-level APIs for
-most tasks so if you know what Jabber IDs are, understand how
-resources and priorities relate to each other and have an
-understanding of the concept of Multi User Chatrooms (MUC), you
-should not have any difficulties.
+Recommended to implement the tests of you plugin:
+`Python <http://pytest.org/>`_.
```
zoni: We have an entire section on testing plugins, so I think we should just remove these two lines entirely and not mention it at this stage at all.

Other than that though, yes, :+1:

PR #399, opened 2015-07-12
[Slack] Match ACLs against username rather than cryptic userid

zoni:
This will allow people to enter people's usernames for ACL entries
rather than having to look up their user IDs (which isn't easy to do).

### DIFF
--- a/errbot/backends/__init__.py
+++ b/errbot/backends/__init__.py
@@ -58,10 +58,22 @@ class SimpleIdentifier(DeprecationBridgeIdentifier):
         Returns None is unspecified"""
         return self._fullname
 
+    @property
+    def aclattr(self):
+        """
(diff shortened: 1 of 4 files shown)


gbin:
Interesting, maybe I would do that differently.

Maybe having another conventional entry in identifiers ?
like security_identifier ? [or another name]

you would just need to put:
security_identifier = person  # for all the identities except slack
security_identifier = nick  # for slack

We need to be really sure that 2 people cannot have the same nick or cannot change nick btw.

zoni:
Re-implemented using property on `Identifier`. What do you think of this?

gbin:
I would not make anything inherit from simpleIdentifier, just implement it on it but let the other identifiers inherit from the deprecation bridge or object

Issue #365, opened 2015-05-27
Group core commands and move them into dedicated plugins


zoni:
Some of Err's core commands and related code are already in their own plugins, such as the version check, webserver and MUC management.

After some discussion between me, @gbin and @sijis on gitter we agreed we should move all of Err's core commands into such plugins.

At the moment, the following commands are all implemented directly in `errBot.py`:

```
status
status_load
status_gc
status_plugins
echo
uptime
export_configs
import_configs
zap_configs
repos_export
restart
blacklist
unblacklist
load
unload
reload
repos_install
repos_uninstall
repos
help
history
about
apropos
repos_update
config
log_tail
```

My proposal is to adjust these as follows:

```
core_plugins/
      |
      +-- core.py
      |      - !status
      |      - !status load
      |      - !status gc
      |      - !status plugins
      |      - !restart
      |      - !uptime
      +-- utils.py
      |      - !echo
      |      - !log tail
      |      - !history
      +-- plugins.py
      |      - !repos install
      |      - !repos uninstall
      |      - !repos
      |      - !repos export
      |      - !plugin load (renamed from !load)
      |      - !plugin unload (renamed from !unload)
      |      - !plugin reload (renamed from !reload)
      |      - !plugin blacklist (renamed from !blacklist)
      |      - !plugin unblacklist (renamed from !unblacklist)
      |      - !plugin config (renamed from !config)
      |      - !plugin zap_configs (renamed from zap_configs)
      +-- help.py
             - !about
             - !help
             - !apropos
```

The current plugins webserver and version check don't expose any commands, `ChatRoom` already has all commands prefixed with `!room` so I think these are fine as-is for now.

I don't know if creating a separate `core` and `utils` is necessary, but my reasoning is that we may get some more utility-like commands in the future, that don't really belong in core because they're useful/nice to have. For example I have a private plugin with a `!whoami` command that displays the JID of the person issuing it (useful for setting up ACL entries and debugging some MUC issues) and a `!broadcast` which broadcasts a message to all joined MUCs, both of which would go well in `utils`.

sijis:
I agree with @gbin comment regarding moving !about to help.py.
:+1:  to ```utils```
everything else seems fine to me.

gbin:
I continue to brainstorm about this ...

I would rename core.py to admin.py (to administrate the bot basically and we are already in core plugins)
!history moved to utils (it is an helper).
rename repos.py to plugins.py (because it manages repos and plugins)

[...]

PR #400, opened 2015-07-14
moved help to its own plugin.

gbin:
This is the first part of a refactoring that will move out all the
bot commands out of the base.backend-errbot-backend hierarchy.


gbin:
More context here: #365

PR #408, opened 2015-07-22
Markdown

gbin:
Historically Err could only do pure unicode text or xhtml-im. Templates were added and the process was like this:
1. Err applies the html-im template with the result of a command
2. Ellrr tries to remake a text version of it by parsing the output of the command in html
3. A message gets created with body (the text version) and html (the html created from the template) 

As we added more and more backend, this become more and more pointless as no backend except xmpp support html-im.

This PR fixed that by first ripping out this old html logic completely.

Now the expected output of any command is *markdown extra*. Jinja2 templates are still used but their extension is .md instead of .html.

Err exposes some helpers to the backends to be able to make the most of the formatting from the markdown version.

Here is the test document rendered under different backends:

### Console
ansi

### Slack
slack

### Hipchat
hipchat

### Gitter
gitter


zoni:
I'm just missing one thing, how are links formatted on the various back-ends?

zoni:
Also, before we merge this, we should squash all the commits down to one I think, keep a cleaner history.

gbin:
links are represented in the intermediate tree as A, I haven't tried they might be gobbled up in the process as it is.

gbin:
Just merged the branch in a squashed commit.

gbin:
@zoni I fixed on head the links for text rendering by putting them in parenthesis.

PR #413, opened 2015-07-24
Fixes 2 bugs in xmpp backend

rroemhild:
This PR fixes two issues where the bot
- can not join xmpp rooms
- can not identify the user on status change


gbin:
Thanks for helping on this !

Issue #422, opened 2015-07-25
In failing test, py.test produces thousands of lines of debug output from yapsy and others


RobSpectre:
When writing a failing test, thousands of DEBUG statements are outputted from yapsy and others.

Steps to reproduce - create simple plugin file:

```python
from errbot import BotPlugin
from errbot import botcmd


class TestPlugin(BotPlugin):
    min_err_version = '1.6.0'
    max_err_version = '2.3.0'

    @botcmd
    def lookup(self, message, args):
        return message
```

Create test:

```python
from errbot.backends.test import testbot
from errbot import plugin_manager


class TestTwilioLookup(object):
    extra_plugin_dir = '.'

    def test_lookup(self, testbot):
        self.plugin = plugin_manager.get_plugin_obj_by_name('TestPlugin')
        expected = "test"
        self.bot.push_message("!lookup")

        self.assertTrue(expected in self.bot.pop_message(),
                        "Did not get message 'test', instead: "
                        "{0}".format(self.bot.pop_message()))
```

Run `py.test`

zoni:
This should be fixed by #424 now :smiley:. Let us know if it didn't resolve the issue please.

PR #424, opened 2015-07-25
Initialize correctly the logs for testbot

gbin:
This was the bare logging with everything at DEBUG and was noisy.

This is almost a copy-paste from tests/__init__.py where we had the same
issue.

### DIFF
--- a/errbot/backends/test.py
+++ b/errbot/backends/test.py
@@ -413,7 +413,20 @@ def testbot(request):
     def on_finish():
         bot.stop()
 
+    #  setup the logging to something digestable.
+    import logging
+
+    logger = logging.getLogger('')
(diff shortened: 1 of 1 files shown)


errbot/backends/test.py
```diff
@@ -413,7 +413,20 @@ def test_about(self, testbot):
     def on_finish():
         bot.stop()
 
+    #  setup the logging to something digestable.
+    import logging
```
zoni: `logging` is already imported at the top of the file, so this import should be removed here I think.

gbin:
this fixes #422

errbot/backends/test.py
gbin: done

PR #432, opened 2015-08-02
Added err-coffeetime to known repos

Ecno92:
I just created a small plugin for fun. When ``!cofeetime`` is called in a group chat which the Err Bot is part of then Err will pick a random person who should bring the coffee.


gbin:
thx :)

PR #456, opened 2015-08-22
Adds command filters and ported ACLs to it

gbin:
This also includes is_admin.

The main goal is to be able to inject new security schemes to the bot
like 2 factor auth with OTP for example.

### DIFF
--- a/errbot/errBot.py
+++ b/errbot/errBot.py
@@ -77,6 +77,7 @@ class ErrBot(Backend, BotPluginManager):
             log.debug('created the thread pool' + str(self.thread_pool))
         self.commands = {}  # the dynamically populated list of commands available on the bot
         self.re_commands = {}  # the dynamically populated list of regex-based commands available on the bot
+        self.command_filters = []  # the dynamically populated list of filters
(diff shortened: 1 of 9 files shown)


errbot/errBot.py
```diff
@@ -265,6 +266,18 @@ def process_message(self, mess):
...
         """Process and execute a bot command"""
 
+        # first it must go through the command filters
+        try:
+            for cmd_filter in self.command_filters:
+                mess, cmd, args = cmd_filter(mess, cmd, args)
+                if mess is None:
+                    log.info("Command %s deferred." % cmd)
+                    return
+        except ACLViolation as e:
```
zoni: This tightly couples `errBot` itself with the ACL filter, which I think we should try to avoid to avoid *"special casing"* certain filters.

I wonder if we can make this more generic. Have filters signal whether the bot should continue processing the command (having it go through other filters, if any, then run the command) or if it should stop processing (because the filter blocked it)

gbin: Ha you are right. I gonna push the ACLException in the acls plugin with directly a response to the messsage and just let the if none logic here (that block the message basically).

PR #462, opened 2015-08-27
WIP on simplifying the configuration of the bot

gbin:
The goal here is to make a small wizard accessible from the command line.


mrshu:
@gbin If your goal is a command line wizard would it make sense to save this data to some sort of serializable format (yaml, json, ...) so that you would not have to try to match a Python file template? The generated `config.py` might then just load the data inserted via the wizard from the serialized format and set it to appropriate variables.

Of course, the power users could still edit config.py as that would get created as part of the process.

gbin:
@mrshu I am still pondering our options here.

Indeed it kinda also collide with a provisioning option:
like you said: just giving the parameters in json and the setup will create the config.py for the user.

Also, it might be interesting to give an option to do that for the plugin initial setups (imagine setting externally the initial ACLs without having to do that chatting with the bot).

mrshu:
@gbin sorry, what I meant was a bit different.

I thought that since there is a finite amount of configuration variables in `config.py` and since we all agree that creating `.py` files from a template is certainly not the way to go (for one you cannot easily read them in other languages) it might make sense dump the data gathered via the wizzard into a `.json` file or something like that and have a special `config.py` prepared that will load the data from that `.json` into the respective configuration variables.

gbin:
This is overengineered and feels like a feature creep again, I drop this. We gonna redo it with a simple script that makes a data directory and setup a base config there, it will do already 90% of the work.

PR #471, opened 2015-09-15
config: BOT_ADMINS also accepts a unique string

### LINKED ISSUE #461: BOT_ADMINS should also accept a unique string
Most installations might have only one admin and the trailing "," is confusing.

mrshu:
* As noted in #461, in many cases an installation will only have one
  admin and therefore it does not make too much sense to force the
  BOT_ADMINS to be a tuple.

  This commit makes BOT_ADMINS also accept a unique string which is
  then converted to a single element tuple for backwards compatibility.

--------------------------------------------------------

Note that this should fix #461.

### DIFF
--- a/errbot/botplugin.py
+++ b/errbot/botplugin.py
@@ -47,6 +47,10 @@ class BotPluginBase(StoreMixin):
         For exemple you can access:
         self.bot_config.BOT_DATA_DIR
         """
+        # if BOT_ADMINS is just an unique string make it a tuple for backwards
+        # compatibility
+        if isinstance(self._bot.bot_config.BOT_ADMINS, str):
+            self._bot.bot_config.BOT_ADMINS = (self._bot.bot_config.BOT_ADMINS,)
         return self._bot.bot_config
 
     def init_storage(self):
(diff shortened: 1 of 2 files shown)


gbin:
thx !

mrshu:
@gbin thanks for merging this in.

I wonder if the docs should be updated too?

zoni:
I think we should also do this for `CHATROOM_PRESENCE`, `BOT_ALT_PREFIXES`, `BOT_ALT_PREFIX_SEPARATORS` and `DIVERT_TO_PRIVATE`, otherwise the behaviour is quite inconsistent.

Another option, to avoid having to deal with this in code is to make examples in `config-template.py` use lists (`[]`) instead of tuples (`()`), which in practice will give the same behaviour without the confusing case of having to have a trailing `,` if there's only one element.

mrshu:
What @zoni suggests makes a lot of sense. After all, the code always uses `in` when checking for values in these tuples so moving to lists would not be such a big deal funcionality-wise.

When it comes to tuples vs. lists, it seems that in this case lists would make more sense syntactically as touples are sequences in which positions have semantic value whereas the individual items in lists should be functionally equivalent (http://news.e-scribe.com/397).

However, I feel that there must be a reason why @gbin decided to go this way which is why I would like to hear his thoughts on it.

Issue #473, opened 2015-09-16
We are polluting dist-packages/tests


gbin:
This directory should be excluded from installation.

PR #475, opened 2015-09-17
setup.py: Fix pollution of dist-packages by tests

mrshu:
* It seems that when installing err not only the errbot package is
  created but also a 'tests' package. This commit makes setup.py exclude
  that package, as described in http://stackoverflow.com/a/29978235

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -96,7 +96,7 @@ if __name__ == "__main__":
     setup(
         name="err",
         version=VERSION,
-        packages=find_packages(src_root),
+        packages=find_packages(src_root, exclude=['tests']),
         scripts=['scripts/err.py'],
 
         install_requires=deps,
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -96,7 +96,10 @@ def read(fname):
     setup(
         name="err",
         version=VERSION,
-        packages=find_packages(src_root),
+        packages=find_packages(src_root, exclude=['tests',
+                                                  'tests.*',
```
gbin: we need those other ones ?

gbin:
This fixes #473

mrshu:
@gbin I tested this in a clean environment and can confirm that `tests` are skipped from the distribution.

setup.py
mrshu: @gbin currently we do not, but I would leave them there, just to be safe.

If you do not agree I am happy to remove them.

gbin: yeah I would prefer to leave it to the minimal set, it is clearer. We'll add them as necessary. thx !

mrshu:
@gbin Removed the unnecessary excludes.

gbin:
Thanks !

PR #480, opened 2015-09-27
Added COMPACT_OUTPUT option. 

gbin:
This is IRC only (for now) and is activated by default on IRC.
This simply removes the ascii art around the tables... also for now.

image


zoni:
I propose we make this a general option, available on all back-ends. For example on Telegram the tables are completely messed up too:

screenshot from 2015-09-27 22 00 07

gbin:
@zoni this needs to be added per backend, can you do the telegram one ? (I don't have a setup to test it)

zoni:
I'll give it a try!

Issue #476, opened 2015-09-17
Change script from err.py to errbot


gbin:
For 2 reasons:
1. it is cleaner, usually commands have no .py extensions
2. in general I would like to migrate the name of the bot as much as possible from err to errbot (it is way easier to discover than the generic english "err" and err variable used all over the place for error)

zoni:
+1, this is something I've thought of many times in the past as well.

Two remarks:

* We should probably keep the `err.py` for a while as backward compatibility
* We should use [entry_points](https://pythonhosted.org/setuptools/setuptools.html#automatic-script-creation) for this in `setup.py` while we're at it.

mrshu:
I am also for this, especially since I many times ran into `err` which was really an abreviation of error rather than the name of this bot.

@zoni would you mind expanding a bit on why would be

zoni:
> @zoni would you mind expanding a bit on why would `entry_points` be better than the script that is currently in place?

Certainly. There are two main benefits to using `entry_points`:

1. It will ensure that, on windows, a proper `.exe` executable is created, making it a little bit easier for people on that OS.
2. But more importantly, it makes sure that all Python path settings and such are always correct. For example, if you installed Err into a virtualenv (lets say at `/foo/bar/venv/`) you can start Err directly by calling `/foo/bar/venv/bin/errbot` and it would just work. No need to source `/foo/bar/venv/bin/activate` first or call `/foo/bar/venv/bin/python /foo/bar/venv/bin/errbot`.

mrshu:
@zoni Thanks, I understand the matter way better now and agree that `entry_points` should be a win in general, especially in the future.

mrshu:
@gbin can we say this has already been done in #481 ?

gbin:
ha yes sure. Small tip: if you say in one of your commit message: fixes #476, it will automagically close the issue for you ;)

mrshu:
@gbin yeah, I know. It's just that I am never sure if it actually fixes that =)

PR #481, opened 2015-09-27
update: Moved from script to entry_points

mrshu:
* Moved err.py script to `errbot` entry point that will ensure that a
  proper executable is created on Windows and simplify life with Python
  path settings and such.

  For further information see https://github.com/gbin/err/issues/476

### DIFF
--- a/errbot/errbot.py
+++ b/errbot/errbot.py
@@ -140,7 +140,7 @@ def get_config(config_path):
     return config
 
 
-if __name__ == "__main__":
+def main():
 
     execution_dir = getcwd()
 
@@ -259,3 +259,6 @@ if __name__ == "__main__":
 
     main(backend, logger, config, restore)
     log.info('Process exiting')
+
+if __name__ == "__main__":
+    main()
--- a/setup.py
+++ b/setup.py
@@ -97,7 +97,11 @@ if __name__ == "__main__":
         name="err",
         version=VERSION,
         packages=find_packages(src_root, exclude=['tests']),
-        scripts=['scripts/err.py'],
+        entry_points={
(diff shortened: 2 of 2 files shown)


errbot/errbot.py
```diff
@@ -140,7 +140,7 @@ def get_config(config_path):
     return config
 
 
-if __name__ == "__main__":
```
gbin: maybe leave the option to start it directly too ?
if __name__ == "__main__":
    main()

So we can still start it from the dev environment with no install ?

mrshu:
Note a few things here:

- This should fix #476
- The debug output will change dramatically, most notably `err.py` will be replaced by `errbot.errbot` which might not be desirable
- The script new file is called `errbot.py` in the `errbot/` directory which might be confusing, especially since there is also `errBot.py` file in the same directory. Another option might be to use a name like `__main__.py` or something similar which people do use for their `entry_points` to big Python applications but considering how often this name will show up in the debug output I think that it might be even worse than what I managed to come up with.
- While I tried to test as much as I could I was unable to check `ispydevd`, that is whether the condition `frame[1].endswith("pydevd.py")` will ever be met in the current setup, as the Python's way of loading `entry_point`s is somewhat complex

In any case I believe that this move is a move in the right direction.

errbot/errbot.py
mrshu: @gbin I was under the impression that in the dev environment you will always do `python setup.py develop` and therefore this will not be necessary but I will gladly add such option.

mrshu: @gbin should be there now.

gbin:
thx

PR #482, opened 2015-09-28
irc: Added nickserv_password option

mrshu:
* Added an option to specify nickserv_password for the IRC backend. If
  this option is present the bot will try to identify to NickServ with
  this password.

### DIFF
--- a/errbot/backends/irc.py
+++ b/errbot/backends/irc.py
@@ -259,6 +259,7 @@ class IRCConnection(SingleServerIRCBot):
                  ssl=False,
                  password=None,
                  username=None,
+                 nickserv_password=None,
                  private_rate=1,
                  channel_rate=1,
                  reconnect_on_kick=5,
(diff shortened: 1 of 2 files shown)


mrshu:
cc @gbin @ZucchiniZe

errbot/backends/irc.py
```diff
@@ -289,6 +291,13 @@ def connect(self, *args, **kwargs):
 
     def on_welcome(self, _, e):
         log.info("IRC welcome %s" % e)
+
+        # try to identify with NickServ if there is a NickServ password in the
+        # config
+        if self.nickserv_password:
+            msg = 'identify %s' % self.nickserv_password
```
mrshu: @gbin note that I am only using this convention because I found it in other places in the codebase. Under normal circumstances I would use `.format` here and I am happy to change this accordingly.

gbin:
thx

PR #484, opened 2015-09-28
readme: Updated to reflect err.py changes

mrshu:
* Updated README.rst in order to reflect the overall move from `err.py`
  to `errbot`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -144,17 +144,17 @@ connect to your chosen backend (XMPP, Hipchat, Slack ...) server.
 The first time you start Err, it is recommended to run it in foreground mode. This can
 be done with::
 
-    <path_to_install_directory>/scripts/err.py
+    <path_to_install_directory>/errbot/err.py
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -144,17 +144,17 @@ connect to your chosen backend (XMPP, Hipchat, Slack ...) server.
 The first time you start Err, it is recommended to run it in foreground mode. This can
 be done with::
 
-    <path_to_install_directory>/scripts/err.py
+    <path_to_install_directory>/errbot/err.py
```
gbin: maybe we can encourage people to do the cleaner `pip install -e .` instead and execute `errbot` ?

mrshu: @gbin I wanted to keep the changes to `README` at the bare minimum, but if you want I can rewrite this  whole section so that `errbot` will be the only suggested option.

gbin:
I think it would be better to document the proper way if you have time. Thanks a lot !

PR #487, opened 2015-10-01
Create bot.all_commands, which unites commands and re_commands

jamesob:
references https://github.com/gbin/err/issues/486

Introduces `ErrBot.all_commands`, which is just a united `dict` consisting of both `commands` and `re_commands`. 

This prevents `!help`, the QT client, and some ACL operations from ignoring re_commands.

### DIFF
--- a/tests/dummy_plugin/dummy.py
+++ b/tests/dummy_plugin/dummy.py
@@ -10,7 +10,7 @@ class DummyTest(BotPlugin):
         """This runs foo."""
         return 'bar'
 
-    @re_botcmd(pattern=ur"")
+    @re_botcmd(pattern=ur"plz dont match this")
     def re_foo(self, msg, match):
         """This runs re_foo."""
         return 'bar'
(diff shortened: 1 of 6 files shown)


jamesob:
tests on the way...

tests/dummy_plugin/dummy.py
```diff
@@ -6,4 +7,10 @@ class DummyTest(BotPlugin):
     """
     @botcmd
     def foo(self, msg, args):
+        """This runs foo."""
+        return 'bar'
+
+    @re_botcmd(pattern=ur"plz dont match this")
+    def re_foo(self, msg, match):
```
mrshu: @jamesob is the `u` really necessary here?

jamesob: yeah, just fixed that. Whatever insane mechanism backports python 3 -> 2 on `pip install` left me with that handy artifact.

mrshu:
@jamesob this LGTM, however I think there might be a reason why this hasn't been done in the past.

It would be also great if you could squash your commits into one so that the log would not get polluted with small commits that are not described in detail.

gbin:
@mrshu I think this is simply an overlook when we introduced re_botcmds. Thanks a lot @jamesob for the fix !

jamesob:
Sure thing! Thanks for the great software!

gbin:
squashed and pushed

PR #488, opened 2015-10-02
Allow @arg_botcmd decorated handlers to  yield results

andrevdm:
When using the @arg_botcmd decorator yields don't work correctly. ErrBot returns "<generator object....>" rather than the actual text.

This is because @arg_botcmd is wrapping the handler in a new function and 2when _execute_and_send calls inspect.isgeneratorfunction on this wrapper it always gets a false, 
because its checking the wrapper not the underlying handler. 

This fix forces the wrapper to always be a generator which then works if the underlying 
handler is a generator or not.

### DIFF
--- a/errbot/__init__.py
+++ b/errbot/__init__.py
@@ -4,6 +4,7 @@ import logging
 import re
 import shlex
 import sys
+import inspect
 
 from .core_plugins.wsview import bottle_app, WebView
 from .utils import compat_str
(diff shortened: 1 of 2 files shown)


andrevdm:
@gbin test added and passing. BTW that second return does seem to be required, i.e. rather than a return. Its been years since I've written python but that's not what I would have expected... It works perfectly though

gbin:
Thanks a lot !

andrevdm:
No problem, thanks for ErrBot! :)

PR #497, opened 2015-10-09
added prereq for hipchat backend

johnsmclay:
added prereq for hipchat backend.  In Python 3.4 it's more apparent, but in 2.7 it was just a cryptic "unable to find backend hipchat"


gbin:
thx

PR #498, opened 2015-10-09
Slack changes to facilitate bot interaction

jamesob:
We've got a monitoring bot that posts certain alerts into our slack (via webhook) when various metrics break certain thresholds. We want our errbot to grab the graphs associated with these metrics and paste them in; this involves reading attachments from bot users in Slack.

selection_027

(riemann is a webhook bot, Scruffy is an errbot)

This PR includes the changes necessary to accomplish the above:

- Slack backend no longer errors on messages from bot users
- The `SlackMessage` class exposes `attachments` to plugin writers
- Messages with blank `text` fields (like those sent from the bot in this case) are no longer ignored
- Markdown rendering is made optional

cc @cyc @greglook

### DIFF
--- a/errbot/backends/base.py
+++ b/errbot/backends/base.py
@@ -40,18 +40,27 @@ class Message(object):
     make sense in the context of other back-ends.
     """
 
-    def __init__(self, body='', type_='chat', frm=None, to=None, delayed=False):
+    def __init__(self,
+                 body='',
+                 type_='chat',
+                 frm=None,
+                 to=None,
+                 delayed=False,
+                 extras=None):
         """
         :param body:
             The plaintext body of the message.
         :param type_:
             The type of message (generally one of either 'chat' or 'groupchat').
+        :param extras:
+            Extra data attached by a backend
         """
         self._body = compat_str(body)
         self._type = type_
         self._from = frm
         self._to = to
         self._delayed = delayed
+        self._extras = extras or dict()
 
     @property
     def to(self):
@@ -144,6 +153,10 @@ class Message(object):
     @delayed.setter
     def delayed(self, delayed):
         self._delayed = delayed
+
+    @property
+    def extras(self):
+        return self._extras
 
     def __str__(self):
         return self._body
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -3,6 +3,7 @@ import logging
 import re
 import time
 import sys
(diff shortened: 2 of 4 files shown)


errbot/backends/base.py
```diff
@@ -517,7 +517,7 @@ def __init__(self, _):
         self._reconnection_multiplier = 1.75  # Delay multiplier
         self._reconnection_jitter = (0, 3)    # Random jitter added to delay (min, max)
 
-    def send_message(self, mess):
+    def send_message(self, mess, convert_to_md=True):
```
gbin: maybe this should be named:
convert_from_md
or maybe: raw_message ?

gbin:
Thx for the PR, it makes a lot of sense, just have a small comment on the param name. Otherwise @zoni  any comment on the Slack side ?

gbin:
Actually, looking at it more carefully, can't we imagine a way to do that with the standard markdown instead ?

Like what is a slack attachement ?
https://api.slack.com/docs/attachments

Everything can be mapped to markdown:
link: check
url: check
image: check
fallback: text check
fields: ~ a table in markdown
color: check

Maybe we can make a specific markdown template called attachment.md that will work in all the backends and be magically recognized for slack to inject the parameters in the message as attachement ?

jamesob:
@gbin actually, the markdown conversion changes aren't strictly needed to enable the behavior I described above. We were hoping to use it to format the graph links more prettily, but it looks like [slack's RTM client doesn't support formatting anyway](https://github.com/slackhq/hubot-slack/issues/43#issuecomment-143618259). I'll remove those changes and we can come up with a better solution later.

gbin:
I think disabling the markdown conversion with an option is ok (we need it also for the botcmd decorators also BTW).
It is more about:
one way the plugin want to send an attachement, it could be done with a conventional markdown template so it would be intercepted by the slack backend and render it natively and for other backends, it would be rendered a correctly formatted message.

The other way is more tricky: how to get an attachment from slack to the plugins.
maybe a reasonable text format would do it.

I think we should split those in 3 PRs as they can be done independently. Do you want me to help on a couple ?

gbin:
Note: feel free to discuss it on https://gitter.im/gbin/err if you have questions

jamesob:
Good to go here?

gbin:
Thx

PR #505, opened 2015-10-22
Respect optional prefix with re_botcmd

Kromey:
When the setting BOT_PREFIX_OPTIONAL_ON_CHAT is set to True, plugin commands decorated with the re_botcmd decorator would ignore it and still require the bot's prefix; this fixes that by adding an additional check when testing re_botcmds.

This resolves gbin/err#504.

### DIFF
--- a/errbot/errBot.py
+++ b/errbot/errBot.py
@@ -229,7 +229,7 @@ class ErrBot(Backend, BotPluginManager):
         # Try to match one of the regex commands if the regular commands produced no match
         matched_on_re_command = False
         if not cmd:
-            if prefixed:
+            if prefixed or (type_ == "chat" and self.bot_config.BOT_PREFIX_OPTIONAL_ON_CHAT):
                 commands = self.re_commands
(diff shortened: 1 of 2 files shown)


gbin:
awesome! May I ask to add a unittest with that ?

Kromey:
Can't you just take my word that it works? ;-P

I'll try to, but it will take a little while as I'm still rather new to the practice in general, and never done it in Python at all, and not entirely sure how to write one for this code anyway. But I'm sure I can figure it out...

Kromey:
So I'm not sure if you consider it appropriate to be relying on dummy_plugin in my tests, especially matching against the pattern 'plz dont match this', but I'm also not sure how I would go about adding an additional plugin for this test, or if it should be done by adding another command to dummy_plugin.

That being said, though, the tests work, in the sense that they fail without the fix but pass with it.

gbin:
This is good enough ! Thanks for the contribution !

PR #506, opened 2015-10-22
Various small "improvements" based around !help

jamesob:
selection_036

I quoted "improvements" because I think some of these things are only *subjectively* better; no less, I think this changeset does improve the usability of err.

### Changes

- Add `DISABLED_PLUGINS` config option; we found that a lot of the builtin plugins we don't need, and they just pollute `!help` and confuse people.
- Various `!help` changes
  - make help lookups case-insensitive; enforcing case here is pretty annoying and isn't really necessary
  - make the default `!help` more like hubot's; i.e., list all of the commands instead of just the plugins. I've found this is more newb-friendly, and by the time you prune your plugins with `DISABLED_PLUGINS`, the list shouldn't be terribly long
  - show the required regexp for `re_botcmd`; without this, help for these commands is basically useless

Hope you guys like these changes!

cc @greglook @cyc


sijis:
I dig this. I agree the case sensitivity for !help is annoying.

gbin:
I like it too !
Please try to split the PRs when you have multiple independent changes like that. Here is why :)

1. the long help, it was like that originally but under IRC it is really hard for them to have a long response (especially with the rate limiter!) but like you I do prefer the long form on almost all the other backends ... is it possible to have the behavior either under a setting ?

2. the DISABLED_PLUGINS doesn't exist because we have the "!plugin blacklist" that should work exactly the same way. I am open to remove the annoying admin warning :).

3. regexes and case insensitivity, awesome ! :)

jamesob:
@gbin cool, I'll make those changes. But for `!plugin blacklist` is it possible to do this statically within the config file? That's what I'd really like to do.

gbin:
@jamesob 
I would like to avoid putting redundant state in the core.db and config.py. 

I guess this is because you want to provision bots with a static configuration.

Maybe put the bot in a good state (correctly blacklisted plugins, webserver configured correctly etc.) and issue a `!backup` command that would save the empty state of your bot ?

Then at provision time you can do a `errbot --restore`

This is not ideal but in general I would like to find a way to "preconfigure" correctly errbot but still haven't found a clean way of doing that. I you have any idea, I am interested :)

jamesob:
@gbin would really like to be able to have static configuration instead of doing some db-snapshot type thing. We use docker, and so snapshotting state is not easy for us. Is it okay with you if I leave this in `config.py` to facilitate our usecase? Otherwise we have to maintain a fork :\

gbin:
Any progress on the core plugins whitelisting ?

jamesob:
hey @gbin sorry -- been AFK for a bit. Will try to knock that out today or tomorrow.

gbin:
Made the blacklisting part on #583

gbin:
@jamesob can you restart a PR just with the help changes ? thx.

PR #508, opened 2015-10-22
WIP: Redis implemented and tested

jamesob:
I've added a tested and working proof of concept for using Redis storage. The implementation is fairly simple; you provide redis config data in the `config.py`, then `ErrBot` injects a wrapped redis client into each plugin instance. The wrapped client automatically namespaces based on the plugin classname.

Obviously there's a bit more polish to do here in terms of doc, but I can start on that if you like the approach in general.

### DIFF
--- a/errbot/backends/test.py
+++ b/errbot/backends/test.py
@@ -128,6 +128,22 @@ class TestMUCRoom(MUCRoom):
         return self._name == other._name
 
 
+class FakeRedisClient():
+
+    def __init__(self, *args, **kwargs):
+        self.state = {}
+
+    def set(self, key, val, **kwargs):
+        self.state[key] = val
+
+    def get(self, key, **kwargs):
+        return self.state.get(key)
+
+
+def get_fake_redis(*args, **kwargs):
+    return FakeRedisClient()
+
+
 class TestBackend(ErrBot):
     def __init__(self, config):
         config.BOT_LOG_LEVEL = logging.DEBUG
@@ -135,7 +151,7 @@ class TestBackend(ErrBot):
         config.BOT_IDENTITY = {'username': 'err'}  # we are testing with simple identfiers
         self.bot_identifier = self.build_identifier('Err')  # whatever
 
-        super().__init__(config)
+        super().__init__(config, get_redis_client=get_fake_redis)
         self.incoming_stanza_queue = Queue()
         self.outgoing_message_queue = Queue()
         self.sender = self.build_identifier(config.BOT_ADMINS[0])  # By default, assume this is the admin talking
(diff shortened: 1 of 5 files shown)


gbin:
Is it possible to plug that behind the standard plugin persistance ? And maybe augment it if we need a feature and emulate it under the standard pickle storage ?

My worry here is plugin fragmentation : some plugins will require redis, some don't, some installations will have redis, some don't.
And also, why make redis special ? what about postgresql, riak, etcd ?

jamesob:
The existing persistence interface IMO is kind of rough -- inheritance (instead of composition) is clunky and harder to understand and test.

If we can discuss and agree on a better interface for dealing with multiple different storage mechanisms, I'm definitely willing to try implementing. I'm not wild about trying to write some kind of ultimate polyglot storage abstraction layer; the returns there diminish very quickly. I think we have to be honest with ourselves -- this is just a chatbot, and the most important thing here IMO is simplicity and ease of use. I absolutely agree that we should come up with an interface that plugin writers can rely on, but like I said, I don't want to overengineer.

jamesob:
Also: consider that some users may want to use different types of storage. Maybe redis is good for some things while postgres is good for others.

gbin:
@jamesob 
I thought a little bit more about this.

The goal of the self.shelf is to be universal and simple enough for storing small stuff.
Simple as it doesn't introduce dependencies, doesn't introduce any daemon you need to run beside your plugin etc. All you need to know is how to use a python dictionary to use it (no ACID, no messaging, no queries...).

This makes sense for storing stuff related to ... chatting ! with who I talked last ? what is my state ? who should I ambush once they are online ? etc ...

If you intend to use it as your **business** storage, this is not the right choice then use your own specialized storage which is a dependency *of the plugin*, not the bot !

An example of that: timemachine uses lucene https://github.com/gbin/err-timemachine/blob/master/timeMachine.py#L30

Redis seems to land squarely on the second category.
Now how can we simplify the life of the people using redis instead of stuff on disk ?

Why not making a simple python package introducing a RedisBotPlugin base class ?
- plugins would inherit from it to use it
- the config could be centralized for all the plugins using this as a base

What do you think ?

jamesob:
@gbin yeah actually, that sounds good. I'm not necessarily into the inheritance pattern, but I think we can definitely separate the redis stuff out and make it easy for plugins to do something like
```python
import redis_err

class SomePlugin(BotPlugin):

  @cached
  def get_redis(self):
    return redis_err.get_client_for_plugin(self, ...)
  ...
```
or something like that.

gbin:
yes, that works too indeed !

PR #516, opened 2015-10-26
Various arg_botcmd improvements

zoni:
This PR introduces a number of enhancements to how the `arg_botcmd` decorator works:

*  Ensure arg_botcmd error and help output goes into chat and avoid raising SystemExit

	By default, argparse will print errors, usage and help text directly to
	stdout/stderr. It will also raise SystemExit to make the program stop
	with an appropriate error code.

	While this is appropriate for a CLI program, it is horrible for Err. The
	SystemExit causes some strange behaviour where it doesn't make the bot
	exit, but does prevent some processing (!restart no longer works, for
	example) and the printing to stdout/stderr means you get no feedback
	when issuing commands with incorrect arguments.

	With these changes, all this output is nicely returned to the user,
	instead.

* Ensure the command name is used in the usage as well, instead of the
default `sys.arv[0]` which resulted in one of `err.py`, `errbot` or
`run_tests.py`.

* Convert — to -- when parsing arg_botcmds

	Some clients (looking at you, Telegram) convert certain ASCII character
	combinations to nicer-looking unicode representations, which break the
	argument parser. Account for this and undo the conversion to avoid user
	frustration.

*  Make "argument unpacking" on arg_botcmd optional

	With this change it becomes possible to define a single `args` argument
	to hold all the argparser arguments, removing the need to define them in
	the function signature itself.

	This is useful especially on commands with lots of argument flags, as
	otherwise you have to make sure you define all your arguments in just
	the right order, which is very error prone.

### DIFF
--- a/errbot/__init__.py
+++ b/errbot/__init__.py
@@ -148,7 +148,7 @@ def re_botcmd(*args, **kwargs):
 
 
 def arg_botcmd(*args, hidden=False, name=None, admin_only=False,
-               historize=True, template=None, **kwargs):
+               historize=True, template=None, unpack_args=True, **kwargs):
     """
     Decorator for argparse-based bot command functions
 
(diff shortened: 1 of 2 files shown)


zoni:
Preview of the usage/help text on Telegram (which removes blank lines for some reason):

2015-10-26 15 03 37 selection 001

Issue #519, opened 2015-10-28
md_escape only escapes [ '{', '}' ]


richiebbb:
Hi, love the program, recently upgraded to 3 and needed to escape some things to avoid it being seen as markdown.

In `/errbot/rendering/__init__.py`, we have:
```
def md_escape(txt):
    """ Call this if you want to be sure your text won't be interpreted as markdown
    :param txt: bare text to escape.
    """
    txt = txt.replace('{', '\{')
    txt = txt.replace('}', '\}')
    return txt
```

This only replaces `{` and `}`. I replaced my local copy with:
```
    """ Call this if you want to be sure your text won't be interpreted as markdown
    :param txt: bare text to escape.
    """
    for char in Markdown.ESCAPED_CHARS:
        txt = txt.replace( char, '\\' + char)
    return txt
```
which works.

Not really sure if this is an issue or if I'm making some sort of mistake in my usage, so feel free to close this if it is irrelevant.

zoni:
Since @gbin wrote all the markdown stuff I'll wait for him to confirm, but I think your version is the proper way to do it indeed. Would you be willing to open a pull request with these changes applied?

gbin:
Indeed @richiebbb it is quite incomplete, I am fixing it.

PR #520, opened 2015-10-28
Added a comprehensive md escaping.

gbin:
It should now preserve the original text from markdown transformations.

### DIFF
--- a/errbot/rendering/__init__.py
+++ b/errbot/rendering/__init__.py
@@ -9,6 +9,12 @@ ATTR_RE = re.compile(AttrListTreeprocessor.BASE_RE)
 # Here are few helpers to simplify the conversion from markdown to various
 # backend formats.
 
+_md_escape_re = re.compile('|'.join(re.escape(c) for c in Markdown.ESCAPED_CHARS))
+
+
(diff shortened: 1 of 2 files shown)


gbin:
Should fix #519, @richiebbb please LGTM if it works for you.

errbot/rendering/__init__.py
```diff
@@ -9,6 +9,12 @@
 # Here are few helpers to simplify the conversion from markdown to various
 # backend formats.
 
+_md_escape_re = re.compile('|'.join(re.escape(c) for c in Markdown.ESCAPED_CHARS))
+
+
+def _md_escape_trans(match):
```
mrshu: @gbin should this "helper" function be inside the `md_escape` function? I do not suppose it will be used anywhere else so it makes sense to encapsulate it in `md_escape`.

gbin: Done + the naming convention for the constant was off compared to the other regexp in the file, fixed that too.
thx

zoni:
Gets a :+1: from me.

richiebbb:
likewise :+1:, thanks man

Issue #511, opened 2015-10-23
templating doc is outdated (still uses html and import)


gbin:
http://errbot.net/user_guide/plugin_development/messaging/#templating

gbin:
This should look like (#4) : http://klaig.blogspot.com/2015/10/errbot-recipes.html

sijis:
I can put this together.

zoni:
This got fixed, documentation has been updated.

PR #521, opened 2015-10-29
updated messaging dev guide to reference Markdown features (#511)

sijis:
We've had the ability to support Markdown and this changes our
documentation to reflect that new feature and its new capabilities.

I'd still have to include the new send_templated() feature that was recently added.

### DIFF
--- a/docs/user_guide/plugin_development/messaging.rst
+++ b/docs/user_guide/plugin_development/messaging.rst
@@ -55,24 +55,20 @@ For Err=>2.3 you need to use : self.build_identifier("user@host.tld/resource") b
 Templating
 ----------
 
-It's possible to send `XHTML-IM
-<http://xmpp.org/extensions/xep-0071.html>`_ responses using `Jinja2
+It's possible to send `Markdown
+<http://daringfireball.net/projects/markdown/>`_ responses using `Jinja2
 <http://jinja.pocoo.org/>`_ templates.
 
 To do this, first create a directory called *templates* in the
 directory that also holds your plugin's *.plug* file.
 
-Inside this directory, you can place XHTML-IM templates (with a
-*.html* extension) that extend `base.html` and contain a block
-called `body` in place of the content you wish to show. For example
-this *hello.html*:
+Inside this directory, you can place Markdown templates (with a
+*.md* extension) in place of the content you wish to show. For 
+example this *hello.md*:
 
-.. code-block:: html
+.. code-block:: python
 
-    {% extends "base.html" %}
-    {% block body %}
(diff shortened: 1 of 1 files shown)


sijis:
I'm not exactly sure by the fixing the 'import' in the issue.

docs/user_guide/plugin_development/messaging.rst
```diff
@@ -110,12 +106,12 @@ like so:
...
         def hello(self, msg, args):
             """Say hello to someone"""
-            response = tenv().get_template('hello.html').render(name=args)
+            response = tenv().get_template('hello.md').render(name=args)
             self.send(msg.frm, response, message_type=msg.type)
 
 .. note::
     In both cases a plain-text version of your template is automatically
-    generated for clients and back-ends that do not support XHTML-IM.
+    generated for clients and back-ends that do not support Markdown.
```
zoni: I think this note doesn't quite make sense anymore in this context. I think we should replace this with something along the lines of *"Not all networks support the full range of markup available with Markdown so backends convert the markup into a representation that is most suited for the network you are using. Because of this, you may find that your template is rendered differently from how you intended, or the template look different when compared between the various backends."*.

sijis:
I think that's fair. 
So in essence the markdown will not fallback to text only but will try to render the template as best as possible in respect to the corresponding backend.

docs/user_guide/plugin_development/messaging.rst
```diff
@@ -110,12 +106,12 @@ like so:
         @botcmd(template="hello")
         def hello(self, msg, args):
             """Say hello to someone"""
-            response = tenv().get_template('hello.html').render(name=args)
+            response = tenv().get_template('hello.md').render(name=args)
```
gbin: on 3.1.x this is still true
but on 3.2 it will be send_templated (current head)

docs/user_guide/plugin_development/messaging.rst
```diff
@@ -110,12 +106,12 @@ like so:
...
             self.send(msg.frm, response, message_type=msg.type)
 
 .. note::
     In both cases a plain-text version of your template is automatically
```
gbin: yeah, this is obsolete: before we did an horrible hack job doing im-xtml to plain text conversion.

Now it is markdown to whatever the backend supports.

So you can simply remove the note IMHO.

sijis:
I could add a note that says prior to 3.1.1 use the tenv() hack. So current plugin devs have the right info.

gbin:
@sijis send_templated is a 3.2 feature so you can make the change directly on master (it will be published when we gonna release 3.2).
thx !

PR #523, opened 2015-10-29
Ported botplugin to type annotations.

gbin:
Also added types as a dependency for backward compatibility with python
3.3 and 3.4.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ from setuptools import setup, find_packages
 py_version = sys.version_info[:2]
 PY2 = py_version[0] == 2
 PY3 = not PY2
-PY35 = py_version[:2] >= [3, 5]
+PY35 = py_version[:2] >= (3, 5)
 
 ON_WINDOWS = system() == 'Windows'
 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -23,6 +23,8 @@
 py_version = sys.version_info[:2]
 PY2 = py_version[0] == 2
 PY3 = not PY2
+PY35 = py_version[:2] >= (3, 5)
```
zoni: I think this name is misleading. Perhaps use `PY35_OR_GREATER` instead?

zoni:
This is quite nice. Sphinx prints these in the generated documentation as well, and there's a [Sphinx extension](https://pypi.python.org/pypi/sphinx-autodoc-annotation) that enhances that a little further, which will make it look like so:

2015-10-30 12 09 39 selection 001

There's [an issue](https://github.com/hsoft/sphinx-autodoc-annotation/issues/2) on the project which I also ran into when I added it, but I was able to patch that quite trivially so I'll submit a PR upstream. Then I can add it to our sphinx extension list as well.

gbin:
@zoni this is awesome for our users !

zoni:
Yup!

setup.py
gbin: done

Issue #322, opened 2015-01-22
Make Err connect to multiple back ends at the same time


gbin:
See if we can configure several back ends at the same time without complicating the setup process.

gbin:
With the backend plugins, we are one step closer to this.

What we could do make the storage smarter so several instances of Errbot can share it and then instanciate several backends (ErrBot instances) sharing the plugins.

gbin:
We can now do it with some limitations with several instances of errbot as the instances can share the same persistence. Maybe if there is some interest about making a generic bridge using the backends as implementations, we could start it as a separate project.

PR #528, opened 2015-11-02
Fixes #439 - Slack: inconsistency in BOT_ADMINS.

### LINKED ISSUE #439: Slack: inconsistency in BOT_ADMINS
I previously filed #436, which I had to prefix my username with '@' character. However, that seems to be inconsistent with how acl's are checked, as I'm now seeing 'this command requires bot-admin privileges' when doing things like ```!plugin reload <pluginname>```.

The temporary fix is to just add my username without the prefixed '@' character and privileges message goes away and works as expected.

Rohja:
When using Slack as Backend, allowed administrators username
are now @username, username or #channel/username.

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -496,6 +496,8 @@ class SlackBackend(ErrBot):
                                 "should be #channelname/username or @username : '%s'" % txtrep)
             channelname, username = plainrep.split('/')
         else:
+            username = txtrep
+        if not username or not len(username):
             raise Exception("Unparseable slack identifier, " +
                             "should be #channelname/username or @username : '%s'" % txtrep)
 


zoni:
I don't see how this would fix anything. By removing that exception you open up the case where the code continues without ever setting `username` to a value. This would then cause an exception a few lines down where it tries to [determine the userid](https://github.com/gbin/err/pull/528/files#diff-4d2174262014c51fc8b38e02b99962bdL502), but username is not defined when it reaches that code.

Perhaps we can assume that an identifier which doesn't start with `@` and neither contains a `/` to be a username though.

Rohja:
That's true, I forgot this case.
Let me fix that.

Rohja:
With this code ( f93254f ): If the username start with '@', then the '@' is remove and the username is the last part.
If the username start with '#', then the result is splited by '/' (first part -> Channel, second part -> username).
Else, the username is used as is.
Finally, if the username is None or length is 0, it raise exception.

Did I forget something this time ?

gbin:
LGTM, I have just a couple questions.

```
@foo == user(foo)
#bar/foo == channel(bar), user(foo)
foo == user(foo)
```

But what about `bar/foo` ?

sijis:
At first glance, i would think, we should check for a ```/``` in username and then parse that as if ```#bar/foo``` was passed.

Although, since this is backend specific, I suppose it makes sense to encourage users to use that backend's username format. Will doing this impact any long term ideas like #322?

gbin:
for #322 it has not much impact because identifiers would be per backend anyway

zoni:
@sijis 
> I suppose it makes sense to encourage users to use that backend's username format.

Yes, that is an approach I strongly advocate and the reason I didn't like supporting `@`-less identifiers on Slack.

I have come up with a more extensive approach to this problem in #534, it would be great if you could try that out and see if you encounter any issues.

sijis:
@gbin it may have some impact, since BOT_ADMINS is a defined in  config.py but parsed by each backend. It's not like each backend has a separate list of ADMINS.

On an unrelated note, been thinking adding a notification flag to an admin entry, since some ADMINS may not care about plugin error or other stuff sent to admins. Those ADMINS needs enough right to do simple admin 'thingy' but not everything.

PR #534, opened 2015-11-04
Slack fixes

### LINKED ISSUE #439: Slack: inconsistency in BOT_ADMINS

### LINKED ISSUE #477: build_identifier("#channel") should be possible on Slack.
First, gotta say; very good job on the bot!

I'm making a plugin that requires me to send a message to a channel, but not to a specific user, on Slack ( would like it to work everywhere tho). I dug through the source and found 2 potential methods: `send` and `bare_send`

`send` will not let me simply send a message without specifying a user, and `bare_send` doesnt seem to work at all on Slack.

The only solution I can think to do is to get a random person from the channel and send it to them as a groupchat.

Not sure if what I want is built in to the bot or not. Any help would be greatly appreciated!

Thanks!

### LINKED ISSUE #533: VersionChecker complains about Unparseable slack identifier
I'm getting the following when err VersionChecker runs:
```
Error: VersionChecker failed to start : Unparseable slack identifier, should be #channelname/username or @username : 'saviles'
```
In config.py i do have the following:
```
BOT_ADMINS = (
            '@saviles',
            'saviles',
            'U03FXXXX',
            )
```

I do think that #528 may be related to this.

zoni:
Quoting from `CHANGES.rst`:

- Slack ACLs now properly check against usernames starting with `@`
- Slack identifiers can now be built from a bare `#channel` string (without a username part)
- Slack identifiers can now be built from a `<#C12345>` or `<@U12345>` string (the webclient formats them like this automatically when chatting with the bot)

Fixes #477 
Fixes #439 
Fixes #533
Closes #528

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -481,9 +481,10 @@ class SlackBackend(ErrBot):
 
         return parts
 
-    def build_identifier(self, txtrep):
+    @staticmethod
+    def extract_identifiers_from_string(text):
         """
-        Build an Identifier from the given string txtrep.
+        Parse a string for Slack user/channel IDs.
 
         Supports strings with the following formats::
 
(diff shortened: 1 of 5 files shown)


errbot/backends/slack.py
```diff
@@ -79,11 +79,11 @@ def __init__(self, *args, error='', **kwargs):
 class SlackIdentifier(DeprecationBridgeIdentifier):
     # TODO(gbin): remove this deprecation warnings at one point.
 
-    def __init__(self, sc, userid, channelid=None):
-        if userid[0] not in ['U', 'B']:
+    def __init__(self, sc, userid=None, channelid=None):
+        if userid is not None and userid[0] not in ('U', 'B'):
             raise Exception('This is not a Slack user or bot id: %s' % userid)
```
gbin: Maybe express the normal forms of a slack id here

gbin:
Maybe making a first class function of it and some simple unittests?
```
foo
foo/bar
@foo/bar
<>
<@>
#foo/bar/baz
#/
```

zoni:
That may be a good idea. I'll see if I have time to work on that tomorrow.

zoni:
@gbin I've added some test coverage. This is the best I could do without investing significantly more time refactoring a bunch of things, that I'm just not willing to do right now.

gbin:
thx !

PR #535, opened 2015-11-07
Type hints on botcmd, re_botcmd and arg_botcmd

gbin:
Also refactor the common tagging part that was copy pasted 3 times.

### DIFF
--- a/errbot/__init__.py
+++ b/errbot/__init__.py
@@ -6,7 +6,7 @@ import logging
 import re
 import shlex
 import inspect
-from typing import Callable, Any
+from typing import Callable, Any, Tuple
 
 from .core_plugins.wsview import bottle_app, WebView
 from errbot.backends.base import Message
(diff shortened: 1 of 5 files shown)


errbot/__init__.py
```diff
@@ -73,25 +120,37 @@ def some_command(self, msg, args):
...
-        return func
-
-    if len(args):
-        return decorate(args[0], **kwargs)
-    else:
-        return lambda func: decorate(func, **kwargs)
-
-
-def re_botcmd(*args, **kwargs):
+    if args:
```
zoni: I'm trying to wrap my head around why this is needed, when this is needed. Is also wonder if we can't do...

``` python
if not args:
   args = args[0]

return _tag_botcmd(bla...)
```

instead, to avoid having to more or less duplicate these two calls.

zoni:
:+1:

PR #538, opened 2015-11-09
Materialized Identifier and MUCIdentifier as ABC

gbin:
With type hinting on the API surface, it makes more sense now to materialize the Identifiers so users can autocomplete/explore them.


zoni:
I finally had the opportunity to take this for a spin and it worked great. :smile: I tested this on Slack and Telegram and encountered no issues on those two backends so I'm quite confident we can safely merge this now.

gbin:
thx @zoni

PR #539, opened 2015-11-09
Documentation fixes, mostly for pop & push_message

anitawoodruff:
Now these are methods on testbot they no longer need to be imported.

Updated the testing documentation in all the places I noticed it was wrong.

(Also couldn't help fixing a couple of typos on the 'contributing.html' page whilst I was reading it :P )

I'm not 100% sure I corrected the FullStackTest docs in test.py correctly - would be grateful if someone could double-check.

Cheers :)

@zoni thanks for the encouragement & offer of help with this!

btw I tried running the tests locally but one failed - utils_tests.test_storage - it fails for me on master too though.. any ideas? (I posted this failure on gitter too)

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -14,12 +14,12 @@ as soon as we can! Depending on our thoughts, we might decide to merge it in
 right away, or we may ask you to change certain parts before we will accept the
 change.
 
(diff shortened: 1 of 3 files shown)


gbin:
Thx for your contribution !

PR #544, opened 2015-11-12
Allow non-standard hipchat server

barakschiller:
Pass in server specified in the config to allow connection to a hosted hipchat server


zoni:
Thanks! This shouldn't actually be required (I think) because your JID should already include the correct domain even with HipChat server (please correct me if I'm wrong), but nontheless we should support this to be consistent with the XMPP backend.

I've merged this but ended up rebasing it in my how-I-deal-with-PRs-at-work mindset so the commit hash changed. :smile: It's in as 959798e1be02cbcfd0afcfc1d1a50ee321c0448b now.

PR #546, opened 2015-11-15
Add bind address and IPv6 support to irc backend

mRB0:
Hand-tested the following cases:

* IPv4 connection
* IPv4 bind_address
* IPv4 SSL
* IPv4 SSL + bind_address
* IPv6 versions of all the above with a couple different bind_addresses. IPv6 is a vhoster's dream.

Didn't update/write any tests, because there were no existing tests for the IRC backend.


mRB0:
Oops, tests fail. Will fix.

PR #551, opened 2015-11-21
Fix broken google plus community link

anitawoodruff:
It's odd - the old links works when I'm *not* signed in to Google Plus, but when I am signed in then I get a 403 error and can't access the page (tested on two accounts). The new link works whether signed in or not.


gbin:
Ha thx so much, this link has been a mystery for so long !

anitawoodruff:
No problem

PR #557, opened 2015-12-01
[WIP] Add script to add config.py from installation

sijis:
I copied the oneliner python code from the README as the basis for this script script. This could possibly, also be the basis for #462 and be more 'interactive'.


sijis:
@gbin if the reference PR is considered scope creep does this one fall in the same category?

gbin:
Not really, but I want to do that comprehensively with the bot data dir etc.. This will be the last part for the general "provisioning" for the bot.

Issue #558, opened 2015-12-03
XMPP backend does not ignore its own messages.


brycied00d:
This appears to be a regression of #99 and #247. Debug output below shows SEND followed by RECV back from the server and subsequent processing and matching of the RECV message.

```
00:23:45 DEBUG    MyTest     Bryce Chidester has just updated issue \#3043 "Test Issue"
00:23:45 DEBUG    errbot.errBot             Trigger callback_message on Webserver
00:23:45 DEBUG    errbot.errBot             Trigger callback_message on Plugins
00:23:45 DEBUG    errbot.backends.xmpp      This is a groupchat message, strip the resource.
00:23:45 DEBUG    errbot.backends.xmpp      send_message to firehose@conference.company.local
00:23:45 DEBUG    errbot.backends.xmpp      build reply ...
00:23:45 DEBUG    sleekxmpp.xmlstream.xmlst SEND: <message type="groupchat" to="firehose@conference.company.local" xml:lang="en"><body>Bryce Chidester has just updated issue #3043 &quot;Test Issue&quot;</body></message>
00:23:45 DEBUG    errbot.backends.xmpp      send_message to bryce.chidester@company.local/Pidgin_two
00:23:45 DEBUG    sleekxmpp.xmlstream.xmlst RECV: <message from="firehose@conference.company.local/ErrBot" type="groupchat" to="errbot@company.local/Bot" xml:lang="en"><body>Bryce Chidester has just updated issue #3043 &quot;Test Issue&quot;</body></message>
00:23:45 DEBUG    sleekxmpp.xmlstream.xmlst Event triggered: message
00:23:45 DEBUG    sleekxmpp.xmlstream.xmlst Event triggered: groupchat_message
00:23:45 DEBUG    sleekxmpp.xmlstream.xmlst Event triggered: muc::firehose@conference.company.local::message
00:23:45 DEBUG    errbot.backends.xmpp      incoming_message frm : firehose@conference.company.local/ErrBot
00:23:45 DEBUG    errbot.backends.xmpp      incoming_message frm node: firehose
00:23:45 DEBUG    errbot.backends.xmpp      incoming_message frm domain: conference.company.local
00:23:45 DEBUG    errbot.backends.xmpp      incoming_message frm resource: ErrBot
00:23:45 DEBUG    errbot.errBot             *** frm = firehose@conference.company.local/ErrBot
00:23:45 DEBUG    errbot.errBot             *** username = firehose@conference.company.local/ErrBot
[...]

PR #560, opened 2015-12-04
Self check correction on XMPP.

gbin:


gbin:
Fixes #558

zoni:
Gets my :+1:

brycied00d:
:+1: tested and works for me.

PR #562, opened 2015-12-08
Splice '.' when adding ':' to MSG_ERROR_OCCURRED

Cbeck527:
The resulting message in chat was `See log for details.:` which is
grammatically incorrect. Other strings have periods, so to remain
consistent it made sense to splice the message before sending.

image


gbin:
It broke the tests, if the string is only used in one place, I would just remove the . from it :)

Cbeck527:
Yep, just saw that. Will close and re-open with the new solution. Thanks!

PR #565, opened 2015-12-15
Use the NickMask helper for parsing IRC Identity and proper ACL

xintron:
The irc-library provides a good helper class to parse the nick-mask for
irc identities. This change simply wraps that helper.

ACL is verified against the user + host (which is guaranteed to be more
unique and harder to spoof) instead of the current nickname of the user.
This also allows connected admins to change their nickname however they
see fit and still be able to manage the bot.


gbin:
Thx !

mrshu:
:+1: I wanted to fix this for long long time!

PR #566, opened 2015-12-16
[WIP] Feature/plugin registry

sijis:
I'm just hacking away an idea to use a simple json file to get a 'realtime' pull of available plugins. The json is based on the script in `tools/` directory.

I changed the format of the plugin to be `<owner>/<plugin-name>`. The only reason I did that was in the even there were two plugins with the same name. In our case, we have multiple forks of the same plugin. Its possible the bot owner may want to try all of them and figure out the best one of the batch.

I did realize there is an `err-catalog` plugin. I honestly haven't checked it.

### DIFF
--- a/errbot/core_plugins/plugins.py
+++ b/errbot/core_plugins/plugins.py
@@ -45,7 +45,21 @@ class Plugins(BotPlugin):
         """
         if not args.strip():
             return "You should have a repo name as argument"
-        repos = self._bot.get(self._bot.REPOS, {})
+        repos = {}
+        _installed = self._bot.get_installed_plugin_repos()
+
+        # Fix to migrate exiting plugins into new format
+        for short_name, url in _installed.items():
+            name = ('/'.join(url.split('/')[-2:])).replace('.git', '')
+
+            t_installed = {name: {
+                'path': url,
+                'documentation': 'Unavilable',
+                'python': None,
+                'avatar_url': None,
+                }
+            }
+            repos.update(t_installed)
         if args not in repos:
             return "This repo is not installed check with " + self._bot.prefix + "repos the list of installed ones"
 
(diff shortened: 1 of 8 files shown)


sijis:
I've only been able to partially test it locally and i haven't invested in the tests/ cases yet.

gbin:
This is a good idea, I think we should generate the json file at the same time as the wiki page. some tests would be nice too indeed.

sijis:
Yeah, my idea was the data that generates the wiki page is the same set used for the plugins.json file (or whatever we end up calling it).

sijis:
I couldn't add a file to the wiki, so I dumped the contents of the json in a [gist](https://gist.githubusercontent.com/sijis/95ac411600f085818f22/raw/84ef97b1fbc5c00ebf46c1558d6fd83c07091e6a/repos.json).
I forgot but the example json file is hosted locally on my machine. I didn't want to commit some oddball url for testing, so i left it at localhost.

I'll work on the scripts in tools/ to generate this properly.

sijis:
OK. I think this is ready for a review. Everything needed for this should be good now (except for a final location of `repos.json`).

gbin:
can you check the broken tests ?

errbot/core_plugins/plugins.py
```diff
@@ -69,13 +70,33 @@ def repos_uninstall(self, mess, args):
...
+
+            installed = False
+            public = False
+
+            try:
+                description = KNOWN_PUBLIC_REPOS[repo_name].get('documentation', 'Unavailable')
+            except KeyError:
+                description = installed_repos[repo_name].get('path')
+
+            if repo_name in KNOWN_PUBLIC_REPOS:
```
gbin: public = repo_name in KNOWN_PUBLIC_REPOS

sijis: I'm not sure what you are referring to here? Are you suggesting those lines could be simplified with that single `public = repo_name in KNOWN_PUBLIC_REPOS` line?

sijis:
In the tests commit, i modified `test_plugin_cycle` and tried to include a loop to keep reinstalling the err-helloworld via several methods:
1. using the git url
2. using the 'registry' name
3. using a tar.gz (not included)

However, I run into a problem which i'm unable to solve. The first loop (regardless of method) will always work. The second iteration will fail with either a `storage open` error or unable to find the `!hello` command. I think i've caused a problem with how the plugin system recognizes the plugins but i don't know where hat is set.

PR #567, opened 2015-12-22
Adapt to new slackclient interface

peterdemin:
In [PR-54](https://github.com/slackhq/python-slackclient/pull/54) slackclient is changing return type of api_call, simplifying it's usage.

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -1,4 +1,3 @@
-import json
 import logging
 import re
 import time
@@ -66,7 +65,7 @@ USER_IS_BOT_HELPTEXT = (
 class SlackAPIResponseError(RuntimeError):
     """Slack API returned a non-OK response"""
 
-    def __init__(self, *args, error='', **kwargs):
+    def __init__(self, error='', *args, **kwargs):
         """
         :param error:
             The 'error' key from the API response data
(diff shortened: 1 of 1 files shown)


errbot/backends/slack.py
```diff
@@ -200,7 +199,7 @@ def api_call(self, method, data=None, raise_errors=True):
         """
         if data is None:
             data = {}
-        response = json.loads(self.sc.server.api_call(method, **data).decode('utf-8'))
+        response = self.sc.server.api_call(method, **data).json()
```
zoni: In the referenced PR, [line 135](https://github.com/slackhq/python-slackclient/pull/54/files#diff-599dbee7d2139b2e66b89a120d50a0a1R135) returns `Response.text`, which returns a unicode object. This means it should still be decoded as `utf-8` and fed to `json.loads` (there's no `.json()` available, that's only a function on the original `Response` object).

That means there should be no changes necessary on our side, which in my opinion is how it should be. You don't want to change the public API, as that might break a lot of users of library.

gbin:
It doesn't look upstream has been merged, what is the status about this ?

peterdemin:
There was no action in python-slackclient since Nov 25.
However in response to zoni's comment I think, I should change PR-54 to keep public API the same.

zoni:
> However in response to zoni's comment I think, I should change PR-54 to keep public API the same.

Looking at the diff of https://github.com/slackhq/python-slackclient/pull/54, nothing about the public API changes though.

peterdemin:
`SlackRequest.do` was returning `urlopen()` result, and I changed it to `requests.post()` result that is not compatible.

gbin:
yes it brakes:
   response = json.loads(self.sc.server.api_call(method, **data).decode('utf-8'))
AttributeError: 'str' object has no attribute 'decode'

gbin:
but yours too -_-
    response = self.sc.server.api_call(method, **data).json()
AttributeError: 'str' object has no attribute 'json'

gbin:
Pushed the fix directly on master but it still completely fails for some identity issue.

PR #570, opened 2015-12-27
irc: Fix random `UnicodeDecodeError`s

### LINKED ISSUE #569: Ignore all input that cannot be decoded
Hello. I am getting an UnicodeDecodeError randomly on irc. I found a solution at https://pypi.python.org/pypi/irc. But I don't know how to apply it to errbot.

mrshu:
* In #569 it was reported that `UnicodeDecodeError`s are thrown randomly
  when the IRC backend is used. This commit tries to resolve this
  situation by using a replacement character for unrecognized byte
  sequences as described at https://pypi.python.org/pypi/irc

---------------------------------------------------

Note that this should fix #569.

### DIFF
--- a/errbot/backends/irc.py
+++ b/errbot/backends/irc.py
@@ -304,6 +304,11 @@ class IRCConnection(SingleServerIRCBot):
         super().__init__([(server, port, password)], nickname, username, reconnection_interval=reconnect_on_disconnect)
 
     def connect(self, *args, **kwargs):
+        # Decode all input to UTF-8, but use a replacement character for
+        # unrecognized byte sequences
+        # (as described at https://pypi.python.org/pypi/irc)
+        self.connection.buffer_class.errors = 'replace'
+
         connection_factory_kwargs = {}
         if self.use_ssl:
             import ssl


errbot/backends/irc.py
```diff
@@ -304,6 +304,11 @@ def __init__(self,
         super().__init__([(server, port, password)], nickname, username, reconnection_interval=reconnect_on_disconnect)
 
     def connect(self, *args, **kwargs):
+        # Decode all input to UTF-8, but use a replacement character for
+        # unrecognized byte sequences
+        # (as described at https://pypi.python.org/pypi/irc)
+        self.connection.buffer_class.errors = 'replace'
```
mrshu: @perlop Would you find this sufficient? I realize that your issue #569 suggested that all input that is impossible to decode should be ignored -- this seems like a bit more robust solution.

zoni:
I think this is a sane thing to do. With no way to reliably determine the used encoding, assuming utf-8 seems the best bet as it is slowly becoming the de-facto encoding used for interoperability.

PR #571, opened 2015-12-28
Added automated tests for mentions.

gbin:


zoni:
It looks like the tests are failing on Python 2. Doesn't appear to be the occasional flaky test, I restarted the build twice but it fails consistently on the same test, so I think there's really a bug there.

Issue #572, opened 2015-12-29
Mention callback for bundled backends tracking bug


gbin:
Implement the callback_mention on the following backends:

- [x] Text
- [x] Graphic
- [x] Slack
- [x] Hipchat
- [x] IRC
- [x] Telegram
- [ ] XMPP

mrshu:
@gbin I'll take IRC.

qlixed:
mrshu, please use the on_* events from the irc module. I'm fixing some of the current callbacks code and use them for better reliability.

mrshu:
@qlixed well, i added the mention callback for IRC already in the PR mentioned above. Does the IRC lib support some sort of `on_` callback for mention?

qlixed:
Oh, gosh, sorry, for Mention there is no on_* event. My bad! (long day)

mrshu:
No problem @qlixed, please keep the questions coming!

zoni:
Going to mark Telegram as "done" because it's not currently possible to implement with the limited API available to bots.

qlixed:
Callback mention on IRC don't work properly please see #681

qlixed:
The mention Callback Is fixed for IRC

gbin:
Hipchat is done, XMPP seems to have something for mentions in an obscure XEP probably nobody implements, I'll skip it for now. If somebody really needs this please open another bug.

PR #574, opened 2015-12-29
irc: Add mention callback

mrshu:
* Add mention callback for the IRC backend by exctracting all possible
  IRC nicknames from a message and then checking them against the list
  of users in the room this message originated from.

### DIFF
--- a/errbot/backends/irc.py
+++ b/errbot/backends/irc.py
@@ -4,6 +4,7 @@ import sys
 import threading
 import subprocess
 import struct
+import re
 
 from markdown import Markdown
 from markdown.extensions.extra import ExtraExtension
@@ -47,6 +48,8 @@ IRC_CHRS = CharacterTable(fg_black=NSC('\x0301'),
                           end_fixed_width='',
                           inline_code='',
                           end_inline_code='')
+
+IRC_NICK_REGEX = r'[a-zA-Z\[\]\\`_\^\{\|\}][a-zA-Z0-9\[\]\\`_\^\{\|\}-]+'
 
 try:
     import irc.connection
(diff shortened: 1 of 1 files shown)


mrshu:
to @gbin 

Note this partially fixes #572

errbot/backends/irc.py
```diff
@@ -343,6 +346,12 @@ def on_pubmsg(self, _, e):
         msg.nick = msg.frm.nick  # FIXME find the real nick in the channel
         self.callback.callback_message(msg)
 
+        possible_mentions = re.findall(IRC_NICK_REGEX, e.arguments[0])
+        room_users = self.channels[room].users()
+        mentions = filter(lambda x: x in room_users, possible_mentions)
```
gbin: I think you need to do:
mentions = filter(lambda self.build_identifier(x): x in room_users, possible_mentions)

Otherwise it will inject strings to the plugins instead of identifiers

errbot/backends/irc.py
```diff
@@ -343,6 +346,12 @@ def on_pubmsg(self, _, e):
...
+        if mentions:
+            self.callback.callback_mention(msg, mentions)
```
gbin: self.callback.callback_mention(msg, [self.build_identifier(mention) for mention in mentions]) to pass IRCIdentifiers to the plugins instead of strings no ?

mrshu: @gbin you are absolutely right, will fix

mrshu:
@gbin I am also wondering what tests should I write for this. Unfortunately there does not seem to be one fore the IRC backend...

gbin:
Yes, we have a gap in our testing on the backends. Mocking seems to be the only solution but this gonna be very heavy to implement...

gbin:
thx

PR #578, opened 2016-01-05
Fixes join when password is set for room.

mslehto:
_join_room() was expecting compat_str() to bail out with None.
That is not the case anymore (commit 4ed92ff24bf3c7917c9f46b88d800a144aaaaf24).
Therefore room JID was junk and password enabled XMPP MUC rooms could not be joined.

### DIFF
--- a/errbot/core_plugins/chatRoom.py
+++ b/errbot/core_plugins/chatRoom.py
@@ -33,7 +33,7 @@ class ChatRoom(BotPlugin):
                     self.log.exception("Joining room %s failed", repr(room))
 
     def _join_room(self, room):
-        if isinstance(room, tuple):
+        if isinstance(room, tuple) or isinstance(room, list):
             room_name = compat_str(room[0])
             room_password = compat_str(room[1])
             room, username, password = (room_name, self.bot_config.CHATROOM_FN, room_password)
--- a/errbot/core_plugins/chatRoom.py
+++ b/errbot/core_plugins/chatRoom.py
@@ -33,12 +33,15 @@ class ChatRoom(BotPlugin):
                     self.log.exception("Joining room %s failed", repr(room))
 
     def _join_room(self, room):
-        room_name = compat_str(room)
-        if room_name is not None:
-            room, username, password = (room_name, self.bot_config.CHATROOM_FN, None)
+        if isinstance(room, tuple):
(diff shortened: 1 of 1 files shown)


errbot/core_plugins/chatRoom.py
```diff
@@ -33,12 +33,15 @@ def callback_connect(self):
                     self.log.exception("Joining room %s failed", repr(room))
 
     def _join_room(self, room):
-        room_name = compat_str(room)
-        if room_name is not None:
-            room, username, password = (room_name, self.bot_config.CHATROOM_FN, None)
+        if isinstance(room, tuple):
```
zoni: Checking for just tuples won't be sufficient here as a user could be using lists as well, which would give the same behaviour.

zoni:
Damn, I thought I'd checked all the callers of that function so carefully :disappointed:

errbot/core_plugins/chatRoom.py
mslehto: Ah, ok. I saw only tuples in config template so that's what I thought can ever exist:
https://github.com/errbotio/errbot/blob/master/errbot/config-template.py#L149

mslehto:
@zoni, is commit cc3d4d95bc70e6e2a85b0c761f40dc89d6e8e506 addressing your minor comment?

errbot/core_plugins/chatRoom.py
```diff
@@ -33,12 +33,15 @@ def callback_connect(self):
                     self.log.exception("Joining room %s failed", repr(room))
 
     def _join_room(self, room):
-        room_name = compat_str(room)
-        if room_name is not None:
-            room, username, password = (room_name, self.bot_config.CHATROOM_FN, None)
+        if isinstance(room, tuple) or isinstance(room, list):
```
zoni: This can be written as `if isinstance(room, (list, tuple)):`

zoni:
@mslehto I'm going to merge when I have time tomorrow (go ahead and poke me again if I fail to).

I left a minor comment but don't worry about it, I'll change it myself when I merge it so you don't have to bother with such a trivial change yourself.

zoni:
See fd23ab7cd67f7b070bcb9d4bc352d5c8428a4f47 :smile: Thanks!

PR #579, opened 2016-01-11
Changes proposed for the config file template for errbot

fernand0:
The idea is to be able to store config.py passwords in the system's keyring, if available.
The code has been commented in order to not to cause any problems in the keyring module is not available. It needs some manual action (setting up the password in the keyring; it can be also automated but I'm not sure if we will want to add more code). 

I've changed the way username and password are stored in the file in order to minimize the changes needed to add this new 'feature'.
Notice that this is not a significative security improvement: if you are logged in the system the password is available to you, anyway. At least, it is hidden for the casual observer.

### DIFF
--- a/errbot/config-template.py
+++ b/errbot/config-template.py
@@ -17,6 +17,9 @@
 ##########################################################################
 
 import logging
+# import keyring
+# You'll need to have the relevant software in your systems. Instructions at:
+# https://pypi.python.org/pypi/keyring
 
 ##########################################################################
 # Core Err configuration                                                 #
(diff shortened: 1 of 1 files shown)


zoni:
I'm a little bit torn on including all these instructions in `config.py` (there's already a lot of stuff in there to go through). I think detailed instructions as (an optional) part of the userguide on http://errbot.io/ may be a better fit, with perhaps a simple one-line comment along the lines of *"# See http://errbot.io/user_guide/some/page/ for instructions on how to use the system-keyring to store passwords instead of configuring them as plain-text here"* in `config.py`.

What are your thoughts on this approach?

fernand0:
Probably you'are true. This proposal complicates things a bit (and it makes the install process a bit longer). Maybe your proposal is adequate if somebody is interested in this small security improvement. In fact, I posted the idea in the G+ community just with the idea of having some comments about the proposal and, maybe, helping others to thin about these questions.
For me it is ok any solution you finally adopt.

gbin:
As a second thought about this I agree with@zoni, this is an awesome idea and it should be documented, maybe not upfront on the config.py template, can you move that to a doc page with an example ?

fernand0:
Not sure if it's me who has to move this to a doc page. If yes, not sure about how to do this...

gbin:
@fernand0 if you can it would be awesome and helpful !

fernand0:
Hello again,

I'm not trying to argue with you but in [Configurarion](http://errbot.io/user_guide/setup.html#configuration) section it says 'Read the documentation within this file and edit the values as needed so the bot can connect to your favorite chat server.' I'm not sure where this text could go (and I'm worried about my not so good English). Anyway, my proposal. There can be also problems of style and so on.... I can work more on this if you provide some guidance.

The proposed text:

==================
In the standard config.py file you have to include your credentials for the backend selected ('username' and 'password'). Some of you may not feel comfortable storing your credentials in the config file so you can use an alternative approach, based on the [keyring](https://pypi.python.org/pypi/keyring) module. Provided you have it correctly installed, you can store credentials there. 
You can see there the way to store your credentials and so on, we will not replicate them here.

Notice that the credentials storage will rely on the operating system and depending on the configuration anybody logged into your account will have access to them without password (you'll need this if you want a bot that can start in an unattended way). We are only obtaining the added security of not having the credentials stored in the config.py file but not much more.

Then you need to make some changes in the config.py file. First of all, importing the module:

{% highlight python %}
import keyring
{% endhighlight %}

Later, before the BOT_IDENTITY section you can add three variables, for the server (needed in order to select the account in the keyring), username and password:
 
{% highlight python %}
server = 'jabber'
username = myUserName
password = keyring.get_password(server,username)
{% endhighlight %}

Finally, in the BOT_IDENTITY structure, in the backend you have selected, you
can put (XMPP backend, for example): 

{% highlight python %}
    'username': username,  # The JID of the user you have created for the bot
    'password': password,       # The corresponding password for this user
{% endhighlight %}

In this way when the bot starts gets the credential from the keyring.

gbin:
Thx for the blog post about it !

PR #583, opened 2016-01-17
Implements a filter for core plugins.

gbin:
Some bundled plugins are not useful for everybody and creates noise in
the Help system and potential security issues if they are not configured
correctly.

This adds a CORE_PLUGINS entry to the config.py so users can select the
set of core plugins they need at config time.

If it is None or unspecified, all the coreplugins are started.


sijis:
Tested this on python 3.4.2 and it worked perfectly.

Issue #575, opened 2016-01-04
Built-in backends should allow for easier requirement installation


asfaltboy:
Installing plugins and external backends can automatically install requirements for use (when setting `AUTOINSTALL_DEPS=True`). It should be equally as easy to get requirements for built-in backends and these are not installed by default when installing `errbot`.

I reccomend adding an `extras_require` section in in `setup.py` for each backend that is included with errbot; so taht users can do `pip install errbot[telegram]` and not worry about error messages later.

zoni:
As discussed on Gitter, I fully agree we should implement this.

gbin:
Made it quickly, I wonder how we can test it with "pip install -e ." ?

zoni:
I could be mistaken, but I don't think it can be tested like that... Maybe upload a test package to the PyPIi testserver?

asfaltboy:
`pip install -e .` supports the same context as `pip install`, so extras may be installed just the same:

    # install only a specific extra (+ general deps)
    $ pip install -e .[hipchat]
    
    # install multiple extras
    $ pip install -e .[graphic,hipchat,IRC,slack,telegram,XMPP]

Not sure if this is what @gbin meant in "test", but if automated testing is required, we can probably write a test which asserts that running the setup and importing those required deps doesn't raise a `ImportError` (though, personally, I think this is a `pip/setuptools` feature, and does not require further testing).

gbin:
Weird, test as in, locally testing at least :) it didn't work when I tried, maybe a zsh thing I gonna try again. Thanks for the tips !

asfaltboy:
It would only work if we have an extra named "test", like [here](
https://github.com/mapbox/mapbox-sdk-py/blob/master/setup.py). It
definitely should work, regardless of shell, myself I use it in multiple
projects.

Can you specify how it fails to work for you? Any issue you're seeing?

zoni:
Implemented in #584. Should anyone find their way to this issue in the future, if you're seeing any issues, please open a new issue.

asfaltboy:
Thanks for the fix! But guys, shouldn't we mention it in docs though?

Probably best to preface the ["Extra dependencies"][1] section, or even postfix the ["Installation"][2] section to be more user friendly.

[1]: http://errbot.io/user_guide/setup.html#extra-dependencies
[2]: http://errbot.io/user_guide/setup.html#installation

zoni:
[...]

PR #584, opened 2016-01-18
Added optional dependencies for the backends.

gbin:


gbin:
Fixes #575

setup.py
```diff
@@ -120,6 +120,14 @@ def read(fname):
         package_data={
             '': ['*.txt', '*.rst', '*.plug', '*.md'],
         },
+        extras_require={
+            'graphic':  ['PySide', ],
+            'hipchat': ['hypchat', 'sleekxmpp', ],
+            'IRC': ['irc', ],
+            'slack': ['slackclient', ],
+            'telegram': ['python-telegram-bot', ],
+            'XMPP': ['sleekxmpp', ],
```
zoni: sleekxmpp needs the `pyasn1` and `pyasn1-modules` in order to do certificate validations but doesn't have a hard dependency on them, so we should add those too.

setup.py
```diff
@@ -120,6 +120,14 @@ def read(fname):
         package_data={
             '': ['*.txt', '*.rst', '*.plug', '*.md'],
         },
+        extras_require={
+            'graphic':  ['PySide', ],
+            'hipchat': ['hypchat', 'sleekxmpp', ],
```
zoni: sleekxmpp needs the `pyasn1` and `pyasn1-modules` in order to do certificate validations but doesn't have a hard dependency on them, so we should add those too.

setup.py
gbin: Done

zoni:
Gets my :+1:

PR #588, opened 2016-01-20
Enforce storage key is a str

ethan-eb:
In cases where a unicode key is provided, this was breaking

### DIFF
--- a/errbot/storage.py
+++ b/errbot/storage.py
@@ -69,4 +69,4 @@ class StoreMixin(MutableMapping):
             yield i
 
     def __contains__(self, x):
-        return x in self.shelf
+        return str(x) in self.shelf


gbin:
It apparently breaks the unit tests, can you check out what is going on ?

gbin:
def __contains__(self, key):
>       return key in self.dict
E       TypeError: gdbm key must be string, not unicode

ethan-eb:
Yeah, I'll have a look - also just found in my python 2.7 pip installed errbot str() turned into unicode() ...

ethan-eb:
confirmed that's what 3to2 is doing:

    $ 3to2 -v errbot/storage.py
    ...
         def __contains__(self, x):
    -        return str(x) in self.shelf
    +        return unicode(x) in self.shelf

gbin:
maybe:
```
if PY2 and type(x) == str:
   x = x.encode('utf-8')
return x in self.shelf
```

ethan-eb:
whoa, a lot of changes have happened, is this fixed now?

gbin:
No I don't think so, the issue should remain, it just moved to storage/ :)

ethan-eb:
Ok, I'm going to re-try this as a different PR today ... I need to pull in the latest first

This is actually the fix that got me going https://github.com/ethanmiller/errbot/commit/113b1d03fd1339815dddf4d5429e8511953843c2 because 3to2 does this:

    -        return bytes(x) in self.shelf
    +        return str(x) in self.shelf

But I'm not sure if that will work for tests. I will try.

ethan-eb:
I was not able to reproduce the issue with the latest code

Issue #377, opened 2015-06-05
Template file names must be unique.


brycied00d:
Through trial and error, I have discovered that template file names must be unique.

If I have two plugins, both with their own "templates/issue.html", which "issue.html" that gets used by a given plugin seems to be random or depend on the fill order of some internal list. Naturally this causes problems because one plugin's template is not compatible with another.

I don't see this as an urgent bug in err itself, but rather a note for the documentation.

zoni:
I'm labelling this as a bug, as it's not unthinkable for two plugins to use the same template names. Without having looked at the code in question, I suspect this happens because we just use a single Jinja template loader and add the template path for each plugin to it. Then we get a single loader with a big global list of paths, the order of which is simply dependent on the order in which plugins were discovered by Yapsy, the plugin system.

Long term, I think we could try to look up templates only for the plugin that the botcmd is a part of, perhaps falling back to looking across all plugins the way it does now when that isn't found (to maintain backward-compatibility).

Short term, I'll add a warning to the documentation and link back to this issue as it's not a very urgent bug and not likely to get changed soon unless someone else steps up to tackle the problem.

TomNeyland-FD:
FWIW Django has this same issue and they just make you name your files/subdirectories in a manner that avoids a namespace clash.

pendragon-:
Depending on whether you are willing to transport the plugin path together with the template name on the function, there is no need to create different loaders. `get_template` accepts a `parent` parameter.
From the documentation I assume that plugin_dir + template subdir will make this pick up the one correct template. I'd be willing to give this a go if you are interested.

gbin:
@pendragon- sure give it a shot !

asfaltboy:
Just looked at this while working on #876 and seems by default [`jinja2.Environment` does not use the `parent` argument][1].

Perhaps the simplest solution would be to subclass [`jinja2.BaseLoader`][2] (or `FileSystemLoader`) and have it's `get_source` method read a naming convention such as `<plugin_dir>.<template_path>` where `plugin_dir` is an absolute path, and `template_path` is relative to the plugin's `templates` dir.

[1]: https://github.com/pallets/jinja/blob/master/jinja2/environment.py#L757
[2]: http://jinja.pocoo.org/docs/dev/api/#loaders

sijis:
New release created https://pypi.org/project/errbot/6.2.1/ which should include this fix.

PR #598, opened 2016-02-11
only use templates belonging to a plugin

boidolr:
This is a proof of concept for #377. Using `inspect` feels like a hack - I didn't see any way to get to the directory a plugin is located in withouth explicitly adding it (it is present in in `plugin_manager` as `pta_item.path` though).
To be consistent, `send_templated` has to be extended, too.

Let me know whether this change is something you want to include or if there's something which should be done differently. I'll look into adding tests when I get some more time.


gbin:
cool, yapsy has the path to the plugin already in http://yapsy.sourceforge.net/PluginInfo.html#yapsy.PluginInfo.PluginInfo.path

We actually add the template path from this parameter here: https://github.com/errbotio/errbot/blob/1acb24799a10723e7b148d00e829a4cbf8fa934f/errbot/plugin_manager.py#L290


Now we need to clean way to inject that in a private member of the public right at instantiation there maybe ?

pendragon-:
Yes, that's what I was considering, too. I'll either set it right after getting the `plugin_object`[here](https://github.com/errbotio/errbot/blob/1acb24799a10723e7b148d00e829a4cbf8fa934f/errbot/plugin_manager.py#L275) or at the place where the templates are added (the one you refer too). I prefer the second option, as then the related code is closer together.
Probably something like `obj._directory = pta_item.path` would do the trick.

gbin:
@pendragon- sound like a good plan !

errbot/templating.py
```diff
@@ -26,24 +27,37 @@ def make_templates_from_plugin_path(plugin_path):
...
-def add_plugin_templates_path(path):
+def add_plugin_templates_path(path, identifier=None):
     global env
     tmpl_path = make_templates_from_plugin_path(path)
     if os.path.exists(tmpl_path):
         log.debug("Templates directory found for this plugin [%s]" % tmpl_path)
+        plugin_templates = [tmpl_path, system_templates_path]
+        if identifier is not None:
+            plugin_paths[identifier] = FileSystemLoader(plugin_templates)
         template_path.append(tmpl_path)  # for webhooks
```
pendragon-: I think adding to `template_path` might be unnecessary. It keeps everything backwards compatible though.

pendragon-:
The `Environment#get_template` didn't really work as I initially thought - it usually does not add up the parts. Additionally, the main plugins have templates which are not in a template subdirectory and therefore, not picked up if only these locations are added and then searched by prefixed path.

gbin:
Ok, this is not good :)

Can't we move the templating to the plugin itself then ? Here it does:
Errbot [with templating info] -> plugin [without templating implementation] -> Errbot [with both]

Can't we transform that to ?
Errbot -> plugin [with both]

pendragon-:
I am unsure what exactly this would entail. Are all the built-in methods wrapped in plugins right now (health, status, plugin management, ...)? That not all templates were in the correct subdirectory is what made the fallback necessary. Also, as there is the method n the bot itself https://github.com/errbotio/errbot/blob/master/errbot/errBot.py#L172 there always might be a 'wrong' access of templates.

errbot/errBot.py
```diff
@@ -417,10 +424,12 @@ def _process_command(self, mess, cmd, args, match):
                                    template_name=f._err_command_template)
 
     @staticmethod
-    def process_template(template_name, template_parameters):
+    def process_template(template_name, template_parameters, plugin_name):
```
pendragon-: you'd move this part into plugin instead of bot? Then it would be possible to call it on `f.__self__` below.
Then the question is if the method here (or rather, send_templated) would be needed still.

pendragon-:
I'll try again by storing templates with the plugins itself. As a consequence, the as of yet unreleased `BotPlugin#send_templated` would need to be removed though - or at least, would need to use some sort of handle on the different environments for templating.

gbin:
it is ok, the bot one is a private API, it can disappear. thx !

pendragon-:
Ok, what about this?
I am still trying to figure out how to adapt `base_backend_tests.py` though (as the commands now have to be in a plugin, not a backend).

gbin:
we need at least a common directory between them all, for example we plan to put a feature to generate the tables in a jinja2 macro from an include.
So maybe the tests could change this kind of system path ?

pendragon-:
So you need to have a common directory included in all environments? That's actually already the case - as the environment is still built in a central place, it's easy to include a fixed directory there.
My issue is more with sideloading some sort of `DummyPlugin` into the `DummyBackend` used in that particular test.

pendragon-:
Is there a reason to use `DummyBackend` instead of the testbot? Having the different kinds of commands in a separate plugin (with in `dummy_plugin`?) would be a lot easier. The fixture/the test bot allows injecting specific plugins already.

pendragon-:
@gbin / @zoni any hint on how to proceed with adapting the test case?
If the fixture can not be used, a lot of workaround code would need to be added for `base_backend_tests.py`, duplicating the code in the bot & plugin setup itself.

tests/base_backend_tests.py
```diff
@@ -257,15 +258,20 @@ def test_BOT_ADMINS_unique_string(self):
...
         self.example_message = self.dummy.build_message('some_message')
         self.example_message.frm = self.dummy.build_identifier('noterr')
         self.example_message.to = self.dummy.build_identifier('err')
 
-        assets_path = os.path.join(os.path.dirname(__file__), 'assets')
-        templating.template_path.append(templating.make_templates_path(assets_path))
-        templating.env = templating.Environment(loader=templating.FileSystemLoader(templating.template_path))
+        duplicates_path = os.path.join(assets_path, 'duplicate_template')
+
+        self.tenv = templating.add_plugin_templates_path(assets_path)
```
gbin: assets_path it not in the scope.

gbin:
I tried to catch up with all the merge conflicts but the tests are broken all over the place.
I am a little bit lost, if the change doesn't touch the external plugin API + removed the internal backend template processing the tests fixes should stay minimal no ?

boidolr:
Not for `base_backend_tests.py`. There's a second implementation of the backend - `DummyBackend` -  which also does plugin loading & activation manually (as opposed to relying on the already implemented mechanisms).
At the same time, it acts as a plugin so that the different `botcmd` variants can be tested. And that's where this change here comes in as template handling is tested as well. Instead of these workarounds, it would be easier to either extend the existing `DummyPlugin` with these commands, or write a new one and inject it via the test fixtures. In both cases, the new plugin would be initialized correctly and get it's own environment to resolve it's templates.

tests/base_backend_tests.py
boidolr: I've stopped fixing the test when I found that I'd have to reimplement all the existing logic to make it work.

boidolr:
I've rebased, so it should be easier to get what I am talking about. Note that with the changes in master, dispatching to the plugin in error cases currently doesn't look quite right, and the test still won't run.
Basically the problem is the manual injection of commands in https://github.com/boidolr/errbot/blob/plugin_specific_templates/tests/base_backend_tests.py#L111

boidolr:
This doesn't seem to be going anywhere, and was a bigger change than I anticipated anyway.
If you want to use any of these changes, you are welcome to do so - I am not going to invest more time here though.

gbin:
@boidolr thx ! We gonna revisit that, it is needed and this work helped pinpointing the issues.

PR #599, opened 2016-02-14
err.py: Add version parameter

mrshu:
* Add version parameter to the err.py (errbot) script.

* Do not discriminate in the ArgumentParser's description (errbot is not
  just XMPP bot).


gbin:
Funny this really really old leftover :)

PR #609, opened 2016-02-21
Support globbing on BOT_ADMINS and ACCESS_CONTROLS

zoni:
Title says it all. See modified tests for examples.

Regarding the changelog: I put this in to ensure its not forgotten when a new release is made, and set the version to 9.9.9 because 3.2.0 has already been released. Some of those items are incorrect though (already in 3.2.0) so we'll have to doublecheck the changelog when making a release.

Maybe merge this file with the one from the 3.2 branch?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,4 +1,4 @@
-v3.2.0 (unreleased)
+v9.9.9 (unreleased)
 -------------------
 
 features:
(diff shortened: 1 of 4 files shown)


gbin:
Usually I don't update the ChangeLog until I make a release, then I review all the commits between the releases. It is way easier otherwise you have to "backport" the changelog.

gbin:
thx

Djiit:
Works fine when I change `~djiit@mysuperdynamichost` to  `*djiit@*`

PR #597, opened 2016-02-11
IRC: Synchronize join and joined events.

gbin:
Fixes #595

PR #611, opened 2016-02-21
Fix IRC channel join hang on startup

zoni:
The IRC backend is currently hanging on startup when it tries to join
the rooms, because the join room call blocks other processing of server
events somehow (regression introduced by #597).

It eventually times out raising an exception about not being able to
join the channel, then continues to run the rest of bot startup.

When run in it's own thread, the bot successfully connects without
issue and properly joins channels.


gbin:
argh yeah we created a deadlock

gbin:
Let's do that for now, I hope it won't create race conditions all over the place

Issue #595, opened 2016-02-11
IRC Events are not synchronous.


qlixed:
Trying to get info from the joined channels I found out on python3 that the code that fires up the event callbacks is not synchronous, so the event fires weirdly like we can see here in this log:

> [qlixed@dhcp-0-118 report-errbot]$ vi plugins-extra/test_irc_sync.py 
> [qlixed@dhcp-0-118 report-errbot]$ err.py --config config.py
> 15:45:38 INFO     errbot.err                Config check passed...
> 15:45:38 INFO     errbot.err                Selected backend 'IRC'.
> 15:45:38 INFO     errbot.err                Checking for './data'...
> 15:45:38 INFO     errbot.backend_manager    Backends search paths ['/usr/lib/python3.4/site-packages/errbot/backends']
> 15:45:38 INFO     errbot.backend_manager    Found those backends available:
> 15:45:38 INFO     errbot.backend_manager    	     Slack  (/usr/lib/python3.4/site-packages/errbot/backends/slack.py)
> 15:45:38 INFO     errbot.backend_manager    	  Telegram  (/usr/lib/python3.4/site-packages/errbot/backends/telegram.py)
> 15:45:38 INFO     errbot.backend_manager    	   Graphic  (/usr/lib/python3.4/site-packages/errbot/backends/graphic.py)
> 15:45:38 INFO     errbot.backend_manager    	      Text  (/usr/lib/python3.4/site-packages/errbot/backends/text.py)
> 15:45:38 INFO     errbot.backend_manager    	   Hipchat  (/usr/lib/python3.4/site-packages/errbot/backends/hipchat.py)
> 15:45:38 INFO     errbot.backend_manager    	      Null  (/usr/lib/python3.4/site-packages/errbot/backends/null.py)
> 15:45:38 INFO     errbot.backend_manager    	       IRC  (/usr/lib/python3.4/site-packages/errbot/backends/irc.py)
> 15:45:38 INFO     errbot.backend_manager    	      XMPP  (/usr/lib/python3.4/site-packages/errbot/backends/xmpp.py)
> 15:45:38 INFO     errbot.backend_manager    	      Test  (/usr/lib/python3.4/site-packages/errbot/backends/test.py)
> 15:45:38 INFO     errbot.main               Found Backend plugin: 'IRC'
> 						Description: This is the IRC backend for Err.
[...]

PR #612, opened 2016-02-23
Fixing #595 and #597 (join/joined mess on IRC)

qlixed:
Now Really fixes all those.
The code now use the join and endofnames IRC events to fire up the
event.
So now when we fire the callback_room_joined list of room
occupants is ready.
No more annoying "Not Joined yet" Messages!
The code works even if the room don't have occupants, because the IRC
sent ALWAYS a namerply/endofname, with the name of sole occupant.


qlixed:
Oh, I forgot to mention that add the __repr__, __str__, __unicode__ on IRCMUCOccuppant and IRCMUCRoom to simplify the readability during the debug.
Also add an __all_events method (commented out) to simplify future debuging on the the IRC Backend

errbot/backends/irc.py
```diff
@@ -299,12 +313,32 @@ def __init__(self,
...
         if username is None:
             username = nickname
         self.transfers = {}
         super().__init__([(server, port, password)], nickname, username, reconnection_interval=reconnect_on_disconnect)
+        # DEBUG HELPER
+        # Uncomment this to get an "All events" handler for any irc command/message sent
+        #  to the server, don't forget to uncomment __all_events too!.
+        #self.connection.add_global_handler("all_events", self.__all_events, 98)
+
+    # DEBUG HELPER
```
gbin: I am not fan of commented out code, can you make an if and see if the logger for this module is at DEBUG level and log it then ?

qlixed:
@gbin I Implement the changes mentioned, In fact I get rid of the debug helper, reviewing with more detail the current code, I see that the debug helper that I add is currently active on the backend, so now the code only add the fix for that event.

errbot/backends/irc.py
```diff
@@ -416,6 +430,21 @@ def on_dcc_connect(self, dcc, event):
     def on_dcc_disconnect(self, dcc, event):
         self.transfers.pop(dcc)
 
+    def on_endofnames(self, c, e):
+        # The e.arguments[0] contains the channel name.
+        # We filter that to avoid a misfire of the event.
+        room_name = e.arguments[0]
+        if room_name in self.__recently_joined_to:
```
gbin: This is a race condition on _recentry_joined_to

errbot/backends/irc.py
```diff
@@ -299,6 +313,7 @@ def __init__(self,
             self.send_public_message = rate_limited(channel_rate)(self.send_public_message)
         self._reconnect_on_kick = reconnect_on_kick
         self._pending_transfers = {}
+        self.__recently_joined_to = set()
```
gbin: rename to _recently_joined_to

errbot/backends/irc.py
```diff
@@ -254,7 +267,7 @@ def occupants(self):
         occupants = []
         try:
             for nick in self._bot.conn.channels[self.room].users():
-                occupants.append(IRCMUCOccupant(nick=nick, room=self.room))
+                occupants.append(IRCMUCOccupant(nick, self.room))
```
zoni: Keeping keyword arguments here would actually be safer (and more explicit, thus more clear)

errbot/backends/irc.py
```diff
@@ -566,7 +595,7 @@ def query_room(self, room):
         :returns:
             An instance of :class:`~IRCMUCRoom`.
         """
-        return IRCMUCRoom(room, bot=self)
+        return IRCMUCRoom(room, self)
```
zoni: Again, while not necessarily invalid, keeping keyword arguments here would actually be safer (and more explicit, thus more clear)

errbot/core_plugins/chatRoom.py
```diff
@@ -25,6 +25,8 @@ def callback_connect(self):
             self.connected = True
             for room in self.bot_config.CHATROOM_PRESENCE:
                 self.log.debug('Try to join room %s' % repr(room))
+                if 'password' not in self.bot_config.BOT_IDENTITY.keys():
+                    self.bot_config.BOT_IDENTITY['password']=None
```
zoni: This change most definitely doesn't belong here. The chatroom plugin is generic to all backends so it should contain as little (and preferably no) backend-specific code at all.

It looks like maybe you were trying to account for a user not specifying a `password` in their `BOT_IDENTITY` on IRC? If that is the case, it should be handled in the `__init__` of the IRC backend itself.

errbot/backends/irc.py
```diff
@@ -416,6 +430,21 @@ def on_dcc_connect(self, dcc, event):
     def on_dcc_disconnect(self, dcc, event):
         self.transfers.pop(dcc)
 
+    def on_endofnames(self, c, e):
```
zoni: You seem to have a pretty clear idea of what the `endofnames` represents and when it is fired. Could you add a docstring to the function briefly explaining it? That way it's clear to everyone (I wouldn't know without looking it up right now, for example).

errbot/backends/irc.py
zoni: Specifically, there's a chance that `on_endofnames` is called right around the same time multiple times from different threads (because of lot of these callbacks are handled in their own thread). What might happen is that they both attempt to update the set at the same time, so you have to guard against this.

There's `threading.Lock()` for this. I'd be happy to explain how to use it if you have never used it before.

errbot/backends/irc.py
zoni: Also, `c` and `e` are pretty bad names for parameters, although I'm guessing you copied them from the IRC library directly.

What would they stand for? If I had to guess, `callback` and `event`? Using such names might be more clear.

zoni:
@qlixed thanks a lot for working on this! There's a couple of issues me and @gbin spotted but I'm certain we can help you get those straightened out.

Do note that this branch has conflicts that must be resolved as well, it's going to be hard for us to merge it otherwise. Would you mind fixing the conflicts?

errbot/core_plugins/chatRoom.py
qlixed: My bad, this lines keeps for a previous local iteration of the code, I will remove it.

errbot/backends/irc.py
qlixed: Why? this not require a external usage, is for object internal use only this variable. 
Maybe it's some specific project requirement, There is some guidelines for this? I check here: http://errbot.io/contributing.html#contributing-to-errbot-itself. Maybe i skip a additional reference?

errbot/backends/irc.py
qlixed: I remove this because in some part (I can't recall now) the keyword argument was incorrect.
Besides this, I don't really think that is "safer" to use kwargs, but i will reset that as they would.

errbot/backends/irc.py
zoni: A single underscore denotes a "private" variable that should only be used internally by the class itself (note that in python, this is by convention only, it's not actually enforced).

Using a double underscore invokes something that is called name wrangling, which internally rewrites the variable name to have the classname prefixed. You use this if you want to be sure that if your class is subclassed and defines the same name, the two don't clash.

See https://docs.python.org/2/tutorial/classes.html#private-variables-and-class-local-references for the official explanation.

In this case, you do not want name wrangling to take place. It would in fact break subclasses from properly working. Hence, you should use a single underscore (the cases where you should use a double underscore are extremely rare in practice).

errbot/backends/irc.py
qlixed: Ah, indeed here is not nick the kwarg name, is mask, that's why i remove both, to avoid this kind of issues.

errbot/backends/irc.py
qlixed: @zoni To avoid an Unnecessary thread locking, as this _recently_joined_to variable is Thread local variable, what do you think if i use [threading.local()](https://docs.python.org/3/library/threading.html#thread-local-data) and put this set into a the thread local variable context?

zoni: That seems like a very bad thing to do here. There's no guarantee that `on_join` (which puts the room in `_recently_joined_to`) is going to be handled by the same thread as the `on_endofnames` here.

And even if that were the case now, there's no guarantee it would remain that way in the future. If it were thread local and different threads handled it, it would cause a very subtle bug that would likely be extremely hard to track down.

I wouldn't worry too much about thread locking. Especially not in this instance, as room joins are pretty rare.

qlixed:
@zoni @gbin I Think that make all he requested changes, please review it.

zoni:
I'll try to find the time to take a look at this tomorrow for you.

This branch still has conflicts against master which prevent merging though, could you try to fix those?

qlixed:
Conflicts free :)

gbin:
LGTM, thx

PR #613, opened 2016-02-23
Repos configuration, cache and search

gbin:
Search just work, gonna work from there for the other regressions.


sijis:
I gave this a quick run through. Searching is great!

I was getting a `TypeError: the JSON object must be str, not 'bytes'` for reading the json file and the fix is the following:
```
diff --git a/errbot/repo_manager.py b/errbot/repo_manager.py
index 0e3d33c..21e27ed 100644
--- a/errbot/repo_manager.py
+++ b/errbot/repo_manager.py
@@ -110,7 +110,7 @@ class BotRepoManager(StoreMixin):
                     log.debug('Update from local source %s...', source)
                     src_file = open(source, 'r')
 
-                index.update(json.loads(src_file.read()))
+                index.update(json.loads(src_file.read().decode('utf-8')))
             except (HTTPError, URLError, IOError):
                 log.exception('Could not update from source %s, keep the index as it is.', source)
                 break
```

I also got an `HTTPError` for being unable to get to the `repos.json`. I tried a few things (like `urllib.errors.HTTPError` in the except clause but that did not appear to work.
```
22:10:49 ERROR    errbot.repo_manager       Could not update from source http://errbot.io/repos.json, keep the index as it is.
  File "/home/saviles/data/git/github/errbot/errbot/repo_manager.py", line 108, in index_update
    src_file = urlopen(url=source, timeout=10)
  File "/usr/lib64/python3.4/urllib/request.py", line 153, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib64/python3.4/urllib/request.py", line 461, in open
    response = meth(req, response)
  File "/usr/lib64/python3.4/urllib/request.py", line 571, in http_response
    'http', request, response, code, msg, hdrs)
  File "/usr/lib64/python3.4/urllib/request.py", line 499, in error
    return self._call_chain(*args)
  File "/usr/lib64/python3.4/urllib/request.py", line 433, in _call_chain
    result = func(*args)
  File "/usr/lib64/python3.4/urllib/request.py", line 579, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 404: Not Found
```

sijis:
A quick fix for passing no arguments to `!repos search`:
```
diff --git a/errbot/core_plugins/plugins.py b/errbot/core_plugins/plugins.py
index bafd626..446b822 100644
--- a/errbot/core_plugins/plugins.py
+++ b/errbot/core_plugins/plugins.py
@@ -98,7 +98,7 @@ class Plugins(BotPlugin):
         """
         if not args:
             # TODO(gbin): return all the repos.
-            return "Please specify a keyword."
+            return {'error': "Please specify a keyword."}
         return {'repos': self._bot.repo_manager.search_repos(args)}
 
     @botcmd(split_args_with=' ', admin_only=True)
diff --git a/errbot/core_plugins/templates/repos2.md b/errbot/core_plugins/templates/repos2.md
index d46b2bf..4d7848e 100644
--- a/errbot/core_plugins/templates/repos2.md
+++ b/errbot/core_plugins/templates/repos2.md
@@ -1,4 +1,8 @@
+{% if error %}
+{{error}}
+{% else %}
 Status  | Name                    | Description
 ------- | ----------------------- | ------------
 {% for repo in repos %}         | {{('**'+repo.entry_name+'**').ljust(22)}} | {{repo.documentation}}
 {% endfor %}
+{% endif %}
```

zoni:
`plugin-gen.py` contains a bug as well:

```
DEBUG:__main__:Checking Djiit/err-meetup...
  File "plugin-gen.py", line 172, in <module>
    main()
  File "plugin-gen.py", line 164, in main
    find_plugins()
  File "plugin-gen.py", line 155, in find_plugins
    check_repo(repo)
  File "plugin-gen.py", line 96, in check_repo
    avatar_url = get_avatar_url(repo)
  File "plugin-gen.py", line 48, in get_avatar_url
    time.sleep(PAUSE)
NameError: name 'PAUSE' is not defined
```

zoni:
Also, the `user_cache` file should be removed from the repository and added to `.gitignore`.

gbin:
I checked it in because it is a long lived cache, no need to refresh it for at least a couple months.

zoni:
It looks like plugins may fail to actually install right...

```
>>> !repos install zoni/err-converter
<<< A new plugin repository has been installed correctly from zoni/err-converter. Refreshing the plugins commands...
<<< Plugins reloaded without any error.
```

..but at this point the plugin doesn't show up in `!status`. Even after a bot restart, there's no trace of the plugin. 

There isn't any relevant debug logging either.

It looks like err didn't clone the git repo at all, as it isn't present in the `BOT_DATA_DIR`.

errbot/main.py
```diff
@@ -60,8 +60,13 @@ def setup_bot(backend_name, logger, config, restore=None):
     if not path.exists(botplugins_dir):
         makedirs(botplugins_dir, mode=0o755)
 
+    plugin_indexes = getattr(config, 'BOT_PLUGIN_INDEXES', ('http://errbot.io/repos.json',))
```
zoni: I think it might be better to host this under a subdomain, something like `repos.errbot.io/repos.json`.

This will ensure it doesn't conflict with the sphinx-built generation so we don't have to interweave the two together, and would allow us to do HTTPS (if we don't put it on github pages) which I think we should absolutely do here.

errbot/repo_manager.py
```diff
@@ -30,28 +44,112 @@ def get_known_repos():
...
         self.plugin_dir = plugin_dir
         self.open_storage(storage_plugin, 'repomgr')
 
+    def check_for_index_update(self):
+        if REPO_INDEX not in self:
+            log.info('No repo index, creating it.')
+            self[REPO_INDEX] = {LAST_UPDATE: 0}
+
+        if datetime.fromtimestamp(self[REPO_INDEX][LAST_UPDATE]) < datetime.now() - REPO_INDEXES_CHECK_INTERVAL:
+            log.debug('Index is too old, update it.')
```
zoni: I'd log this at INFO level rather than DEBUG.

zoni:
When doing a `!repos search bla` and the index is outdated *and* fails to update, I think we should send a message to the user informing them about this.

gbin:
@sijis applied your patch for the error, thx

gbin:
@zoni fixed the bug on the avatars, thx

errbot/repo_manager.py
gbin: done

errbot/main.py
gbin: done

gbin:
So install/uninstall was not working... it should work now but still rough on the edges.

We lose the description we had of the repos from the known repo list, maybe I'll make a kind of heuristic that if a repo only contains 1 plugin, the repo description will take the description of the plugin.

PR #616, opened 2016-02-28
documentation update: sphinx should correctly split parameters

pendragon-:
See for example http://errbot.io/errbot.html?highlight=send#errbot.BotPlugin.send_stream_request
The parameter description is mixed in with the description of the method itself.


gbin:
thx

PR #624, opened 2016-03-01
Plugin and repo error handling

gbin:


errbot/core_plugins/plugins.py
```diff
@@ -17,24 +19,27 @@ def repos_install(self, mess, args):
...
-                mess.frm,
-                ("A new plugin repository has been installed correctly from "
-                 "%s. Refreshing the plugins commands..." % args),
-                message_type=mess.type
-            )
-        loading_errors = self._bot.plugin_manager.activate_non_started_plugins()
-        if loading_errors:
-            return loading_errors
-        return "Plugins reloaded without any error."
+            yield "You should have an urls/git repo argument"
```
zoni: Since we're changing all this code anyway, lets clarify this error message to *"Please specify a repository listed in '!repos' or give me the URL to a git repository that I should clone for you."*

errbot/core_plugins/plugins.py
```diff
@@ -17,24 +19,27 @@ def repos_install(self, mess, args):
...
+            return
+        try:
+            local_path = self._bot.repo_manager.install_repo(args)
```
zoni: Maybe it's necessary to do `args.strip()` here too?

errbot/core_plugins/plugins.py
gbin: done

errbot/core_plugins/plugins.py
gbin: Done

errbot/core_plugins/plugins.py
```diff
@@ -17,24 +19,27 @@ def repos_install(self, mess, args):
...
+            errors = self._bot.plugin_manager.update_dynamic_plugins()
+            if errors:
+                yield 'Some plugins are generating errors:\n' + '\n'.join(errors.values())
+                # if the load of the plugin failed, uninstall cleanly teh repo
+                for path in errors.keys():
+                    if path.startswith(local_path):
+                        yield 'Removing %s as it did not load correctly.' % local_path
+                        shutil.rmtree(local_path)
```
zoni: What happens if I have a plugin in `BOT_EXTRA_PLUGIN_DIR` that fails to load? Wouldn't it then remove that too (possibly destroying my hard work).

Maybe we can check and only remove the plugin we just installed, if it has errors?

gbin: So this is what I am doing if there is no bug: 
local_path = self._bot.repo_manager.install_repo(args)   # gets the path to the repo this command is installing

if path.startswith(local_path)  # only remove the one we are talking about.

errbot/repo_manager.py
```diff
@@ -164,6 +168,14 @@ def get_all_repos_paths(self):
         return [self.plugin_dir + os.sep + d for d in self.get(INSTALLED_REPOS, {}).keys()]
 
     def install_repo(self, repo):
+        """
+        Install the repository from repo
+
+        :repo the url, git url or path on disk of a repository. It can point to either a git repo or
+        a .tar.gz of a plugin
+        :returns the path on disk where the repo has been installed on.
+        :raises RepoException if an error occured.
```
zoni: Sphinx won't render this correctly, use:

```
:param repo: 
    The url, git url or path on disk of a repository. It can point to either a git repo or
     a .tar.gz of a plugin
:returns:
    The path on disk where the repo has been installed on.
:raises: :class:`~RepoException` if an error occured.
```

gbin: done

errbot/core_plugins/plugins.py
zoni: Ah, gotcha! :+1:

PR #625, opened 2016-03-01
Adding exception catches for files not existing.

jasedit:
This safely bypasses if the user_cache or blacklist doesn't exist, but the GitHub API token must exist.

### DIFF
--- a/tools/plugin-gen.py
+++ b/tools/plugin-gen.py
@@ -16,7 +16,7 @@ log.setLevel(logging.DEBUG)
 DEFAULT_AVATAR = 'https://upload.wikimedia.org/wikipedia/commons/5/5f/Err-logo.png'
 
 try:
-    user,token = open('token', 'r').read().strip().split(':')
+    user, token = open('token', 'r').read().strip().split(':')
     # token is generated from the personal tokens in github.
     AUTH = HTTPBasicAuth(user, token)
 except FileNotFoundError:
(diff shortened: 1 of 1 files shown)


zoni:
We should remove the hardcoded gbin too while we're at it. :smile:. Maybe put it in the same token file, split with a `:`? As in `gbin:afef243uhhfef3ur38rhl`.

jasedit:
Should it default to gbin is the user is not described? Or is it acceptable to have it simply fail if tokens doesn't contain a user?

gbin:
@zoni ha year agreed, sorry about that :)

zoni:
> Should it default to gbin is the user is not described? Or is it acceptable to have it simply fail if tokens doesn't contain a user?

It's fine to fail if the user is missing. This script is mainly for us maintainers (@gbin and myself) and will soon be loaded automatically so normal users won't generally have to run it.

jasedit:
Updating token parsing as requested, and it behaves as requested.

zoni:
Looks good. If you could just fix the pep8 errors though (you can find them on Travis but I'll put them in here too), then we can have this merged.

Thanks!

------------------------------ 

```
 ./run_tests.py

./tools/plugin-gen.py:19:9: E231 missing whitespace after ','

    user,token = open('token', 'r').read().strip().split(':')

        ^

./tools/plugin-gen.py:35:5: E265 block comment should start with '# '

    #File doesn't exist, so we continue on

    ^

1       E231 missing whitespace after ','

1       E265 block comment should start with '# '

running check

Pep8 failed.
```

gbin:
thx

PR #634, opened 2016-03-05
err.py: Fix BOT_EXTRA_BACKEND_DIR in `-l`

mrshu:
* When `errbot` or `err.py` run with the `-l` parameter which should
  list the available backends the BOT_EXTRA_BACKEND_DIR configuration
  option is ignored. This commit fixes that problem.

-----------------------

to @gbin


gbin:
thx

PR #649, opened 2016-03-08
Rework shutdown and restart logic

### LINKED ISSUE #647: bot not restarting with !restart
Logs stops with `Bye.`.

The instance is started with
``` bash
#!/bin/sh
source ./bin/activate # virtualenv
./bin/errbot -d
```

v3.2.3

zoni:
* This fixes #647, because with `--daemonize` the bot couldn't find its
  `config.py` on restart.
* This fixes shutdown, which was keeping the bot hanging but with storage
  and plugins shut down. It also greatly simplifies the shutdown logic.
* It provides a `--kill` flag to instantly kill the bot with a SIGKILL.
* It renames `killbot` to `shutdown` to more accurately reflect what it
  does and be consistent with the `!restart` naming.

### DIFF
--- a/errbot/core_plugins/health.py
+++ b/errbot/core_plugins/health.py
@@ -1,9 +1,9 @@
 import gc
 import os
 import signal
-from datetime import datetime
 
-from errbot import BotPlugin, botcmd
+from datetime import datetime
+from errbot import BotPlugin, botcmd, arg_botcmd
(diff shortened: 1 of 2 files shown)


gbin:
LGTM even if you killed my 2001 reference :P

zoni:
I know, that's why I put it up for review especially! :grin:

PR #655, opened 2016-03-11
Allow bot configuration from groupchat

zoni:
I've always been annoyed how the bot forces me to send `admin_only` commands via a direct message, because I'm often in restricted, private channels with just the bot alone.

I think it should be up to end-users to decide whether they want to be able to issue admin_only commands in rooms as well, or whether they should be forced through a private message. Our ACLs make this really easy to configure now if this is desired so there's no reason for us to force this on users with no way to turn it off.

### DIFF
--- a/errbot/backends/test.py
+++ b/errbot/backends/test.py
@@ -96,7 +96,7 @@ class TestOccupant(TestPerson, RoomOccupant):
         return self._room
 
     def __unicode__(self):
-        return self._person + '@' + self._room
+        return self._person + '@' + str(self._room)
 
     __str__ = __unicode__
 
--- a/errbot/core_plugins/acls.py
+++ b/errbot/core_plugins/acls.py
@@ -93,23 +93,13 @@ class ACLS(BotPlugin):
                     dry_run
             )
 
-        return msg, cmd, args
-
-    @cmdfilter
-    def admin(self, msg, cmd, args, dry_run):
-        """
-        Check command against the is_admin criteria.
-
-        :param msg: The original message the commands is coming from.
-        :param cmd: The command name
-        :param args: Its arguments.
-        :param dry_run: pass True to not act on the check (messages / deferred auth etc.)
-
-        """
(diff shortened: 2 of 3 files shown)


gbin:
This is a pretty scary proposition TBH.
1. in general identities in one to ones are more defined that groupchat (see nicknames on IRC, full names on Hipchat etc ...)
2. when configuring plugins and putting credentials, it might be recorded in the chatroom so any of your user get hacked (even their email with a "what you missed from the chatroom", your infrastructure is gone

On the usability side, this is a one shot operation most of the time.

zoni:
I originally wanted to avoid making yet another config option for this, but fair enough, we should probably protect users by default who may not think the consequences through.

Would you be opposed to an optional `ALLOW_ADMIN_COMMANDS_INSIDE_ROOMS` which defaults to `False` (the old behaviour) but can be set to `True` to allow this?

gbin:
Can it be done through default ACLs ?

zoni:
Not really, because we have no way to say *"Apply these ACLs to `admin_only` commands"*.

zoni:
One way in which it could work that way is if this code is merged with the ACL checks.

Then the default could be not to allow it via groupchat, but if the command specifically has `allowmuc: True` in the `ACCESS_CONTROLS`, it could override this.

gbin:
Actually if we can have is_admin as a conventional group, could we suppress the BOT_ADMINS setting and have a reasonable default behavior ?

gbin:
We are kind of doing that...  'uptime': {'allowusers': BOT_ADMINS},

zoni:
That's just restricting commands which aren't already marked as `admin_only` commands to only be executed by people in `BOT_ADMINS` too. The `admin_only` flag is set in the command definition itself so we can't *remove* that flag from commands.

I've merged the ACL and admin_only *cmdfilters* together though, how do you feel about this approach? It's still not possible by default, but now you can specifically override this behaviour by defining an ACL for the command with `'allowmuc': True` in it.

PR #658, opened 2016-03-12
Add a "new plugin wizard" to errbot

zoni:
This adds a new flag, `--new-plugin` to errbot which asks a series of
questions to create a skeleton for a new plugin.

Example:

```
$ errbot --new-plugin /tmp
This wizard will create a new plugin for you in '/tmp'.
What should the name of your new plugin be?
> Example Bot
What may I use as a short (one-line) description of your plugin?
> An example bot
Which python version will your plugin work with? 2, 2+ or 3?
> 2+
Which minimum version of errbot will your plugin work with? Leave blank to support any version
> 0.0.0
Which maximum version of errbot will your plugin work with? Leave blank to support any version
> 4.0.0
Success! You'll find your new plugin at '/tmp/example_bot.py'

$ cat /tmp/example_bot.plug
[Core]
name = Example Bot
module = example_bot

[Documentation]
description = An example bot

[Python]
version = 2+

[Errbot]
min = 0.0.0
max = 4.0.0

$ head /tmp/example_bot.py
from errbot import BotPlugin, botcmd, webhook

class ExampleBot(BotPlugin):
    """
    An example bot
    """

    def activate(self):
        """
```

### DIFF
--- a/errbot/err.py
+++ b/errbot/err.py
@@ -21,6 +21,7 @@ import logging
 import argparse
 from os import path, sep, getcwd, access, W_OK
 from platform import system
+from .plugin_wizard import new_plugin_wizard
 from .version import VERSION
 
 PY3 = sys.version_info[0] == 3
@@ -161,6 +162,8 @@ def main():
     mode_selection.add_argument('-r', '--restore', nargs='?', default=None, const='default',
                                 help='restore a bot from backup.py (default: backup.py from the bot data directory)')
     mode_selection.add_argument('-l', '--list', action='store_true', help='list all available backends')
+    mode_selection.add_argument('--new-plugin', nargs='?', default=None, const='current_dir',
+                                help='create a new plugin in the specified directory')
     mode_selection.add_argument('-T', '--text', dest="backend", action='store_const', const="Text",
                                 help='force local text backend')
     mode_selection.add_argument('-G', '--graphic', dest="backend", action='store_const', const="Graphic",
@@ -173,6 +176,19 @@ def main():
                                   help='Specify the pid file for the daemon (default: current bot data directory)')
 
     args = vars(parser.parse_args())  # create a dictionary of args
+
+    # This must come BEFORE the config is loaded below, to avoid printing
+    # logs as a side effect of config loading.
+    if args['new_plugin']:
+        directory = os.getcwd() if args['new_plugin'] == "current_dir" else args['new_plugin']
+        logging.getLogger('').setLevel(level=1000)  # Avoid logging stuff
+        try:
+            new_plugin_wizard(directory)
+        except KeyboardInterrupt:
+            sys.exit(1)
+        finally:
+            sys.exit(0)
+
     config_path = args['config']
     # setup the environment to be able to import the config.py
     if config_path:
--- a/errbot/err.py
+++ b/errbot/err.py
@@ -181,7 +181,8 @@ def main():
     # logs as a side effect of config loading.
     if args['new_plugin']:
         directory = os.getcwd() if args['new_plugin'] == "current_dir" else args['new_plugin']
-        logging.getLogger('').setLevel(level=1000)  # Avoid logging stuff
+        for handler in logging.getLogger().handlers:
+            logger.removeHandler(handler)
         try:
             new_plugin_wizard(directory)
         except KeyboardInterrupt:
--- a/errbot/err.py
+++ b/errbot/err.py
@@ -187,6 +187,9 @@ def main():
             new_plugin_wizard(directory)
         except KeyboardInterrupt:
             sys.exit(1)
+        except Exception as e:
+            sys.stderr.write(str(e) + "\n")
+            sys.exit(1)
         finally:
             sys.exit(0)
 
--- a/errbot/plugin_wizard.py
+++ b/errbot/plugin_wizard.py
@@ -39,10 +39,17 @@ def new_plugin_wizard(directory=None):
     description = ask(
         "What may I use as a short (one-line) description of your plugin?"
     )
+    if PY2:
+        default_python_version = "2+"
+    else:
+        default_python_version = "3"
     python_version = ask(
-        "Which python version will your plugin work with? 2, 2+ or 3?",
-        valid_responses=['2', '2+', '3']
+        "Which python version will your plugin work with? 2, 2+ or 3? I will default to "
+        "{version} if you leave this blank.".format(version=default_python_version),
+        valid_responses=['2', '2+', '3', '']
     )
+    if python_version.strip() == "":
+        python_version = default_python_version
     errbot_min_version = ask(
         "Which minimum version of errbot will your plugin work with? "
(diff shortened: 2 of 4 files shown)


errbot/plugin_wizard.py
```diff
@@ -0,0 +1,144 @@
...
+        )
+        ask(
+            "Press Ctrl+C to abort now or type in 'overwrite' to confirm overwriting of these files.",
+            valid_responses=["overwrite"],
+        )
+
+    with open(plug_path, 'w') as f:
+        plug.write(f)
+
+    plugin_code = plugin_code.replace("class Skeleton(", "class %s(" % class_name)
```
gbin: Maybe we could base it off a jinja template ?

errbot/err.py
```diff
@@ -154,29 +156,39 @@ def main():
...
         option_group.add_argument('-p', '--pidfile', default=None,
                                   help='Specify the pid file for the daemon (default: current bot data directory)')
 
     args = vars(parser.parse_args())  # create a dictionary of args
+
+    # This must come BEFORE the config is loaded below, to avoid printing
+    # logs as a side effect of config loading.
+    if args['new_plugin']:
+        directory = os.getcwd() if args['new_plugin'] == "current_dir" else args['new_plugin']
+        logging.getLogger('').setLevel(level=1000)  # Avoid logging stuff
```
gbin: Maybe something like that ?
logging.removehandler(handler) for handler in logging.handlers

errbot/plugin_wizard.py
zoni: That's not a bad idea. However my initial idea was to specifically reuse the [plugin skeleton](https://github.com/errbotio/plugin-skeleton) so that any improvements made there are automatically pulled in.

Having Jinja markers in that file might look a little weird to people browsing the repository directly to use as reference.

errbot/plugin_wizard.py
```diff
@@ -0,0 +1,144 @@
...
+    ).strip()
+    module_name = name.lower().replace(' ', '_')
+    class_name = "".join([s.capitalize() for s in name.lower().split(' ')])
+
+    description = ask(
+        "What may I use as a short (one-line) description of your plugin?"
+    )
+    python_version = ask(
+        "Which python version will your plugin work with? 2, 2+ or 3?",
+        valid_responses=['2', '2+', '3']
```
gbin: Maybe it should default to 2+ if the interpreter is python 2 and 3 if it is python 3

errbot/plugin_wizard.py
```diff
@@ -0,0 +1,144 @@
...
+    input = raw_input
+
+
+def new_plugin_wizard(directory=None):
+    """
+    Start the wizard to create a new plugin in the current working directory.
+    """
+    if directory is None:
+        print("This wizard will create a new plugin for you in the current directory.")
+        directory = os.getcwd()
```
gbin: This should be BOT_EXTRA_PLUGINS_DIR if this is specified

errbot/err.py
zoni: I was kinda lazy. :smiling_imp: But yeah should probably do it that way.

errbot/plugin_wizard.py
zoni: I considered this, but I'd prefer to let the user really think about which Python versions they want to support, instead of automatically defaulting to what they're currently using to develop on.

errbot/plugin_wizard.py
gbin: I think I might prefer maybe a simpler version of the skeleton plugin bundled with errbot so we can maintain it consistently with errbot. This looks very fragile if in another repo IMHO.
Then the Jinja markers are not really a problem as they are internal details.

errbot/err.py
gbin: Man, if only I were as productive as you are while being "lazy" :)

errbot/plugin_wizard.py
zoni: This would complicate things a little and require loading the config before starting the wizard. It would also become more unpredictable, picking the current dir if `BOT_EXTRA_PLUGIN_DIR` is unspecified or that directory if it is specified.

I'd prefer to keep it as-is and require the user to pass in the directory to where they want it created if they don't want it in the current directory, which is more in line with how many other unix tools behave as well.

errbot/plugin_wizard.py
```diff
@@ -0,0 +1,144 @@
...
+        print(
+            "Error: I couldn't download the plugin skeleton from GitHub "
+            "which is needed to write out a new plugin file for you."
+        )
+        print("Details: {!s}".format(e))
+        sys.exit(1)
+
+    name = ask(
+        "What should the name of your new plugin be?",
+        validation_regex=r'^[a-zA-Z][a-zA-Z0-9 _-]*$'
```
gbin: \w might help

zoni: I avoid those unless it becomes quite unreadable otherwise. I find I tend to forget which character ranges are included exactly in which `\x` set and I imagine other people suffer from the same thing.

Considering the validation is also printed on incorrect input, I think this is easier for people to understand.

errbot/plugin_wizard.py
gbin: fair enough, maybe a little nudge to push python 3 ?
"In doubt, choose python 3 :) "

errbot/plugin_wizard.py
gbin: ok

errbot/plugin_wizard.py
gbin: ok

gbin:
Another thing, I think it should create a subdirectory as err-nameofplugin (name.tolower()) and put the plugin in there.

errbot/plugin_wizard.py
```diff
@@ -0,0 +1,144 @@
...
+    )
+    python_version = ask(
+        "Which python version will your plugin work with? 2, 2+ or 3?",
+        valid_responses=['2', '2+', '3']
+    )
+    errbot_min_version = ask(
+        "Which minimum version of errbot will your plugin work with? Leave blank to support any version"
+    ).strip()
+    errbot_max_version = ask(
+        "Which maximum version of errbot will your plugin work with? Leave blank to support any version"
```
mrshu: @zoni Maybe the string "current" could be also accepted here and replaced with the version of errbot that is currently running?

It migh also be desirable in `errbot_min_version`'s case.

zoni: Done. :+1:

zoni:
> Another thing, I think it should create a subdirectory as err-nameofplugin (name.tolower()) and put the plugin in there. 

Done.

errbot/plugin_wizard.py
zoni: I changed it slightly. The question is still asked, but if you just press enter without inputting anything it will pick a default, which on Python 3 is just *3*, and on Python 2 is *2+*.

PR #675, opened 2016-03-15
Automatically create BOT_DATA_DIR if it doesn't exist

zoni:
I'm not sure whether it would be desirable to automatically create this directory, but this will do it.

### DIFF
--- a/errbot/err.py
+++ b/errbot/err.py
@@ -224,9 +224,11 @@ def main():
     # Check if at least we can start to log something before trying to start
     # the bot (esp. daemonize it).
 
-    log.info("Checking for '%s'..." % config.BOT_DATA_DIR)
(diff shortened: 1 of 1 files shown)


gbin:
Still torn about this one, any other opinion ? Especially if there is a default in config template it could do some weird thing depending on the platform (mac/win/linux).

zoni:
Random thought. How about prompting the user whether it should try and create this directory when it doesn't exist yet? A simple yes or no question before it actually does the `os.makedirs()` call?

sijis:
If we do the config wizard idea, then that should handle this creation at that point.

I think since folks have to manually update the config file, then the directory should be manually created too.

zoni:
I'm going to close this for now, when we do implement a config wizard it can absolutely take care of it then.

Also, while hacky, you could check whether the directly exists at import time of `config.py` (by just putting the relevant Python code in there) and create it then and there if it doesn't. :smile:

PR #680, opened 2016-03-18
Format URLs on Slack

zoni:
This converts URLs on Slack so that instead of printing the URL after
the link text it actually hyperlinks just the link text itself.

As a side effect, this changes how messages are sent from using the
RealTime Messaging API to the generic web API because the RTM API
does not support markup like this, only the web API does.

This is described at https://api.slack.com/rtm which states:

> The RTM API only supports posting simple messages formatted using
> our default message formatting mode. It does not support attachments
> or other message formatting modes. To post a more complex message as
> a user clients can call the chat.postMessage Web API method


zoni:
#### Before:

screenshot from 2016-03-18 09-21-39

#### After:

screenshot from 2016-03-18 09-20-58

gbin:
Excellent

PR #690, opened 2016-03-24
Update botcommands.rst

kevinlondon:
All `type` arguments must be a callable, so if it's used as a string it will cause the plugin to error.


zoni:
Good catch, thanks!

PR #691, opened 2016-03-27
Fix #682 callback_room_topic is not fired on IRC Backends 

### LINKED ISSUE #682: callback_room_topic is not fired on IRC Backends
The change of a topic in a IRC connection is not firing the callback_room_topic.
There are errors in the current implementation.
I think that I can work them out.

qlixed:
Here is the fix of the topic callback and a rework of how IRCMUCRoom is handled that make the fix possible, additionally to that, the rework of IRCMUCRoom handling now is more efficient.
Now there is a _room dict on the bot that contains the IRCMUCRoom objects, so now the object is created once and accessed using the room name in the _room dict.
There is a TODO marked there to review if we have some memory leak (as a persistent object in memory that is not more used). I tested and apparently there is no problem, but I prefer to let there and make further revision on different configs.


errbot/backends/irc.py
```diff
@@ -146,10 +148,14 @@ def __repr__(self):
...
 
 class IRCRoom(Room):
+
     def __init__(self, room, bot):
         self._bot = bot
         self.room = room
         self.connection = self._bot.conn.connection
+        self._topic = None
+        # TODO: Check if this leaks a room on reconnect
+        self._bot._rooms[self.room] = self
```
zoni: You have a race condition here. `self._bot._rooms` may be modified concurrently by different threads.

errbot/backends/irc.py
```diff
@@ -186,6 +192,7 @@ def leave(self, reason=None):
             reason = ""
 
         self.connection.part(self.room, reason)
+        del self._bot._rooms[self.room]  # Destroy the room that we're leaving
```
zoni: We're deleting an entry from the rooms dict so it's obvious what this code is doing. Comments should generally not describe *what* code is doing (as that should be obvious), only *why* it is doing it if this isn't immediately clear.

Also, race condition on `self._bot._rooms` again. :smile:

errbot/backends/irc.py
```diff
@@ -408,7 +420,9 @@ def send_public_message(self, to, line):
         except ServerNotConnectedError:
             pass  # the message will be lost
 
-    def on_disconnect(self, _, e):
+    def on_disconnect(self, connection, event):
+        del self.bot._rooms  # Destroy all rooms on Disconnect
+        self.bot._rooms = {}
```
zoni: The `del self.bot_rooms` line (424) may be removed entirely, as you're setting it to an empty dict on the next line (425).

errbot/backends/irc.py
```diff
@@ -481,8 +495,56 @@ def on_join(self, connection, event):
         # We need to wait to endofnames message.
         room_name = event.target
         with self._recently_joined_lock:
+            t = IRCRoom(room_name, self.bot)
```
zoni: `t` appears to be unused here.

errbot/backends/irc.py
```diff
@@ -481,8 +495,56 @@ def on_join(self, connection, event):
...
+            :param:
+                connection: Is an 'irc.client.ServerConnection' object
+
+            :param:
+                event: Is an 'irc.client.Event' object
+                the event.arguments[1] contains the topic of the room.
+                the event.arguments[0] contains the room name
+        """
+        room_name = event.arguments[0]
+        self.bot._rooms[room_name]._topic = event.arguments[1]
```
zoni: There's a race condition on `self.bot._rooms` here.

zoni: I also don't like how the code is directly setting a private attribute (`_topic`) of `IRCRoom` here. Perhaps create a `callback_topic` on the `IRCRoom` class with a docstring stating it is called by the errbot backend code and not meant to be used by users directly.

errbot/backends/irc.py
```diff
@@ -481,8 +495,56 @@ def on_join(self, connection, event):
...
+            :param:
+                connection: Is an 'irc.client.ServerConnection' object
+
+            :param:
+                event: Is an 'irc.client.Event' object
+                the event.target contains the room name.
+                the event.arguments[0] contains the topic name
+        """
+        room_name = event.target
+        self.bot._rooms[room_name]._topic = event.arguments[0]
```
zoni: Race condition on `self.bot._rooms` here and same comment as above about directly modifying `_topic`.

errbot/backends/irc.py
```diff
@@ -481,8 +495,56 @@ def on_join(self, connection, event):
...
+
+            :param:
+                connection: Is an 'irc.client.ServerConnection' object
+
+            :param:
+                event: Is an 'irc.client.Event' object
+                the event.arguments[0] contains the room name
+        """
+        room_name = event.arguments[0]
+        self.bot._rooms[room_name]._topic = None
```
zoni: Race condition on `self.bot._rooms` here and same comment as above about directly modifying `_topic`.

errbot/backends/irc.py
```diff
@@ -638,7 +701,9 @@ def query_room(self, room):
         :returns:
             An instance of :class:`~IRCMUCRoom`.
         """
-        return IRCRoom(room, bot=self)
+        if room not in self._rooms.keys():
+            self._rooms[room] = IRCRoom(room, self)
+        return self._rooms[room]
```
zoni: Possible race condition on `self._rooms` here as well.

errbot/backends/irc.py
```diff
@@ -651,9 +716,7 @@ def rooms(self):
         :returns:
             A list of :class:`~IRCMUCRoom` instances.
         """
-
-        channels = self.conn.channels.keys()
-        return [IRCRoom(channel) for channel in channels]
+        return self._rooms
```
zoni: `self._rooms` is a dictionary, but the API of `rooms()` says this should return a list of instances so this should be `return self._rooms.values()` instead.

zoni:
Great work so far on improving how the IRC backend works @qlixed. I haven't actually tested this yet, but I've left some comments on things I could see need improving further still.

errbot/backends/irc.py
qlixed: Maybe we can make callback_topic a private function? so we can use _cb_topic instead to set the value?
I don't like the idea of put as public method that requires a special treatment on the user side. I prefer that the public methods can be fully used by the user.

qlixed: Furthermore the private attributes indeed are in part to be used in this way, internally or by another "friend" class that. In this case the IRCBackend fully manage the IRCRoom, so, I see pretty logically the usage in this way here, but if you prefer I can make the private callback method as mentioned in the previous comment.

errbot/backends/irc.py
```diff
@@ -481,8 +495,56 @@ def on_join(self, connection, event):
...
+            If the room that you join don't have a topic set, nothing happens.
+            :param:
+                connection: Is an 'irc.client.ServerConnection' object
+
+            :param:
+                event: Is an 'irc.client.Event' object
+                the event.arguments[1] contains the topic of the room.
+                the event.arguments[0] contains the room name
+        """
+        room_name = event.arguments[0]
```
gbin: unpacking works well here too:
```
name, topic = event.argument
with self.room_lock: # or whatever you choose to avoid the race condition
     self.bot._rooms[name]  = topic
```

errbot/backends/irc.py
```diff
@@ -638,7 +701,9 @@ def query_room(self, room):
         :returns:
             An instance of :class:`~IRCMUCRoom`.
         """
-        return IRCRoom(room, bot=self)
+        if room not in self._rooms.keys():
```
gbin:
``` python
# can be simplified
if room not in self._rooms:
 [...]
 ```

File: errbot/backends/irc.py (hunk 5 above)
Comment (zoni) [inline errbot/backends/irc.py]: I'd still prefer to see it as a public method as in my previous comment. Private methods should remain internal to the class itself and should not be called by code outside of that class.

**Comment by qlixed** [comment]
Implemented some of the commented changes and modify also the /PART (room.leave()) code. Now the code use the on_part event of the python-irc lib. Let me know if I Forget something.

File: errbot/backends/irc.py [hunk 12]
```diff
@@ -160,6 +173,20 @@ def __str__(self):
...
+        """
+            Internal use only.
+            To set the topic use the topic property.
+            Not use under any circunstance. I REALLY mean it.
+            It will broke things badly.
+            Probably leave you in banckrupcy.
+            :params:
+                current_topic: Well, really? I told you that not
+                use this, so go away and use the topic property.
+        """
```
zoni: Make this docstring a little more professional please. Suggested:

```
"""
Store the current topic for this room.

This method is called by the IRC backend when a `currenttopic`, `topic` or `notopic` IRC event is received to store the topic set for this channel.

This function is not meant to be executed by regular plugins.
To get or set the topic of a channel, use the `topic` attribute instead.
"""
```

errbot/backends/irc.py
```diff
@@ -248,6 +277,9 @@ def topic(self, topic):
         :param topic:
             The topic to set.
         """
+        if not self.joined:
+            raise RoomNotJoinedError("Must join the room to set the topic")
+        self.cb_set_topic(topic)
```
zoni: It feels weird to call the callback function here, when changing the topic. Doesn't the IRC library trigger a `on_topic` event once the `self.connection.topic(self.room, topic)` is received and processed by the server? This line can be removed if that's the case.

errbot/backends/irc.py
```diff
@@ -408,7 +440,8 @@ def send_public_message(self, to, line):
         except ServerNotConnectedError:
             pass  # the message will be lost
 
-    def on_disconnect(self, _, e):
+    def on_disconnect(self, connection, event):
+        self.bot._rooms = {}
```
zoni: `self.bot._rooms` needs to be protected with a thead lock.

errbot/backends/irc.py
```diff
@@ -481,8 +539,61 @@ def on_join(self, connection, event):
         # We need to wait to endofnames message.
         room_name = event.target
         with self._recently_joined_lock:
+            if room_name not in self.bot._rooms:
+                self.bot._rooms[room_name] = IRCRoom(room_name, self.bot)
```
zoni: `self.bot._rooms` needs to be protected with a thead lock.

I would suggest renaming `self._recently_joined_lock` to `self._room_lock` and just reusing that (it wouldn't really add any value to use two separate locks for this here)

errbot/backends/irc.py
```diff
@@ -638,7 +750,9 @@ def query_room(self, room):
         :returns:
             An instance of :class:`~IRCMUCRoom`.
         """
-        return IRCRoom(room, bot=self)
+        if room not in self._rooms:
+            self._rooms[room] = IRCRoom(room, self)
+        return self._rooms[room]
```
zoni: `self._rooms` needs to be protected with a thead lock.

errbot/backends/irc.py
```diff
@@ -651,9 +765,7 @@ def rooms(self):
         :returns:
             A list of :class:`~IRCMUCRoom` instances.
         """
-
-        channels = self.conn.channels.keys()
-        return [IRCRoom(channel) for channel in channels]
+        return self._rooms.values()
```
zoni: `self.bot._rooms` needs to be protected with a thead lock.

errbot/backends/irc.py
```diff
@@ -440,6 +473,30 @@ def on_dcc_connect(self, dcc, event):
     def on_dcc_disconnect(self, dcc, event):
         self.transfers.pop(dcc)
 
+    def on_part(self, connection, event):
+        """
+            Handler of the part IRC Message/event.
+
+            The part message is sent to the client as a confirmantion of a
```
zoni: `confirmantion` -> `confirmation`

errbot/backends/irc.py
```diff
@@ -440,6 +473,30 @@ def on_dcc_connect(self, dcc, event):
...
+            /PART command sent by someone in the room/channel.
+            If the event.source contains the bot nickname then we need to fire
+            the :meth:bot.callback_room_left event on the bot.
```
zoni: Proper Sphinx syntax here is:

```
:meth:`~errbot.backends.base.Backend.callback_room_left`
```

errbot/backends/irc.py
```diff
@@ -481,8 +539,61 @@ def on_join(self, connection, event):
...
+            if room_name not in self.bot._rooms:
+                self.bot._rooms[room_name] = IRCRoom(room_name, self.bot)
             self._recently_joined_to.add(room_name)
 
+    def on_currenttopic(self, connection, event):
+        """
+            When you Join a room with a topic set this event fires up to
+            with the topic information.
+            If the room that you join don't have a topic set, nothing happens.
+            Here is NOT the place to fire the :meth:callback_room_topic event, for
```
zoni: Sphinx syntax should be:

```
:meth:errbot.backends.base.Backend.callback_room_topic
```

errbot/backends/irc.py
```diff
@@ -440,6 +473,30 @@ def on_dcc_connect(self, dcc, event):
...
+        """
+            Handler of the part IRC Message/event.
+
+            The part message is sent to the client as a confirmantion of a
+            /PART command sent by someone in the room/channel.
+            If the event.source contains the bot nickname then we need to fire
+            the :meth:bot.callback_room_left event on the bot.
+
+            :param:
+                connection: Is an 'irc.client.ServerConnection' object
```
zoni: This param syntax is wrong, these should go on the same line, as in `:param connection` otherwise Sphinx will fail to parse it.

errbot/backends/irc.py
```diff
@@ -440,6 +473,30 @@ def on_dcc_connect(self, dcc, event):
...
+
+            :param:
+                event: Is an 'irc.client.Event' object
+                The event.source contains the nickmask of the user that
+                leave the room
+                The event.target contains the channel name
```
zoni: This param syntax is wrong, these should go on the same line, as in `:param event:` otherwise Sphinx will fail to parse it.

errbot/backends/irc.py
```diff
@@ -481,8 +539,61 @@ def on_join(self, connection, event):
...
+    def on_currenttopic(self, connection, event):
+        """
+            When you Join a room with a topic set this event fires up to
+            with the topic information.
+            If the room that you join don't have a topic set, nothing happens.
+            Here is NOT the place to fire the :meth:callback_room_topic event, for
+            that case exist on_topic.
+
+            :param:
+                connection: Is an 'irc.client.ServerConnection' object
```
zoni: This param syntax is wrong, these should go on the same line, as in `:param connection:` otherwise Sphinx will fail to parse it.

errbot/backends/irc.py
```diff
@@ -481,8 +539,61 @@ def on_join(self, connection, event):
...
+
+            :param:
+                event: Is an 'irc.client.Event' object
+                The event.arguments[0] contains the room name
+                The event.arguments[1] contains the topic of the room.
```
zoni: This param syntax is wrong, these should go on the same line, as in `:param event:` otherwise Sphinx will fail to parse it.

errbot/backends/irc.py
```diff
@@ -481,8 +539,61 @@ def on_join(self, connection, event):
...
+        """
+        room_name, current_topic = event.arguments
+        self.bot._rooms[room_name].cb_set_topic(current_topic)
+
+    def on_topic(self, connection, event):
+        """
+            On response to the /TOPIC command if the room have a topic.
+            If the room don't have a topic the event fired is on_notopic
+            :param:
+                connection: Is an 'irc.client.ServerConnection' object
```
zoni: This param syntax is wrong, these should go on the same line, as in `:param connection:` otherwise Sphinx will fail to parse it.

errbot/backends/irc.py
```diff
@@ -481,8 +539,61 @@ def on_join(self, connection, event):
...
+
+            :param:
+                event: Is an 'irc.client.Event' object
+                The event.target contains the room name.
+                The event.arguments[0] contains the topic name
```
zoni: This param syntax is wrong, these should go on the same line, as in `:param event:` otherwise Sphinx will fail to parse it.

errbot/backends/irc.py
```diff
@@ -481,8 +539,61 @@ def on_join(self, connection, event):
...
+        current_topic = event.arguments[0]
+        self.bot._rooms[room_name].cb_set_topic(current_topic)
+        self.bot.callback_room_topic(self.bot._rooms[room_name])
+
+    def on_notopic(self, connection, event):
+        """
+            This event fires ip when there is no topic set on a room
+
+            :param:
+                connection: Is an 'irc.client.ServerConnection' object
```
zoni: This param syntax is wrong, these should go on the same line, as in `:param connection:` otherwise Sphinx will fail to parse it.

errbot/backends/irc.py
```diff
@@ -481,8 +539,61 @@ def on_join(self, connection, event):
...
+
+            :param:
+                event: Is an 'irc.client.Event' object
+                The event.arguments[0] contains the room name
+        """
```
zoni: This param syntax is wrong, these should go on the same line, as in `:param event:` otherwise Sphinx will fail to parse it.

errbot/backends/irc.py
qlixed: yeah, My bad, Mixed up stage and branch and finish with some crazy stuff around. I will fix this and others and update the pull, sorry for make you work on this mixup.

zoni: No worries at all. Happy to review and help iterate on this until it's ready for merging! The IRC backend really needs some of this love so I'm very happy with these improvements you've been making to it.

gbin:
I'd love to merge this before the next release, can we fix the last remaining details ?

qlixed:
Yeah I will fin8sh the work over the next two days.

qlixed:
Delayed, but I think that now is finished, let me know if there is something pending here.
Sorry for the delay.

qlixed:
Can someone get a few minutes to check this? I have another fix in the way that partially depends on this code.

zoni:
I won't be able to review this for a bit as I'll be in Paris over the next few days and have a lot going on in my personal life at the moment. Perhaps @gbin can get a chance to look at it though.

errbot/backends/irc.py
gbin: please fix

errbot/backends/irc.py
gbin: please fix

gbin:
superseded by #739

PR #692, opened 2016-03-28
Another '!' sign

fernand0:
Maybe the proposed solution is not adequate for the project codign style?
I'm not sure about         yield "{!r}".format(suggested_config)
in webserver.py


zoni:
Well spotted. You are however mixing string interpolation with [string formatting](https://docs.python.org/3/library/string.html#format-string-syntax).

Could you make your changes use the `.format()` method as well? http://pyformat.info/ might help you if you're unfamiliar with the API.

zoni:
Superseded by #693.

PR #694, opened 2016-03-29
Flow/Conversation implementation

gbin:
This defines a new type of plugin called "BotFlow" which represent a graph of commands to be executed in sequence either manually, automatically or a mix of both.

This enables not only a way to converse with the bot towards a goal but also to skip part of the conversation (automate the flow) if the trigger of the conversation or an intermediate step provides extra context.

The graph can include botcmd, arg_botcmd and re_botcmd and can mix them.


errbot/__init__.py
```diff
@@ -397,4 +408,20 @@ def decorate(func):
...
+
+    TODO(gbin): example / docs
+    """
+    def decorate(func):
+        if not hasattr(func, '_err_flow'):  # don't override generated functions
+            func._err_flow = True
+        return func
+
+    if len(args):
+        return decorate(args[0], **kwargs)
```
ifigotin: Are you passing just args[0]?

errbot/botplugin.py
```diff
@@ -12,6 +12,22 @@
 log = logging.getLogger(__name__)
 
 
+class CommandError(Exception):
+    """
+    Use this class to report a logical error from your commands.
```
ifigotin: What is 'logical error'? Example, please.

errbot/core_plugins/flows.py
```diff
@@ -0,0 +1,59 @@
...
+from errbot.flow import FlowNode, FlowRoot
+
+
+class Flows(BotPlugin):
+    """ Management commands related to flows / conversations.
+    """
+
+    def recurse_node(self, response: io.StringIO, stack, f: FlowNode):
+        if f in stack:
+            response.write("%s⥀\n" % ("\t" * len(stack)))
```
ifigotin: Does '⥀' indicate a leaf node?

errbot/__init__.py
gbin: args[0] is the function you want to decorate, then we took the convention to only use kwargs for the optional arguments of the decorator

errbot/core_plugins/flows.py
gbin: Nope a node that loops back in the graph.

errbot/core_plugins/flows.py
```diff
@@ -0,0 +1,59 @@
...
+                    response.write(name + ": " + flow.description + "\n")
+            return response.getvalue()
+
+    @arg_botcmd('json_payload', type=str, nargs='?', default=None)
+    @arg_botcmd('flow_name', type=str)
+    def flows_start(self, mess, flow_name=None, json_payload=None):
+        """ Manually start a flow within the context of the calling user.
+        You can prefeed the flow data with a json payload.
+        Example:
+             !flows start poll_setup {\"title\":\"yeah!\",\"options\":[\"foo\",\"bar\",\"baz\"]}
```
seriesoftubes: is it always necessary to escape the quotes in the json_payload argument?

gbin: Oh, yes this is terrible... I gonna fix that in another PR with the other commands (stop etc..), I think it is a limitation coming from arg_botcmd.
Either I'll use a simple botcmd or worst case, I'll use another format.

errbot/botplugin.py
gbin: done

errbot/core_plugins/flows.plug
```diff
@@ -0,0 +1,10 @@
+[Core]
```
ifigotin: It is a little strange to have [Core] section, and Core flag equals 'true' (please excuse my ignorance).

errbot/core_plugins/flows.py
```diff
@@ -0,0 +1,59 @@
+# -*- coding: utf-8 -*-
+import io
+import json
+
+from errbot import BotPlugin, botcmd, arg_botcmd
+from errbot.flow import FlowNode, FlowRoot
+
+
+class Flows(BotPlugin):
```
ifigotin: Why extend from BotPlugin?

errbot/errBot.py
```diff
@@ -453,6 +462,16 @@ def _execute_and_send(self, cmd, args, match, mess, template_name=None):
                 reply = commands[cmd](mess, match) if match else commands[cmd](mess, args)
                 if reply:
                     self.send_simple_reply(mess, self.process_template(template_name, reply), private)
+
+            # The command is a success, check it this has not made a flow progressed
```
ifigotin: typo: 'check it this'

errbot/flow.py
```diff
@@ -0,0 +1,335 @@
...
+from errbot.backends.base import Identifier
+
+log = logging.getLogger(__name__)
+
+Predicate = Callable[[Mapping[str, Any]], str]
+
+
+class FlowNode(object):
+    """
+    This is a step in a Flow/conversation. It is linked a specific botcmd and also a "predicate".
```
ifigotin: Insert "to": It is linked to a specific botcmd

errbot/flow.py
```diff
@@ -0,0 +1,335 @@
...
+
+    The predicate is a function that tells the flow executor if the flow can enter the step without the user
+    intervention (automatically).
```
ifigotin: I assume there is a default?

errbot/flow.py
```diff
@@ -0,0 +1,335 @@
...
+        :param root: the root of this flow.
+        :param requestor: the user requesting this flow.
+        :param initial_context: any data we already have that could help executing this flow automatically.
+        """
+        self._root = root
+        self._current_step = self._root
+        self.ctx = dict(initial_context)
+        self.requestor = requestor
+
+    def next_autosteps(self) -> List[FlowNode]:
```
ifigotin: Do we have to distinguish between autosteps and steps that require user input in Flow interface?

errbot/core_plugins/flows.plug
gbin: haha indeed, never noted this.

core boolean notes that it is a core plugin for errbot (a plugin to manage errbot basically)

code section is a yaspy jargon to say this is the minimum mandatory section in the plugin descriptor

errbot/errBot.py
gbin: done

errbot/flow.py
gbin: done

errbot/flow.py
gbin: done, the default is set line 33 BTW

errbot/flow.py
gbin: as a client of the class you either want to know:

- what can I do next now ? (from executor)
- what is everything I can do  ? (from the UI to propose the user what he can do)

ifigotin: Right, but in the 'what can I do next now' category, I'd expect to see both autosteps and steps that would require user input (at least logically).

errbot/__init__.py
ifigotin: OK, I guess I just don't know the conventions here. If you only expect one argument here (a function to decorate), why pass *args?

ifigotin: OK, I guess I just don't know the conventions here. If you only expect one argument here (a function to decorate), why pass *args?

errbot/flow.py
```diff
@@ -0,0 +1,335 @@
...
+    """
+    pass
+
+
+class Flow(object):
+    """
+    This is a live Flow. It keeps context of the conversation (requestor and context).
+    Context is just a python dictionary representing the state of the conversation.
+    """
+    def __init__(self, root: FlowRoot, requestor: Identifier, initial_context):
```
seriesoftubes: if context is always a dictionary, why not annotate its type in this function declaration?

errbot/flow.py
```diff
@@ -0,0 +1,335 @@
...
+    def __init__(self, name: str, description: str):
+        """
+
+        :param name: The name of the conversation/flow.
+        :param description:  A human description of what this flow does.
+        """
+        super().__init__()
+        self.name = name
+        self.description = description
+        self.auto_triggers = []
```
seriesoftubes: Since auto_triggers seems to only be used for testing membership, would it be better if it was a set instead of a list?

errbot/flow.py
```diff
@@ -0,0 +1,335 @@
...
+
+class FlowExecutor(object):
+    """
+    This is a instance that can monitor and execute flow instances.
+    """
+    def __init__(self, bot):
+        self._lock = RLock()
+        self.flow_roots = {}
+        self.in_flight = []
+        self._pool = ThreadPool(5)
```
seriesoftubes: why must the ThreadPool always have a size of 5?  Doesn't that mean that a maximum of 5 users can concurrently be executing workflows on the same bot?

errbot/flow.py
gbin: done

errbot/flow.py
gbin: very true !

errbot/flow.py
gbin: I gonna extract that to an external constant.
This is true only for flows currently executing in automatic mode, once it has nothing more, it will free up the thread.
so 5 concurrent automatic workflows at the same time is already a lot. If it causes a contention for our users will add that as a config parameter.

gbin: Done

errbot/flow.py
gbin: I think this is what we are doing here, I am confused ...
next_steps returns all the nodes (auto or not): [node for predicate, node in self._current_step.children]
next_autosteps returns the same but only if their predicates evaluates to True: [node for predicate, node in self._current_step.children if predicate(self.ctx)]

errbot/__init__.py
gbin: as we saw online it could be 0 or 1 arg, so we need to test for this existence.

gbin:
`git merge --squash` on master.
4bfcbd07e985de4c723ed0dd30dc9dd50e52fc73

PR #695, opened 2016-03-30
Instead of chunking large slack messages, simply upload as a plain-text file

didiersmith:
This is a big UX improvement, and also simplifies the code.

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -461,12 +461,12 @@ class SlackBackend(ErrBot):
             log.debug('Message size: %d' % len(body))
 
             limit = min(self.bot_config.MESSAGE_SIZE_LIMIT, SLACK_MESSAGE_LIMIT)
-	    if len(body) > limit:
+            if len(body) > limit:
                 self.api_call('files.upload', data={
                     'channels': [to_channel_id],
                     'content': body,
                     'filename': hashlib.sha224(body).hexdigest(),
-		    'filetype': 'text',
+                    'filetype': 'text',
                 })
             else:
                 self.api_call('chat.postMessage', data={
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -1,4 +1,5 @@
 import collections
+import hashlib
 import json
 import logging
 import re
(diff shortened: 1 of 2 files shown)


errbot/backends/slack.py
```diff
@@ -460,12 +461,17 @@ def send_message(self, mess):
             log.debug('Message size: %d' % len(body))
 
             limit = min(self.bot_config.MESSAGE_SIZE_LIMIT, SLACK_MESSAGE_LIMIT)
-            parts = self.prepare_message_body(body, limit)
-
-            for part in parts:
+            if len(body) > limit:
+                self.api_call('files.upload', data={
+                    'channels': [to_channel_id],
+                    'content': body,
```
gbin: body is `self.md.convert(mess.body)` which is a slack dialect (slack_markdown_converter) for formatting the messages does the attachment still supports this formatting or you need to use the basic text formatter (text from https://github.com/errbotio/errbot/blob/master/errbot/rendering/__init__.py#L29) ?

zoni:
I disagree with the notion that this is a UX improvement.

1. This will break rendering of the message, because now you're uploading the raw Markdown-to-Slack-markup-language converted text as a plaintext file.
2. Files are displayed completely differently in slack and not at all like a regular message.
3. I haven't tested how the IRC or XMPP gateways display this, but they'll likely just link to the file on slack.com, making the bot response completely unreadable to users connecting via these gateways.

didiersmith:
My team finds the behaviour preferable than the bot flooding the channel,
but I understand that it may not be generally desirable. If you feel that
that's the case, feel free to disregard. No hard feelings :)

gbin:
We have a bug which is the root cause of all this: The markdown rendering is not "split" aware. The max length should be passed to the renderer and it should split the rendered markdown correctly (not breaking tables, fixed font, lists etc ...).

We need to do this fix for every renderer of course.  It is a known bug but we haven't found time to fix it yet.

gbin:
Also, we have a file transfer feature on errbot, maybe implementing that for slack + your plugin using it in case of long messages would be a good solution ?

didiersmith:
That sounds like a good idea. Is it this?
https://github.com/errbotio/errbot/blob/master/errbot/streaming.py

gbin:
yes, it is implemented for IRC for example here : https://github.com/errbotio/errbot/blob/master/errbot/backends/irc.py#L414

It might be easier to do for slack and it would be really cool to have it !

zoni:
Also note that plugins can call those Slack API endpoints as well, which works along the lines of:

```python
@botcmd
def mycommand(self, message, args):
  self._bot.api_call('files.upload', data={....})
```

That obviously ties it specifically to Slack though so isn't portable across backends, but for internal plugins that may be acceptable.

didiersmith:
That's a great tip, Nick. I just tried it out and it seems to work very
well. Thanks!

PR #698, opened 2016-04-01
Adds a flow_only cmd parameter.

gbin:
It does 2 things:
- blocks the execution outside of a flow for a flow_only command.
- removes it from help if the user doesn't put hidden=False explicitely.

### DIFF
--- a/tests/flow_e2e_tests.py
+++ b/tests/flow_e2e_tests.py
@@ -65,7 +65,7 @@ class TestCommands(FullStackTest):
     def test_flow_only(self):
         self.assertCommand('!a', 'a')  # non flow_only should respond.
         self.bot.push_message('!d')
-        self.assertRaises(Empty, self.bot.pop_message)
+        self.assertRaises(Empty, self.bot.pop_message, timeout=1)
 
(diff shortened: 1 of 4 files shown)


tests/flow_e2e_tests.py
```diff
@@ -59,3 +61,14 @@ def test_no_flyby_trigger_flow(self):
         flow_message = self.bot.pop_message()
         self.assertIn('You are in the flow w1', flow_message)
         self.assertEqual(len(self.bot.flow_executor.in_flight), 1)
+
+    def test_flow_only(self):
+        self.assertCommand('!a', 'a')  # non flow_only should respond.
+        self.bot.push_message('!d')
+        self.assertRaises(Empty, self.bot.pop_message)
```
zoni: I suggest using `timeout=1` on the `pop_message` call, otherwise it will wait for a 5-second timeout. This command executes so fast when it actually triggers, a 1 second timeout should be long enough to be certain it didn't trigger.

gbin: done

PR #701, opened 2016-04-02
Adds a "demo" mode to the text mode.

gbin:
This is quick hack to be able to present errbot from a text console.
The log is still shown before starting the interactive part then it is
disabled from the console.

You can still have the logging sent to a file.

### DIFF
--- a/errbot/backends/text.py
+++ b/errbot/backends/text.py
@@ -10,6 +10,7 @@ from pygments import highlight
 from pygments.formatters import Terminal256Formatter
 from pygments.lexers import get_lexer_by_name
 
+from errbot import err
 from errbot.rendering import ansi, text, xhtml, imtext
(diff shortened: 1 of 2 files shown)


errbot/backends/text.py
```diff
@@ -47,23 +48,31 @@ def __init__(self, config):
         super().__init__(config)
         log.debug("Text Backend Init.")
         self.bot_identifier = self.build_identifier('Err')
+        self.demo_mode = self.bot_config.TEXT_DEMO_MODE if hasattr(self.bot_config, 'TEXT_DEMO_MODE') else False
```
zoni: I think we should add the `TEXT_DEMO_MODE` to the `config-template.py` somewhere (way at the bottom under misc settings?) otherwise nobody will ever find it. :smile:

gbin: done

zoni:
I like it. :smiley:

PR #708, opened 2016-04-07
CLI storage manipulation.

### LINKED ISSUE #608: RFC: Configure plugins outside of chat
Some people have expressed an interest in being able to configure all aspects of the bot programmatically (mainly by using configuration management systems such as ansible, chef, puppet, etc) without actually having to chat with the bot.

Errbot configuration is currently split between two major areas:

1. Stuff in `config.py`. This is almost entirely backend configuration which deals with the chat network to connect to, account credentials, storage selection, etc, and thus not suitable as dynamic runtime configuration. Changes require a full restart.

2. Plugin specific configuration. This is currently kept in a database (shelve-based by default, other options available on master) and can only be configured by bot admins, through chatting with the bot.

The `config.py` configuration file already plays nice with configuration management systems (they can template the file and restart errbot on changes). Plugin configuration on the other hand doesn't, at the moment.

I propose adding a command-line program (either standalone or as part of the `errbot` command directly, the latter being preferred) which is capable of setting the configuration for plugins.

Example usage
--------------------

Get the plugin configuration for the webserver plugin:

```
$ errbot --get-plugin-config Webserver
{'HOST': '0.0.0.0',
 'PORT': 3141,
 'SSL': {'certificate': '',
         'enabled': False,
         'host': '0.0.0.0',
         'key': '',
         'port': 3142}}
```

Set a different configuration for the webserver plugin:

```
$ echo "{'HOST': 'localhost', 'PORT': 1234, 'SSL': {'certificate': '', 'enabled': False, 'host': '0.0.0.0', 'key': '', 'port': 3142}} | errbot --set-plugin-config Webserver
Webserver plugin config updated
```

Similar to passing `--config` to errbot in order to start it, the command would take `--config` here as well when needed:

```
$ errbot --config /opt/errbot/ --get-plugin-config Webserver
{'HOST': '0.0.0.0',
 'PORT': 3141,
 'SSL': {'certificate': '',
         'enabled': False,
         'host': '0.0.0.0',
         'key': '',
         'port': 3142}}
```

Implementation
--------------------

Under the hood this plugin should:

1. Check `config.py` to know where storage is and what storage backend is used.
2. Load storage and update it directly if errbot **isn't running**.
3. Connect to errbot on a socket (somewhere in it's data directory) and have errbot proxy the configuration change if it **is running**.

This would allow configuration management systems to update the configuration with some specific, *"static"* configuration. It would let people do this *before* even starting the bot, so plugins are configured as desired the first time it even starts up.

Pitfalls
--------

Some issues I see, which may require more thinking about:

1. When the bot is running, any configuration that is set can be verified by the plugin, similar to how `!config` verifies it. If the bot isn't running this obviously isn't possible. It may be desirable to have a flag which tells errbot to only set the config if it can connect to a running instance.

gbin:
Those new CLI parameters allows the user to manipulate the plugin
configuration, states but also tweak errbot internals.

Fixes #608.

### DIFF
--- a/errbot/err.py
+++ b/errbot/err.py
@@ -22,8 +22,8 @@ import logging
 import argparse
 from os import path, sep, getcwd, access, W_OK
 from platform import system
-from .plugin_wizard import new_plugin_wizard
-from .version import VERSION
+from errbot.plugin_wizard import new_plugin_wizard
+from errbot.version import VERSION
 
 PY3 = sys.version_info[0] == 3
 PY2 = not PY3
@@ -57,7 +57,7 @@ log = logging.getLogger(__name__)
 def debug(sig, frame):
     """Interrupt running process, and provide a python prompt for
     interactive debugging."""
-    d = {'_frame': frame}  # Allow access to frame object.
+    d = {'_frame': frame}  # import sys Allow access to frame object.
     d.update(frame.f_globals)  # Unless shadowed by global
     d.update(frame.f_locals)
 
@@ -146,6 +146,15 @@ def get_config(config_path):
     return config
 
 
+def _read_dict():
+    import collections
+    new_dict = eval(sys.stdin.read())
+    if not isinstance(new_dict, collections.Mapping):
+        raise ValueError("A dictionary written in python is needed from stdin. Type=%s, Value = %s" % (type(new_dict),
+                                                                                                       repr(new_dict)))
+    return new_dict
+
+
 def main():
 
     execution_dir = getcwd()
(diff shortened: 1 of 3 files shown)


errbot/err.py
```diff
@@ -57,7 +57,7 @@ def foo(param='canary'):
 def debug(sig, frame):
     """Interrupt running process, and provide a python prompt for
     interactive debugging."""
-    d = {'_frame': frame}  # Allow access to frame object.
+    d = {'_frame': frame}  # import sys Allow access to frame object.
```
ifigotin: Is this accidental: "import sys"?

errbot/err.py
```diff
@@ -165,6 +174,16 @@ def main():
     mode_selection.add_argument('-l', '--list', action='store_true', help='list all available backends')
     mode_selection.add_argument('--new-plugin', nargs='?', default=None, const='current_dir',
                                 help='create a new plugin in the specified directory')
+    # storage manipulation
+    mode_selection.add_argument('--storage-set', nargs=1, help='DANGER: Delete the given storage namespace '
+                                                               'and set the python dictionary expression '
+                                                               'passed on stdin.')
+    mode_selection.add_argument('--storage-merge', nargs=1, help='DANGER: Merge in the python dictionary expression '
+                                                                 'passed on stdin into the given storage namespace.')
+    mode_selection.add_argument('--storage-get', nargs=1, help='Dumps the given storage namespace in a '
```
ifigotin: nit: you use 3rd person here, but not earlier: delete vs Deletes, Dump vs Dumps.

ifigotin:
Looks good to me overall.

errbot/err.py
gbin: done

errbot/err.py
gbin: done

PR #714, opened 2016-04-12
Fix message type detection for Telegram backend

dveeden:
both private and group messages are of type telegram.chat.Chat


zoni:
Interesting. It appears the API of `python-telegram-bot` changed with 3.1 (released 2015-12-16) but I hadn't run into this issue myself yet.

I can confirm it was incorrect and this fixes it though, so, awesome, thanks!

PR #718, opened 2016-04-15
XMPP: build_identifier can now build Rooms

### LINKED ISSUE #717: xmpp channel message failing
`self.send(self.build_identifier('dev@conference.jabber.domain.tld'), "test",)`  fails with 

`WARNING errbot.backends.xmpp Received error message: <message to="devbot@jabber.domain.tld/9oednv6no3" type="error" xml:lang="en" from="dev@conference.jabber.domain.tld"><body>test</body><error code="400" type="modify"><bad-request xmlns="urn:ietf:params:xml:ns:xmpp-stanzas" /></error></message>`

Tested with errbot 4.0.3 on openfire 4.0.2

gbin:
but also RoomOccupants by querying the XMPP discovery service.

This fixes #717

### DIFF
--- a/errbot/backends/xmpp.py
+++ b/errbot/backends/xmpp.py
@@ -538,11 +538,10 @@ class XMPPBackend(ErrBot):
                 log.debug('This is a room ! %s', txtrep)
                 return self.query_room(txtrep)
             if category == 'client' and 'http://jabber.org/protocol/muc' in info['disco_info']['features']:
-                node, domain, resource = split_identifier(txtrep)
                 log.debug('This is room occupant ! %s', txtrep)
-                return XMPPRoomOccupant(node, domain, resource, self.query_room(node + '@' + domain))
+                return self._build_room_occupant(txtrep)
         log.debug('This is a person ! %s', txtrep)
-        return XMPPPerson(*split_identifier(txtrep))
+        return self._build_person(txtrep)
 
     def build_reply(self, mess, text=None, private=False):
         """Build a message for responding to another message.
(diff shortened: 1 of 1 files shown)


errbot/backends/xmpp.py
```diff
@@ -1,5 +1,6 @@
 import logging
 import sys
+from functools import lru_cache
```
zoni: Python 2 compatibility requires:

``` python
try:
    from functools import lru_cache
except ImportError:
    from backports.functools_lru_cache import lru_cache
```

errbot/backends/xmpp.py
```diff
@@ -31,6 +32,9 @@
     """)
     sys.exit(-1)
 
+# LRU to cache the JID queries.
+IDENTIFIERS_LRU = 1000
```
zoni: According to the documentation for `lru_cache`, it performs better when using a power of 2, so I suggest changing this to `1024`

errbot/backends/xmpp.py
gbin: done

errbot/backends/xmpp.py
gbin: done

zoni:
Code-wise this looks OK, but I don't have an XMPP setup handy to test this out against a real server right now. I'll try to do that over the weekend.

gbin:
cool, I made a last minute change, I forgot to reuse the _build_room_occupant and _build_person in build_identifier itself.

gbin:
Thanks for the review, I tested it on blah.im but I hope I did not hit any specificity for this server.

PR #727, opened 2016-04-28
Allow multiline tables.

ifigotin:


ifigotin:
Zoni, please keep in mind that I'm new to Python, so you may see some odd code :)

PR #738, opened 2016-05-05
Talk about XMPP resource

cweiske:


gbin:
Tested on blah.im it seems to work. Thanks, merging.

Issue #711, opened 2016-04-09
ACL allowrooms not working with Slack, or format unclear.


jakosky:
I want to restrict Salt:* commands to a specific room.  I have tried 3 different formats for the Slack room: cfops, #cfops, and its Slack channel id G0YRZBSMS. All respond with "You're not allowed to access this command from this room".

I am also not allowing these commands as private messages, which appears to work fine.

# G0YRZBSMS
17:30:36 DEBUG    errbot.backends.slack     Processing slack event: {u'text': u'Cob, salt stg .* test.ping', u'ts': u'1460223035.000046', u'user': u'U02FD73UX', u'team': u'T02FBFM8R', u'type': u'message', u'channel': u'G0YRZBSMS'}
17:30:36 DEBUG    errbot.backends.slack     Saw an event: {u'channel': u'G0YRZBSMS',
 u'team': u'T02FBFM8R',
 u'text': u'Cob, salt stg .* test.ping',
 u'ts': u'1460223035.000046',
 u'type': u'message',
 u'user': u'U02FD73UX'}
17:30:36 DEBUG    errbot.backends.slack     Escaped IDs event text: Cob, salt stg .* test.ping
17:30:36 DEBUG    errbot.errBot             *** frm = #cfops/jayjakosky
17:30:36 DEBUG    errbot.errBot             *** username = U02FD73UX
17:30:36 DEBUG    errbot.errBot             *** text = Cob, salt stg .* test.ping
17:30:36 DEBUG    errbot.errBot             Called with alternate prefix 'Cob'
17:30:36 DEBUG    errbot.plugins.ACLS       Check salt for ACLs.
17:30:36 INFO     errbot.plugins.ACLS       Matching ACL {'allowprivate': False, 'allowrooms': 'G0YRZBSMS'} against username @jayjakosky for command Salt:salt
17:30:36 DEBUG    errbot.backends.slack     Sending channel message to cfops (G0YRZBSMS)
17:30:36 DEBUG    errbot.backends.slack     Message size: 56
17:30:36 INFO     errbot.errBot             Command None blocked or deferred.
17:30:36 DEBUG    errbot.errBot             Triggering callback_message on Help
17:30:36 DEBUG    errbot.errBot             Triggering callback_message on Utils
17:30:36 DEBUG    errbot.errBot             Triggering callback_message on Health
17:30:36 DEBUG    errbot.errBot             Triggering callback_message on ACLS
[...]

PR #740, opened 2016-05-06
Allow simple strings in ACL rules

cweiske:
Previously only tuples or lists were allowed in ACL configurations like
"allowusers" and "allowrooms", while the code iterated over characters
in case of strings.
This led to confusion (like https://github.com/errbotio/errbot/issues/711).

The patch casts incoming strings to tuples, so that you can do

  ACCESS_CONTROLS_DEFAULT = {'allowusers': 'gbin@localhost'}

in addition to

  ACCESS_CONTROLS_DEFAULT = {'allowusers': ('gbin@localhost',)}

now.

### DIFF
--- a/errbot/core_plugins/acls.py
+++ b/errbot/core_plugins/acls.py
@@ -1,4 +1,5 @@
 import fnmatch
+from types import StringType
 from errbot import BotPlugin, cmdfilter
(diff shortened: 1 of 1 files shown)


cweiske:
Sorry, I have no idea how should be in python3.

PR #743, opened 2016-05-06
Cweiske acl string cast

gbin:


zoni:
I'd like to see some tests for this new behavior (in [test_access_controls](https://github.com/errbotio/errbot/blob/master/tests/base_backend_tests.py#L527-L710)) so we know if/when we accidentally break this.

PR #752, opened 2016-05-12
Replace alternate bot prefix into command documentation

### LINKED ISSUE #741: command doc replaces the wrong '!' in case of alternative prefix.
``` python
     def get_doc(self, command):
        """Get command documentation
        """
        if not command.__doc__:
            return '(undocumented)'
        if self.prefix == '!':
            return command.__doc__
        return command.__doc__.replace('!', self.prefix)
```

so you change your prefix from '!' to '$' and your comment is 'hello !'... you'll end up with 'hello @' ?

boidolr:
Fixes #741

### DIFF
--- a/errbot/errBot.py
+++ b/errbot/errBot.py
@@ -17,6 +17,7 @@ from datetime import datetime
 import difflib
 import inspect
 import logging
+import re
 import traceback
 
 from errbot import CommandError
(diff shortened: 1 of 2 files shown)


boidolr:
I'll add the lines, just running all tests again to make sure I haven't missed anything else.

PR #759, opened 2016-05-20
Fix hipchat room.occupants error

mecarus:
when calling occupants on hipchat room, errbot gets an exception because hipchat_user is not being specified when passing in person dictionary. Additionally, resource (user's name) does not exist in xmpp_jid, so needs to be checked for and 'name' attribute should be set as resource directly

### DIFF
--- a/errbot/backends/hipchat.py
+++ b/errbot/backends/hipchat.py
@@ -94,8 +94,12 @@ class HipChatRoomOccupant(XMPPRoomOccupant):
             for k, v in hipchat_user.items():
                 setattr(self, k, v)
             # Quick fix to be able to all the parent.
-            node_domain, resource = hipchat_user['xmpp_jid'].split('/')
-            node, domain = node_domain.split('@')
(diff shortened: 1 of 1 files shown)


errbot/backends/hipchat.py
```diff
@@ -94,8 +94,12 @@ def __init__(self, node=None, domain=None, resource=None, room=None, hipchat_use
             for k, v in hipchat_user.items():
                 setattr(self, k, v)
             # Quick fix to be able to all the parent.
-            node_domain, resource = hipchat_user['xmpp_jid'].split('/')
-            node, domain = node_domain.split('@')
+            if '/' in hipchat_user[u'xmpp_jid']:
```
gbin: please remove the redundant u'  (they will be added back during the py2 conversion)

gbin:
Thx for the fix, I merge it once the small issue is addressed

zoni:
Cherry-picked and amended with the misplaced `u'` markers removed.

PR #762, opened 2016-05-21
Adds a GPLv3 exception for Errbot plugins.

gbin:
This was the original intent, the goal here is to avoid forcing the entire ecosystem under GPLv3 or making them run with a "main" or a separate process. The GPLv3 is not totally clear about it, so I would prefer to totally clarify that.

### DIFF
--- a/gplv3-exceptions.txt
+++ b/gplv3-exceptions.txt
@@ -1,4 +1,5 @@
-As a special exception, the copyright holders of Errbot hereby grant permission 
-for plug-ins and script add-ons to be used with Errbot, provided that you also 
-meet the terms and conditions of the licenses of those plug-ins and script
-add-ons.
+As a special exception, the copyright holders of Errbot hereby grant permission
+for plug-ins, scripts or add-ons not bundled or distributed as part
+of Errbot itself and potentially licensed under a different license, to be
+used with Errbot, provided that you also meet the terms and conditions of the
+licenses of those plug-ins, scripts or add-ons.
--- a/gplv3-exceptions.txt
+++ b/gplv3-exceptions.txt
@@ -1,0 +1,4 @@
+As a special exception, the copyright holders of Errbot hereby grant permission 
+for plug-ins and script add-ons to be used with Errbot, provided that you also 
+meet the terms and conditions of the licenses of those plug-ins and script
+add-ons.


gbin:
@zoni @atalyad @sijis @jamesob if you could +1 this, for the others, let them speak now or forever hold their peace.
Thank you !

sijis:
+1

jamesob:
+1!

zoni:
I will come back with a detailed response on this somewhere over the next couple of days. In short, while I fully support the goal of this 100%, I think there's some additional things that we should discuss and take into account.

gbin:
@zoni, sure, that's why I opened the discussion !

gbin:
Any update on this ?

zoni:
Yes, apologies for the delay.

So there's some conflicting reports on whether the GPL extends to plugins which are loaded by the software (errbot) itself, but the official stance by the Free Software Foundation (FSF) is that it does. This means that currently, all plugins in the ecosystem should be GPL-licensed. :disappointed:

We can certainly add the proposed license exception to clarify our position that this was never the intent. We will have to be very careful with the phrasing however as errbot itself consists largely of plugins and we would not want this exception to apply to that code.

The other option that I think is worth considering is switching the license over entirely. GPL is a complicated beast which takes away many rights in order to ensure the software remains open and free. The alternative is re-licensing under MIT, BSD (either 2- or 3-clause) or Apache 2.0.

Very broadly speaking, these are highly permissive licenses which would allow anyone to do basically anything with errbot, provided some copyright notices and disclaimers remain intact. Due to their highly permissive nature, it would free users of errbot from many license-related issues. It may result in somebody taking all of our code and distributing it in modified form (even commercially) without contributing back to us, but it may also result in novel applications where errbot is in some way embedded into another product which is not otherwise possible with GPL licensing.

It is perhaps worth pointing out that for example Django is BSD licensed and Ruby on Rails is MIT.

Finally, and this applies especially with re-licensing, legally speaking we would need approval from every single person who has contributed code to errbot before we could make changes to the license because we have never asked contributors to sign over the copyright of their contributions to us.

Large projects tend to have Contributor License Agreements (CLAs) for this which is something to keep in mind as well as errbot grows further in popularity, but more of a hassle than it is worth at this point I think.

gbin:
As per discussion on the chat, I gonna extend the notification to all the contributors of Errbot.
Please +1 this clarification of the Errbot license for our plugin designers !

- [x] Guillaume Binet
- [x] Nick Groenen @zoni 
- [ ] Raphael Boidol @boidolr
- [ ] Asen Lekov @L3K0V 
- [ ] Marek Suppa @mrshu 
- [ ] Jason Ziglar @jasedit
- [x] James O'Beirne @jamesob 
- [x] Sijis Aviles @sijis 
- [ ] Tom Neyland @TomNeyland
- [ ] Ciaran Gultnieks @CiaranG 
- [ ] Andrii Kostenko @gugu 
- [ ] Andre Van Der Merwe @andrevdm
- [ ] Travis Veazey @Kromey 
- [ ] Mike Burke @mRB0
- [ ] Mike Kazantsev @mk-fg
- [ ] Milan Kubík @apophys
- [ ] Joshua Tobin @joshuatobin 
- [ ] Chris Becker @Cbeck527

jasedit:
+1

L3K0V:
+1

mk-fg:
+1

mrshu:
+1

apophys:
+1

andrevdm:
+1

joshuatobin:
+1

TomNeyland-FD:
👍

TomNeyland:
👍

gplv3-exceptions.txt
```diff
@@ -0,0 +1,4 @@
+As a special exception, the copyright holders of Errbot hereby grant permission 
+for plug-ins and script add-ons to be used with Errbot, provided that you also 
+meet the terms and conditions of the licenses of those plug-ins and script
+add-ons.
```
zoni: To avoid any ambiguity, I think we should change the wording to:

> As a special exception, the copyright holders of Errbot hereby grant permission
> for plug-ins and script add-ons not bundled and distributed as part of errbot itself,
> potentially licensed under a license other than the GPL, to be used with Errbot,
> provided that you also meet the terms and conditions of the licenses of those
> plug-ins and script add-ons.

This emphasizes that it applies to third-party plugins rather than core plugins shipped as part of errbot itself and further emphasizes that it's OK to license said plugins under a license other than the GPL.

ifigotin: +1

ifigotin:
+1

Kromey:
+1
Always felt the "viral" nature of the GPL was overly restrictive for a "free" license; don't mind the rest of it, but if someone wants to write a plug-in with a different license I'm all for making it crystal clear that we support that!

boidolr:
+1

gplv3-exceptions.txt
gbin: Done, I reworded also a little bit to be sure we include all sorts of external contributions (install scripts, backends etc...)

gbin:
As we have now less than 10 lines of code impacted, I gonna merge this one, Thanks y'all for your support !

Issue #600, opened 2016-02-15
[slack] [documentation] Making @botname work as a prefix


ytjohn:
If I want to let people (on slack) to address err by `@botname`, you can't just put `@botname` in `BOT_ALT_PREFIXES`, it won't work. I did a lot of searching, but couldn't find anyone else mentioning this prior. 

I enabled debugging and found that slack replaces `@username` with `<@UNIQID>`.  So I put that as an alternate prefix, and that started working.

I don't know enough about slack to know if that id is persistent. If it is persistent, than this can just be a documentation addition. Maybe there is an better way to get this id other than just watching err's debug log.

If it's not persistent, or if others think it's worthwhile, a better approach would be to have an option to append `<@UNIQID>` to the `BOT_ALT_PREFIXES` upon connection to slack. 

Thoughts?

sijis:
@ytjohn That ID from slack, i believe, is persistent.

What if you just add '<botname>' to `BOT_ALT_PREFIXES`?

I will double-check later, but I have a config that (i think) allows the bot to respond with messages where referenced as <@sijis> or <sijis>.

ytjohn:
@sijis 

Indeed, I originally had `BOT_ALT_PREFIXES = ('botname', '@botname'). `botname` works, `@botname` does not. That's because slack rewrites the message text to show the uid of the referenced user.  

```
2016-02-15 02:59:05,234 DEBUG    errbot.backends.slack     Saw an event: {u'channel': 
 u'C0chand',
 u'team': u'T0teamid',
 u'text': u'<@U0botid>: help',
 u'ts': u'1455505144.000063',
 u'type': u'message',
 u'user': u'U0myid'}
```

zoni:
Slack ***may*** change those IDs in the future, but I think it's pretty safe to assume that they won't change.

From an end-user perspective, I think the ideal situation is that `@mentioning` the box just works out of the box on both Slack and HipChat (as they both use the very same semantics here), without a user even having to bother with setting up a special `BOT_ALT_PREFIXES` setting for it.

This should be easyish to do. If the Slack API (and same goes for HipChat) exposes some way to get our unique identifier, then we could just insert that into the `BOT_ALT_PREFIXES` property as we're starting up (so somewhere in `__init__()` of the `SlackBackend` and `HipChatBackend`.)

sijis:
@zoni I'm taking a stab at this and I put the following `serve_once` (which works!)
```
[...]

PR #768, opened 2016-06-01
slack: Add botid to alternative prefixes

sijis:
This is related to #600.

Completely not finished.

Needs feedback

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -243,6 +243,26 @@ class SlackBackend(ErrBot):
             )
         return response
 
+    def update_alternate_prefixes(self):
+        """Converts BOT_ALT_PREFIXES to use the slack ID instead of name
+
+        Slack only acknowledges direct callouts `@username` in chat if referred
+        by using the ID of that user.
+        """
+        # convert BOT_ALT_PREFIXES to a list
+        try:
+            bot_prefixes = self.bot_config.BOT_ALT_PREFIXES.split(',')
+        except AttributeError:
+            bot_prefixes = list(self.bot_config.BOT_ALT_PREFIXES)
+
+        converted_prefixes = []
+        for prefix in bot_prefixes:
+            converted_prefixes.append('<@{0}>'.format(self.username_to_userid(prefix)))
+
+        self.bot_config.BOT_ALT_PREFIXES = tuple(converted_prefixes)
+        self.bot_alt_prefixes = tuple(x.lower() for x in self.bot_config.BOT_ALT_PREFIXES)
+        log.debug('Converted BOT_ALT_PREFIXES: %s', self.bot_config.BOT_ALT_PREFIXES)
+
     def serve_once(self):
         self.sc = SlackClient(self.token)
         log.info("Verifying authentication token")
@@ -253,10 +273,7 @@ class SlackBackend(ErrBot):
         self.bot_identifier = SlackPerson(self.sc, self.auth["user_id"])
 
         # Inject bot identity to alternative prefixes
-        alt_prefixes = list(self.bot_config.BOT_ALT_PREFIXES)
-        alt_prefixes.append('<@{0}>'.format(self.bot_identifier.userid))
-        self.bot_config.BOT_ALT_PREFIXES = tuple(alt_prefixes)
-        self.bot_alt_prefixes = tuple(x.lower() for x in self.bot_config.BOT_ALT_PREFIXES)
+        self.update_alternate_prefixes()
 
         log.info("Connecting to Slack real-time-messaging API")
         if self.sc.rtm_connect():
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -439,8 +439,7 @@ class SlackBackend(ErrBot):
 
     def channelname_to_channelid(self, name):
         """Convert a Slack channel name to its channel ID"""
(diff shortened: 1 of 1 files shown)


gbin:
This is cool.

I would do it a little bit smarter: let people put `@errbot` (or whatever the name of their bot is) in BOT_ALT_PREFIXES, detect that and replace it with the id syntax !

So for the users, nothing changed and the general behavior of errbot is uniform.

sijis:
That's a great idea. I'll fix that up.

sijis:
@gbin Make the suggested updates. Am I on the right track here?

errbot/backends/slack.py
```diff
@@ -394,17 +417,21 @@ def _message_event_handler(self, event):
 
     def userid_to_username(self, id_):
         """Convert a Slack user ID to their user name"""
-        user = [user for user in self.sc.server.users if user.id == id_]
+        users = self.api_call('users.list')
```
gbin: what is the difference between self.sc.server.users and self.api_call('users.list') ?

errbot/backends/slack.py
```diff
@@ -394,17 +417,21 @@ def _message_event_handler(self, event):
...
         if not user:
             raise UserDoesNotExistError("Cannot find user with ID %s" % id_)
-        return user[0].name
+        return user[0]
 
     def username_to_userid(self, name):
         """Convert a Slack user name to their user ID"""
-        user = [user for user in self.sc.server.users if user.name == name]
+        if name.startswith('@'):
+            name = name[1:]
```
gbin: name.lstrip('@') ?

gbin:
looks better ! just some comments

errbot/backends/slack.py
```diff
@@ -243,6 +243,26 @@ def api_call(self, method, data=None, raise_errors=True):
...
+        try:
+            bot_prefixes = self.bot_config.BOT_ALT_PREFIXES.split(',')
+        except AttributeError:
+            bot_prefixes = list(self.bot_config.BOT_ALT_PREFIXES)
+
+        converted_prefixes = []
+        for prefix in bot_prefixes:
+            converted_prefixes.append('<@{0}>'.format(self.username_to_userid(prefix)))
+
+        self.bot_config.BOT_ALT_PREFIXES = tuple(converted_prefixes)
```
gbin: don't touch this one I would say, only change self.bot_alt_prefixes

errbot/backends/slack.py
sijis: When trying to introspect `self.sc.server.users` I was unable to at all when testing. Using the `api_call('users.list')` I was able to. That was the main reason around it.

The groups appear to work with `sc.server.groups`. I can go back and leave that original call and see if that works.

errbot/backends/slack.py
sijis: Ya good thought!

errbot/backends/slack.py
sijis: Noted!

zoni:
There's one more thing we should probably tackle in this PR. With the bot prefix set up, when inviting the bot into a channel, it'll receive a message originating from itself, `@botname has joined #channelname` (or something close to it anyway, I'm writing this from memory).

Then you get a `Command "has" / "has joined" not found.` error in the chat. We should probably filter that specific message out in this PR or later in a separate one.

errbot/backends/slack.py
gbin: You changed the contract of the method doing that and maybe broke stuff down the line:
When I compare the 2 lists:
image
and
image

sijis:
Breaking that contract was not the intention. I thought it was a fairly safe swap as the method  returned the same response (from what I gathered) even though the way it gets the user list is different.

errbot/backends/slack.py
sijis: I figured out the main differences:
`api_call('users.list')` can be called without having the bot connected via rtm.
`sc.server.users` only has values after `sc.rtm_connect()` is executed.

Because I was altering the bot_prefixes beofre `sc.rtm_connect()` was executed, I had to use `api_call()` to get a list of users.

To preserve the original method calls, I simply moved the altering of bot_alternatives to just after `sc.rtm_connect()` is called.

sijis:
@zoni i'll test the behavior you note and put in a filter for it.

errbot/backends/slack.py
```diff
@@ -243,6 +243,25 @@ def api_call(self, method, data=None, raise_errors=True):
...
+        try:
+            bot_prefixes = self.bot_config.BOT_ALT_PREFIXES.split(',')
+        except AttributeError:
+            bot_prefixes = list(self.bot_config.BOT_ALT_PREFIXES)
+
+        converted_prefixes = []
+        for prefix in bot_prefixes:
+            converted_prefixes.append('<@{0}>'.format(self.username_to_userid(prefix)))
+
+        self.bot_alt_prefixes = tuple(x.lower() for x in self.bot_config.BOT_ALT_PREFIXES)
```
sijis: I should probably not set to lower() as the case insensitive filters have already applied, so the list here is in the expected case.

machbio:
@sijis need to update the documents on http://errbot.io/en/latest/user_guide/configuration/slack.html - regarding the slack bot alt prefixes

PR #772, opened 2016-06-04
Added an errbot --init and simplified config.

gbin:
Now all it takes to have an errbot up and running in text mode/development environment is:

1. mkvirtualenv errbot
2. pip install errbot
3. mkdir errbotdir; cd errbotdir
4. errbot --init
5. errbot

And you have a ready to use development environment with an example
plugin to get you started.


docs/user_guide/setup.rst
```diff
@@ -4,128 +4,107 @@ Setup
...
-
-1. You can generate it directly from your errbot installation with:
+Once you have installed errbot and did `errbot --init`, you will have to tweak the generated `config.py` to connect
+to your desired chat network.
 
-    `python -c "import errbot;import os;import shutil;shutil.copyfile(os.path.dirname(errbot.__file__) + os.path.sep + 'config-template.py', 'config.py')"`
+You can use https://raw.githubusercontent.com/errbotio/errbot/master/errbot/config-template.py as a base
+ for your `config.py`.
 
-2. Or you can download the :download:`config-template.py` file manually and save it as `config.py`.
```
zoni: I think it's better to keep the `:download:` version instead of linking directly to GitHub, as this will ensure it links to the version of `config-template.py` that matches the version of the docs you're browsing.

errbot/err.py
```diff
@@ -197,6 +196,44 @@ def main():
...
+            shutil.copyfile(os.path.join(templates_dir, 'example.py'), os.path.join(example_plugin_dir, 'example.py'))
+            print('Your Errbot directory has been correctly initialized !')
+            if base_dir == os.getcwd():
+                print('Just do "errbot" and it should start in text/development mode.')
+            else:
+                print('Just do "cd %s" then "errbot" and it should start in text/development mode.' % data_dir)
+            sys.exit(0)
+        except Exception as e:
+            print('The initialization of your errbot directory failed: %s' % e)
+            sys.exit(-11)
```
zoni: I would `sys.exit(1)` instead.

errbot/templates/initdir/config.py.tmpl
```diff
@@ -0,0 +1,18 @@
...
+
+BOT_DATA_DIR = '{{ data_dir }}'
+BOT_EXTRA_PLUGIN_DIR = '{{ extra_plugin_dir }}'
+BOT_LOG_FILE = '{{ log_path }}'
+BOT_LOG_LEVEL = logging.DEBUG
+
+BOT_ADMINS = ('CHANGE ME', )
+AUTOINSTALL_DEPS = True
+
+TEXT_DEMO_MODE = False  # Put that to True if you want to suppress the debug output in Text mode.
```
zoni: I think we should default to true here. I think that's a much nicer user experience by default as the bot output is a lot easier to read like that.

errbot/templates/initdir/example.plug
```diff
@@ -0,0 +1,9 @@
+[Core]
+Name = Example
+Module = example
+
+[Documentation]
+Description = This is a simple plugin example to get you started.
+
+[Python]
+Version = 2+
```
zoni: Missing an EOL here.

errbot/templates/initdir/config.py.tmpl
gbin: Ok, I'll make another change because I linked that to the graphic mode too, maybe I'll split it... we'll see... I put to to True by default meanwhile.

docs/user_guide/setup.rst
gbin: Done

errbot/err.py
gbin: Done

errbot/templates/initdir/example.plug
gbin: Done

PR #774, opened 2016-06-06
Refactor main modules

gbin:
See details in individual commits but those are mainly historic stuff:
main.py was... the old main which is more like the bootstrapping of Errbot so "bootstrap.py"
err.py was... the original errbot command, it is now "cli.py"
ErrBot.py was... the original original main which is now more the core of Errbot so "core.py"

I also split out log initialization into a separate module, we hack some stuff from there here and there (touching the handlers etc...) I'll try to clean that up but in a separate PR.


gbin:
I don't want to wait too much for this one, I'll go ahead an merge it. We will mop up some potential compatibility issues on external backends on master.

PR #776, opened 2016-06-07
hipchat rooms should handle paging

namanbharadwaj:
Currently, if you specify a room by name which isn't in the first 100 returned by the API, errbot fails to initialize. This uses the `contents` method on the `RoomsCollection` of the `hypchat` library to page automatically.

### DIFF
--- a/errbot/backends/hipchat.py
+++ b/errbot/backends/hipchat.py
@@ -444,7 +444,7 @@ class HipchatBackend(XMPPBackend):
             log.debug("Room specified by JID, looking up room name")
             rooms = self.conn.hypchat.rooms(expand='items').contents()
             try:
-                name = [r['name'] for r in rooms['items'] if r['xmpp_jid'] == room][0]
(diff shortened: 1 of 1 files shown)


errbot/backends/hipchat.py
```diff
@@ -442,7 +442,7 @@ def query_room(self, room):
         """
         if room.endswith('@conf.hipchat.com'):
             log.debug("Room specified by JID, looking up room name")
-            rooms = self.conn.hypchat.rooms(expand='items')
+            rooms = self.conn.hypchat.rooms(expand='items').contents()
```
gbin: This fix is not symmetric with the other one, I guess one or the other is buggy ?
Here we to rooms['items'] but on the other we don't ?

namanbharadwaj: sorry - fixed! this one was buggy; I miscopied the patch from my virtualenv

zoni:
Looks good now, thanks!

Issue #771, opened 2016-06-04
Storing mess.frm on Slack with Shelf storage backend does not work


br0ziliy:
Following traceback is generated when storing an Identifier using Slack backend:

```
09:10:32 ERROR    errbot.errBot             An error happened while processing a message ("some message"): Traceback (most recent call last):
  File "/home/devops_bot/ve/lib/python2.7/site-packages/errbot/errBot.py", line 482, in _execute_and_send
    for reply in replies:
  File "/home/devops_bot/ve/lib/python2.7/site-packages/errbot/__init__.py", line 386, in wrapper
    yield func(self, mess, *func_args, **func_kwargs)
  File "/opt/err-plugins/err-ansible/ansible.py", line 111, in ansible_cmd
    raw_result = tasks.run_task(self, ansible_cmd, _from)
  File "/opt/err-plugins/err-ansible/lib/tasks.py", line 22, in run_task
    bot['tasks'] = tasklist
  File "/home/devops_bot/ve/lib/python2.7/site-packages/errbot/storage/__init__.py", line 50, in __setitem__
    return self._store.set(key, item)
  File "/home/devops_bot/ve/lib/python2.7/site-packages/errbot/storage/shelf.py", line 28, in set
    self.shelf[key] = value
  File "/usr/lib64/python2.7/shelve.py", line 132, in __setitem__
    p.dump(value)
PicklingError: Can't pickle <type 'instancemethod'>: attribute lookup __builtin__.instancemethod failed
```

The code I use:
```
    # botplugin.py
    @arg_botcmd('ansible_cmd', type=str, \
                help="param for the command")
    def ansible(self, mess, param1=None):
        ...
        _from = mess.frm
        raw_result = tasks.run_task(self, ansible_cmd, _from)
        return result

    # tasks.py
def run_task(bot, cmd, _from):
      bot['tasks'] = tasklist
      tasklist[cmd] = _from
      bot['tasks'] = tasklist # fails here
```
If I try to pickle the mess.frm with Slack backend (for debugging):

```
Traceback (most recent call last):
  File "/home/devops_bot/ve/lib/python2.7/site-packages/errbot/errBot.py", line 482, in _execute_and_send
    for reply in replies:
  File "/home/devops_bot/ve/lib/python2.7/site-packages/errbot/__init__.py", line 386, in wrapper
    yield func(self, mess, *func_args, **func_kwargs)
[...]

PR #783, opened 2016-06-09
Slack: Fixed identifiers pickling.

gbin:
This is very very ugly, but at least limited to Slack. If you have another less ugly idea feel free to experiment.


gbin:
This fixes #771

gbin:
Finally found a little less ugly way by registering the latest available build_identifier in the class itself. Still a global be at least it is localized to the SlackBackend class.

PR #785, opened 2016-06-10
Python2 removal.

gbin:
A lot of crap all around: try on imports / if PY2/Py3, tests on unicode,
src conversion, travis config, docs ...


.travis.yml
```diff
@@ -11,8 +10,7 @@ before_script:
 
 install:
   - pip install --upgrade pip
-  - pip install -e .
-  - pip install slackclient  # Optional dependency but Slack tests are skipped without it
+  - pip install -e .[slack]
```
zoni: I would leave this the way it was, I think that's a little more clear (and the comment itself is important as well)

zoni:
Aside from the comment above, this looks good in sofar as I can tell.

docs/user_guide/plugin_development/basics.rst
```diff
@@ -140,7 +140,7 @@ Lets go ahead and create ours. Place the following in a file called
     file structure.
 
 Lets look at what this does. We see three sections, `[Core]` ,
-`[Python]` and `[Documentation]`. The `[Core]` section is what tells
+and `[Documentation]`. The `[Core]` section is what tells
```
Ian-Foote:
```
Lets look at what this does. We see two sections, `[Core]` 
and `[Documentation]`. The `[Core]` section is what tells
```

.travis.yml
gbin: I was thinking, maybe another simple test doing `pip install -e .[all]` might be useful to see if the entire dependency tree is still buildable.

PR #787, opened 2016-06-15
New feature: with self.mutable(key) as value:

gbin:
This is a common ugly pattern we make our user do:

```python
d = self['key']
# do somthing with d
self['key'] = d
```

Now they can just do that:

```python
with self.mutable('key') as d:
   # do something on d
```
And d will be saved automatically in the key entry at the end of the
scope.


zoni:
This looks good to me. I do wonder though, should we make this context manager smart enough to include locking so you cannot update the same key with a race condition, or should we leave that up to the end user?

gbin:
implementing locking there might mix 2 problems and might be better handled by the persistence implementation.

PR #793, opened 2016-06-20
Allow flows to be attached to rooms

TooAngel:
Introduce `room_flow` keyword argument on `FlowRoot`. If set to `True`
the flow will be attached to the room. This allows all users in the room
to continue the flow.


gbin:
thx

PR #798, opened 2016-06-25
Allow for disabling slack markdown conversion

harlowja:
When a plugin is being used and that plugin is
only being made to work with slack there is no
need to go through the markdown conversion since
the bot can natively just write the supported
slack format in the first place.

So this allows for that conversion to be turned
off (it is left on by default) via a new RAW_BODY
configuration option.


gbin:
I would prefer to fix our markdown rendering and if something is missing, add a custom extension to it and implement it for the other backends. This option would split the plugins from slack compatible and slack incompatible.
So this is no for this PR.

PR #800, opened 2016-06-26
Implement command shortcuts, by fuzzy matching.

DerFlob:
Allows arbitrary shortcuts of commands, by fuzzy matching them, if no
literal match was found.
The current fuzzy match is a rather naive approach, simply allowing to
omit any character from the command, as long as it stays unambiguous. No
swapping of characters allowed.

### DIFF
--- a/errbot/config-template.py
+++ b/errbot/config-template.py
@@ -371,3 +371,11 @@ REVERSE_CHATROOM_RELAY = {}
 
 # Prevent ErrBot from saying anything if the command is unrecognized.
 # SUPPRESS_CMD_NOT_FOUND = False
+
+# Enable command shortcuts, by fuzzy matching the typed command to the full command names,
+# if no command matches literally.
+# Only an unambiguous match is executed.
+# BOT_COMMAND_SHORTCUTS = True
+
+# Wether or not ErrBot should reply with the expanded command name
+# BOT_COMMAND_SHORTCUTS_EXPANSION_MSG = True
--- a/errbot/core.py
+++ b/errbot/core.py
@@ -270,6 +270,7 @@ class ErrBot(Backend, StoreMixin):
         cmd = None
         command = None
         args = ''
+        ambiguous_shortcut = False
         if not only_check_re_command:
             if len(text_split) > 1:
                 command = (text_split[0] + '_' + text_split[1]).lower()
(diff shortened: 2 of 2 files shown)


gbin:
This is more than just the existing fuzzy match: currently we just ask if the user meant something with a typo, here, it executes command shortcuts. I think it would be better to implement this as a filter plugin.

DerFlob:
Yes, it was supposed to be more than the existing fuzzy matching and notice to the user. I wanted a way easily shortcut existing command. (I'm lazy, when it comes to typing. ;D) Before I often aliased long commands, with a botcmd, that was a shortcut. This is supposed to do it "automatically".

Looking at a filter plugin, I don't think that would work, though? Because filters are only applied, if an actual fully matched command is found? Before I implemented it as an re_botcmd, with a pattern, that always matched. However, this had the disadvantage, that every command was "known" to errbot, and misspelling, that weren't matched by my fuzzy matching, were never processed as an unknown command.

Maybe it might be interesting to look at the return value of _process_command, especially in respect to the blocking of commands by command filters and handling a negative return value of _process_command as an unmatched command?

qlixed:
May I recommend for this the use of [fuzzywuzzy](http://chairnerd.seatgeek.com/fuzzywuzzy-fuzzy-string-matching-in-python/) module? I'm using it to match country names and work great, a quick example:

```
In [1]: from fuzzywuzzy import process

In [2]: command_names = ['test', 'foo', 'bar', 'foobar', 'baz', 'more', 'another', 'one']

In [3]: process.extractBests('ba', command_names, score_cutoff=100)
Out [3]: []

In [6]: process.extractBests('ba', command_names, score_cutoff=90)
Out[6]: [('bar', 90), ('foobar', 90), ('baz', 90)]

In [7]: process.extractBests('bar', command_names, score_cutoff=90)
Out[7]: [('bar', 100), ('foobar', 90)]

In [8]: process.extractBests('bar', command_names, score_cutoff=80)
Out[8]: [('bar', 100), ('foobar', 90)]

In [9]: process.extractBests('bar', command_names, score_cutoff=70)
Out[9]: [('bar', 100), ('foobar', 90)]

In [10]: process.extractBests('bar', command_names, score_cutoff=50)
Out[10]: [('bar', 100), ('foobar', 90), ('baz', 67)]

In [18]: process.extractBests('ba', command_names, score_cutoff=50)
Out[18]: [('bar', 90), ('foobar', 90), ('baz', 90)]

In [19]: process.extractBests('bar', command_names, score_cutoff=50)
Out[19]: [('bar', 100), ('foobar', 90), ('baz', 67)]

In [20]: process.extractOne('bar', command_names, score_cutoff=50)
Out[20]: ('bar', 100)

In [21]: process.extractOne('foo', command_names, score_cutoff=50)
Out[21]: ('foo', 100)
```
You can also pass a dict and you get the output as a tuple of 3 items:
`(dict_item_value, score, dic_item_key)`
For example, searching for 'bird' in the dictionary:   
`{'bard': 'train', 'dog': 'man'}`
may return:    
`[('train', 22, 'bard'), ('man', 0, 'dog')]`

gbin:
It would need maybe to be implemented as a filter plugin or unknown command one

PR #802, opened 2016-06-27
Add slack mrkdwn_in support for cards

harlowja:
Slack api supports telling it that markdown
may be in the title and pretext; if it is not
told this then those do not get rendered correctly.

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -555,11 +555,15 @@ class SlackBackend(ErrBot):
         if isinstance(card.to, RoomOccupant):
             card.to = card.to.room
         to_humanreadable, to_channel_id = self._prepare_message(card)
-        attachment = {}
+        attachment = {
+            'mrkdwn_in': [],
+        }
         if card.summary:
             attachment['pretext'] = card.summary
+            attachment['mrkdwn_in'].append('pretext')
         if card.title:
             attachment['title'] = card.title
+            attachment['mrkdwn_in'].append('title')
         if card.link:
             attachment['title_link'] = card.link
         if card.image:


errbot/backends/slack.py
```diff
@@ -555,11 +555,15 @@ def send_card(self, card: Card):
         if isinstance(card.to, RoomOccupant):
             card.to = card.to.room
         to_humanreadable, to_channel_id = self._prepare_message(card)
-        attachment = {}
+        attachment = {
+            'mrkdwn_in': [],
+        }
         if card.summary:
             attachment['pretext'] = card.summary
+            attachment['mrkdwn_in'].append('pretext')
```
gbin: Then it is slack markdown, no errbot mardown, can we convert it correctly ?

harlowja: Huh, this is telling slack that markdown is in the 'pretext' field; if that's errbot markdown that's fine.

```
By default bot message text will be formatted, but attachments are not. To disable formatting on a non-user message, set the mrkdwn property to false. To enable formatting on attachment fields, set the mrkdwn_in array on each attachment to the list of fields to process.
```

https://api.slack.com/docs/message-formatting

gbin: It is not, what Slack calls "markdown" is something weird. The markdown we use on Errbot is defined by the original markdown syntax: http://daringfireball.net/projects/markdown/syntax + some extra from https://pythonhosted.org/Markdown/extensions/extra.html like table and colors.
Whatever the backends defines (ansi like console, IRC colors, xhtml for xmpp, other gitter markdown flavor etc...) we need to convert the errbot format to it do plugins can write in a consistent markdown. So accepting here the weird slack format straight from the plugins will brake ... hipchat for one.

harlowja:
Gotcha, I think this is still needed anyway, even after converting from the errbot markdown to the slack one, otherwise slack won't even render the converted errbot markdown in the first place. But sure I'll try to get around to doing that conversion process (since it looks like the slack conversion process has issues in general anyway).

gbin:
Indeed we really need to revisit https://github.com/errbotio/errbot/blob/master/errbot/rendering/slack.py ! 
Thanks !

PR #804, opened 2016-07-06
Accept multiple admins in BOT_ADMINS string

acaire:
Following on from #461, allow multiple admins to be specified
when converted to a tuple, rather than just one

### DIFF
--- a/errbot/botplugin.py
+++ b/errbot/botplugin.py
@@ -108,7 +108,7 @@ class BotPluginBase(StoreMixin):
         # if BOT_ADMINS is just an unique string make it a tuple for backwards
         # compatibility
         if isinstance(self._bot.bot_config.BOT_ADMINS, str):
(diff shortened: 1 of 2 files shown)


zoni:
Hi @acaire, thanks for the PR. Unfortunately, I don't think this is desirable behavior.

#461 solves a common point of confusion because `("admin")` is interpreted as `"admin"` by Python, so for a single-entry list, one must specify `("admin",)` (note the trailing slash). That's a very subtle thing so easily overlooked, hence this special-casing.

Your change would change the whole semantics however, by automatically splitting on commas within strings. I'm of the opinion that we should stick to the current paradigm of requiring the user to pass in lists/tuples for distinct entries, rather than resorting to "magic" auto-splitting.

This reduces the amount of config-parsing code, provides "one unambiguous way" to do it and keeps it consistent with other config options that take multiple values.

PR #809, opened 2016-07-19
Preliminary refactoring to introduce dependencies.

gbin:
This is just some preliminary work to introduce a dependency between plugins more easily.

### DIFF
--- a/errbot/plugin_manager.py
+++ b/errbot/plugin_manager.py
@@ -6,6 +6,7 @@ import logging
 import sys
 import os
 import pip
+from yapsy import PluginInfo
 
 from errbot.flow import BotFlow
 from .botplugin import BotPlugin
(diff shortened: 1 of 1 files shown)


errbot/plugin_manager.py
```diff
@@ -465,11 +465,10 @@ def update_dynamic_plugins(self):
 
     def activate_non_started_plugins(self):
         """
-
+        Activates all plugins that are not activated, respecting its dependencies.
         :return: Empty string if no problem occured or a string explaining what went wrong.
```
zoni: If you add a blank line in between the description and `:return:`, Sphinx will render this properly (it treats it as a single line otherwise and won't see `:return:` as a special tag then)

errbot/plugin_manager.py
```diff
@@ -519,13 +518,16 @@ def activate_plugin(self, name):
                 raise PluginActivationException("Plugin already in active list")
             if name not in self.get_all_plugin_names():
                 raise PluginActivationException("I don't know this %s plugin" % name)
-            self.activate_plugin_with_version_check(name, self.get_plugin_configuration(name))
+            plugin_info = self.get_plugin_by_name(name)
+            if plugin_info is None:
+                raise PluginActivationException("get_plugin_by_name_found (should not happen) for %s." % name)
```
zoni: Shouldn't this say ***not*** found?

zoni:
Code wise looks good to me.

PR #813, opened 2016-07-20
Uses plugin name instead of class name everywhere

gbin:
It is recorded beside the doc at `cls.__errname__` so it can be reused.

### DIFF
--- a/errbot/botplugin.py
+++ b/errbot/botplugin.py
@@ -157,10 +157,18 @@ class BotPluginBase(StoreMixin):
         """
         return self._bot.bot_identifier
 
+    @property
+    def name(self) -> str:
+        """
+        Get the plugin name as defined in its .plug file.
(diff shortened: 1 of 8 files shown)


zoni:
Would this change the names of existing plugins? (I'm asking with an eye towards ACLs. If people upgrade and names change, their ACLs might no longer match, so we should make it a major version bump if we do that)

gbin:
the check in the ACL plugin is still : `plugin=type(f.__self__).__name__` which is the class name but it raises a more global question, the name of a plugin is very ambiguous right now between the name in the .plug and the name of the class.

Some plugins are effectively "classless" with the dynamic plugins and the .plug name is the public name of it so in the long run we might want to favor it.

gbin:
this has been superceeded by #1001, with a different implementation

PR #818, opened 2016-07-25
Make some improvements to Slack mention callbacks

acaire:
* Prevent chomping of colons following mentions
* Use regex to find mentions that are surrounded
  by non-alphanumeric characters
* Fix a bug causing all mentions to be returned
  as a duplicate of the first mention
* Move mention processing to separate function
  and write some tests

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -417,21 +417,7 @@ class SlackBackend(ErrBot):
             text = event['text']
             user = event.get('user', event.get('bot_id'))
 
-        mentioned = []
-
-        for word in text.split():
-            if word.startswith('<') or word.startswith('@'):
-                try:
-                    identifier = self.build_identifier(word.replace(':', ''))
-                except Exception as e:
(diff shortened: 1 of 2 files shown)


errbot/backends/slack.py
```diff
@@ -793,6 +779,32 @@ def sanitize_uris(text):
...
+        """
+        mentioned = []
+
+        m = re.findall('<@[^>]*>*', text)
+
+        for word in m:
+            try:
+                identifier = self.build_identifier(word)
+            except Exception as e:
+                log.debug("Tried to build an identifier from '%s' but got exception: %s", word, e)
```
gbin: Maybe makes that a warn otherwise people might miss it.

gbin: Maybe not, the original code was not so I guess it was not a problem.

gbin:
Awesome, thanks !

acaire:
Appreciate it, thanks! :)

PR #820, opened 2016-07-28
Fix extracting Slack identifier w/piped username

acaire:
The channel leave event (and possibly others), return the user
id piped with the username, i.e.:

```
{
    "type": "message",
    "subtype": "channel_leave",
    "ts": "1358877455.000010",
    "user": "U2147483828",
    "text": "<@U2147483828|cal> has left the channel"
}
```

This commit updates extract_identifiers_from_string to allow this


zoni:
Thanks!

Issue #821, opened 2016-07-28
System Encoding Warning


stevejarvis:
The [CLI at line 31](https://github.com/errbotio/errbot/blob/master/errbot/cli.py#L31) compares the encoding from locale with encoding from sys. This leads to silly error message like:

    # errbot --init
    22:47:54 WARNING  errbot.cli                Starting errbot with a default system encoding other than 'utf-8' might cause you a heap of troubles. Your current encoding is set at 'utf-8'

It's not clear why these would give different answers, but the [Python docs](https://docs.python.org/3.3/library/sys.html?highlight=sys#sys.getdefaultencoding) suggest sys.getdefaultencoding() always returns a Unicode encoding. I dunno, locale seems to be the source of truth though, because Errbot certainly does work better once that reads 'utf-8' as well.

    Python 3.5.2 (default, Jul  5 2016, 12:43:10) 
    [GCC 5.4.0 20160609] on linux
    Type "help", "copyright", "credits" or "license" for more information.
    >>> import locale
    >>> import sys
    >>> locale.getpreferredencoding()
    'ANSI_X3.4-1968'
    >>> sys.getdefaultencoding()
    'utf-8'

PR #822, opened 2016-07-28
Clarify warning about system encoding.

stevejarvis:
Sys and locale can have different ideas about what encoding will be
used, and checking one but echoing the other can be weird.

Locale module seems like the source of truth in practice.

Fixes issue #821


zoni:
Good catch, thanks!

PR #825, opened 2016-07-30
Add tox

medecau:
This pull-request adds a tox config file with environments for python 3.3, 3.4, and 3.5.
It also includes environments for codestyle conformance validation and pypi linting.

If you're default python environment is already python 3 all you need to run the tests is tox. Alternatively you can run detox for parallelised tests.

Learn more about tox here: https://tox.readthedocs.io/en/latest/


gbin:
This is actually better than our homebrewed script, we need to bind to travis-ci too... do we need to use that ? https://pypi.python.org/pypi/tox-travis

medecau:
Although tests are green I believe this needs reviewing by at least two devs to makes sure the tests are proper.

gbin:
Looks good ! thanks so much !

PR #829, opened 2016-08-01
Fixing room references from the Slack mobile client.

jasedit:
This fixes room references from the Slack mobile client, which sends room references in the form <#C12345|channelname>.


zoni:
#820 dealt with this same issue and has already been merged. The approach in that PR is similar, but subtly different. I'm going to close this one but would be very interested in hearing if there's anything this PR addresses which #820 did not.

PR #831, opened 2016-08-11
Change docs to use !plugin activate/deactivate

danielnoom:
Now load and unload are shown in docs, but these commands have been replaced.


gbin:
Thanks for spotting that !

PR #832, opened 2016-08-13
Fix bug in user guide

aschokking:
In the send_card example the choice of re-using send_card for the bot method name meant self.send_card was calling that instead of the actual parent implementation. Probably a better idea to use a different command method name?


gbin:
thx

PR #836, opened 2016-08-16
Use pip from subprocess instead of package

### LINKED ISSUE #835: Pip should not be invoked from pip.main(), and is not thread safe
This is linked to the issue I opened at pypa/pip#3889

It was stated that pip has no programming interfaces, and so pip.main() should be avoided at:
https://github.com/errbotio/errbot/blob/master/errbot/plugin_manager.py#L58

The only supported way of calling pip from within another program is via subprocess.Popen().

As pip is not thread safe, invoking pip.main() from within a thread may result in corrupted packages. Also, it will not work in the case of the plugin_manager because of pypa/pip#3889 breaking the pip install in threading.

This issue is preventing the bot to install packages when used in threading mode, which break the usage of threading mode in my case.

rawouter:
This PR fixes #835. It has been tested in python3 venv with `AUTOINSTALL_DEPS = True` while installing a plugin containing a `requierements.txt`:
- with empty venv and fresh plugin installation and `BOT_ASYNC = True`
- with empty venv and fresh plugin installation and `BOT_ASYNC = False`
- with empty venv and already installed plugin and `BOT_ASYNC = True` (this was broken)
- with empty venv and already installed plugin and `BOT_ASYNC = False` 
- with venv containing all the requierements

The last two cases are typical Docker installation usage, the data directory might be persistant but the virtual env might not, causing the bot to fail the plugin at re-install with BOT_ASYNC is True.

It also has an additional commit (66c70b9) fixing a bug of a badly imported module at L63 throwing more cosmetic errors (the newly packages were loaded by the plugin anyway?).

### DIFF
--- a/errbot/plugin_manager.py
+++ b/errbot/plugin_manager.py
@@ -1,11 +1,11 @@
 """ Logic related to plugin loading and lifecycle """
-import traceback
 from configparser import NoSectionError, NoOptionError, ConfigParser
 from importlib import machinery, import_module
 import logging
-import sys
 import os
-import pip
+import subprocess
+import sys
+import traceback
 from yapsy import PluginInfo
 
 from errbot.flow import BotFlow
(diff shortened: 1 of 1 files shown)


zoni:
Thanks for this!

PR #837, opened 2016-08-21
Adding support for stream requests to the Slack backend.

jasedit:
This allows files to be uploaded to Slack using the stream interface. Slack's API for file uploads appears to be a blocking interface, so the Stream returned from the `send_stream_request()` function is marked successful or errored upon return.

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -585,6 +585,8 @@ class SlackBackend(ErrBot):
     def send_stream_request(self, identifier, fsource, name='file', size=None, stream_type=None):
         """Starts a file transfer. For Slack, the size and stream_type are unsupported"""
         stream = Stream(identifier, fsource, name, size, stream_type)
+        log.debug("Requesting upload of {0} to {1} (size hint: {2}, stream type: {3})".format(name,
+                  identifier.channelname, size, stream_type))
         self.thread_pool.putRequest(WorkRequest(self._slack_upload, args=(stream,)))
         return stream
 
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -565,9 +565,8 @@ class SlackBackend(ErrBot):
                 "to %s: %s" % (to_humanreadable, mess.body)
             )
 
-    def send_stream_request(self, identifier, fsource, name=None, size=None, stream_type=None):
+    def send_stream_request(self, identifier, fsource, name='file', size=None, stream_type=None):
         """Starts a file transfer. For Slack, the size and stream_type are unsupported"""
-        name = name if name else 'file'
         stream = Stream(identifier, fsource, name, size, stream_type)
         log.debug('Initiating upload of {0}'.format(name))
         resp = self.api_call('files.upload', data={
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -13,7 +13,7 @@ from markdown.extensions.extra import ExtraExtension
 from markdown.preprocessors import Preprocessor
 
 from errbot.backends.base import Message, Presence, ONLINE, AWAY, Room, RoomError, RoomDoesNotExistError, \
-    UserDoesNotExistError, RoomOccupant, Person, Card
+    UserDoesNotExistError, RoomOccupant, Person, Card, Stream
 from errbot.core import ErrBot
 from errbot.utils import split_string_after
 from errbot.rendering.ansiext import AnsiExtension, enable_format, IMTEXT_CHRS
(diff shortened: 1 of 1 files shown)


errbot/backends/slack.py
```diff
@@ -565,6 +565,24 @@ def send_message(self, mess):
                 "to %s: %s" % (to_humanreadable, mess.body)
             )
 
+    def send_stream_request(self, identifier, fsource, name=None, size=None, stream_type=None):
+        """Starts a file transfer. For Slack, the size and stream_type are unsupported"""
+        name = name if name else 'file'
```
gbin: maybe simply name='file' in the signature ?

errbot/backends/slack.py
```diff
@@ -565,6 +565,24 @@ def send_message(self, mess):
...
+        stream = Stream(identifier, fsource, name, size, stream_type)
+        log.debug('Initiating upload of {0}'.format(name))
+        resp = self.api_call('files.upload', data={
```
gbin: It looks totally synchronous, maybe enqueue a WorkRequest on the internal threadpool ?
Something along that (untested code):

```python
def upload(stream):
     stream.accept()
     resp = self.api_call('files.upload', data={'channels': stream.identifier.channelid, 'filename': stream.name, 'file': stream)
      log.debug('Upload response: {0}'.format(resp))
      if "ok" in resp and resp["ok"]:
        stream.success()
      else:
        stream.error()

and in send_stream_request:
[...]
self.thread_pool.putRequest(WorkRequest(self.upload, [stream, ]))
return stream

```

errbot/backends/slack.py
jasedit: Good catch - done!

errbot/backends/slack.py
jasedit: I'm currently testing this, and it looks like the WorkRequest never gets serviced. There's no error in the log, just an enqueued WorkRequest and then silence. I'm really not familiar with this part of errbot - is there any documentation or guidance in how this subsystem works?

jasedit:
@gbin As far as I can tell, the `thread_pool` system does not work at all. I've got a branch with my experimental version [here](https://github.com/jasedit/errbot/tree/test_thread_pool), which enqueues a `WorkRequest` and immediately waits on it. According to the thread_pool [documentation](https://chrisarndt.de/projects/threadpool/) I can find, this should wait until the `WorkRequest` is completed, and return. In this configuration, errbot hangs indefinitely on an attempted upload. Is there any working example of the thread_pool being used in errbot?

gbin:
I think the asynchronism is put at the wrong place. It should return the stream right away and in the background do the api call.
http://errbot.io/en/latest/user_guide/plugin_development/streams.html#sending-a-file-to-a-user-or-a-room

The stream object should gives the user the feedback about the state of the transfer.

jasedit:
Yes - the point of that branch is to show that the `thread_pool` completely locks up, and never processes the `WorkRequest` as I would expect. Removing the `wait` call results in the bot never uploading the file, but being able to resume operation. Once I can debug why that happens, the actual version placed in this PR would not wait on the `thread_pool`, to allow asynchronous behavior.

jasedit:
After beating my head against this code for a rather long time, I finally discovered the underlying issue which was causing trouble. It appears that any usage of log statements inside a `WorkRequest` silently hangs the `WorkRequest`. Removing all logging statements from the WorkRequest allows the upload to succeed.

As per earlier discussion, now that I've fixed the hanging, this PR now includes the asynchronous behavior without calling `wait()`.

errbot/backends/slack.py
```diff
@@ -565,6 +566,25 @@ def send_message(self, mess):
                 "to %s: %s" % (to_humanreadable, mess.body)
             )
 
+    def _slack_upload(self, stream):
+        """Perform upload defined in a stream."""
+        stream.accept()
```
gbin: I have almost the same code in my branch, with logs working with no problem. The only thing is that failures are silent so you need protect the whole method with a try / catch Exception.

errbot/backends/slack.py
```diff
@@ -565,6 +566,25 @@ def send_message(self, mess):
...
+            'file': stream
+            })
+        if "ok" in resp and resp["ok"]:
+            stream.success()
+        else:
+            stream.error()
+
+    def send_stream_request(self, identifier, fsource, name='file', size=None, stream_type=None):
+        """Starts a file transfer. For Slack, the size and stream_type are unsupported"""
+        stream = Stream(identifier, fsource, name, size, stream_type)
```
gbin: add at least some log debug with the parameters

errbot/backends/slack.py
```diff
@@ -568,20 +568,25 @@ def send_message(self, mess):
...
+                'channels': stream.identifier.channelid,
+                'filename': stream.name,
+                'file': stream
+                })
+            if "ok" in resp and resp["ok"]:
+                stream.success()
+            else:
+                stream.error()
+        except:
+            log.error("Upload of {0} to {1} failed.".format(stream.name, stream.identifier.channelname))
```
gbin: log.exception

errbot/backends/slack.py
jasedit: The final step in fixing it for me was removing the debug statements inside the upload statement. Having them in before and after the `api_call` caused the upload to fail silently, removing them made the upload succeed.

I've wrapped the method in a try/except block to print out a failure message on upload, as requested.

errbot/backends/slack.py
jasedit: Didn't know there was an exception level - changed as requested (and learned something new!)

jasedit: Weird - this fix is part of the PR, but it's still reporting the change as outstanding? Or not updating the display to show the fix?

sheluchin: @jasedit Yep, that's a bit weird. Maybe it requires a fresh commit instead of amending and force pushing?

gbin:
Thanks

jasedit:
Thanks for the review and insights

PR #838, opened 2016-08-23
Adds a setting for the internal threadpool.

gbin:


zoni:
Hah. I'd been meaning to do that for so very long... :smile:

PR #842, opened 2016-09-01
Add a page about exception handling

sheluchin:
I didn't see anything about exceptions in the `Plugin Development` docs, and thought it might be helpful to include a few blurbs. 

I'd be happy to make any changes prior to merging, if you guys think any improvements could be made.

I did notice that the `Next Page` link in the sidebar TOC was not properly updated, but couldn't quite figure out what the problem was. Could anyone point out what I'm doing incorrectly there?


gbin:
It is very clear ! Thanks for the contribution.

PR #846, opened 2016-09-05
Install plugin requirements.txt as a bulk.

gbin:
Not only it simplifies the code but this gives us the huge benefit of
having totally standard requirements.txt.
(before that, we did not have support for version pinning etc.)

### DIFF
--- a/errbot/plugin_manager.py
+++ b/errbot/plugin_manager.py
@@ -58,10 +58,10 @@ def install_packages(req_path):
     try:
         if hasattr(sys, 'real_prefix') or (hasattr(sys, 'base_prefix') and (sys.base_prefix != sys.prefix)):
             # this is a virtualenv, so we can use it directly
-            p = subprocess.check_call(['pip', 'install', '--requirement', req_path])
+            subprocess.check_call(['pip', 'install', '--requirement', req_path])
         else:
             # otherwise only install it as a user package
-            p = subprocess.check_call(['pip', 'install', '--user', '--requirement', req_path])
+            subprocess.check_call(['pip', 'install', '--user', '--requirement', req_path])
     except:
         log.exception('Failed to execute pip install for %s.', req_path)
         return sys.exc_info()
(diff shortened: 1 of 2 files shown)


errbot/plugin_manager.py
```diff
@@ -49,35 +49,33 @@ def populate_doc(plugin):
...
+    log.info("Installing packages from '%s'." % req_path)
     try:
         if hasattr(sys, 'real_prefix') or (hasattr(sys, 'base_prefix') and (sys.base_prefix != sys.prefix)):
             # this is a virtualenv, so we can use it directly
-            p = subprocess.Popen(['pip', 'install', package])
+            p = subprocess.check_call(['pip', 'install', '-r', req_path])
         else:
             # otherwise only install it as a user package
-            p = subprocess.Popen(['pip', 'install', '--user', package])
+            p = subprocess.check_call(['pip', 'install', '--user', '-r', req_path])
```
zoni: It's not a good practice to use short argument flags in scripts, as they're less readable and more likely to change than their long form counterparts. Suggest using `--requirement` instead of `-r`.

errbot/plugin_manager.py
```diff
@@ -49,35 +49,33 @@ def populate_doc(plugin):
...
+
+        Return an exc_info if it fails otherwise None.
     """
-    log.info("Installing package '%s'." % package)
+    log.info("Installing packages from '%s'." % req_path)
     try:
         if hasattr(sys, 'real_prefix') or (hasattr(sys, 'base_prefix') and (sys.base_prefix != sys.prefix)):
             # this is a virtualenv, so we can use it directly
-            p = subprocess.Popen(['pip', 'install', package])
+            p = subprocess.check_call(['pip', 'install', '-r', req_path])
```
zoni: It's not a good practice to use short argument flags in scripts, as they're less readable and more likely to change than their long form counterparts. Suggest using `--requirement` instead of `-r`.

zoni: Also `p` is not referenced anymore now, so suggest removing the assignment altogether.

errbot/plugin_manager.py
zoni: Also `p` is not referenced anymore now, so suggest removing the assignment altogether.

zoni:
Think it's good now, will wait for Travis to approve. I haven't actually tested it locally though. :smile:

PR #848, opened 2016-09-13
Escape asterisk when displaying name in help output

marksull:
If the name of the command includes an asterisk (for example in an re_botcmd regex) ensure the asterisk is escaped to avoid incorrectly formatted markdown

### DIFF
--- a/errbot/core_plugins/help.py
+++ b/errbot/core_plugins/help.py
@@ -155,6 +155,6 @@ class Help(BotPlugin):
             if len(cmd_doc) > 80:
                 cmd_doc = '{doc}...'.format(doc=cmd_doc[:77])
 
-        help_str = '- **{prefix}{name}** - {doc}\n'.format(prefix=prefix, name=name, doc=cmd_doc)
+        help_str = '- **{prefix}{name}** - {doc}\n'.format(prefix=prefix, name=name.replace('*', '\*'), doc=cmd_doc)
 
         return help_str


zoni:
I'm quite unsure whether this is the best approach. Regexes may contain many other characters that are special in markdown, the `[this is a link](http://example.com)` coming to mind especially as `()` and `[]` are used quite frequently in a regex as well.

So if we do take this approach, I think we should have a robust markdown escape function (I could've sworn I had one in one of my own plugins to show as example, but I cannot for the life of me find it anymore).

On the other hand, perhaps it would be even better to display the regex as a fixed-width code block (note: \*\* was replaced with \`):

```python
help_str = '- `{prefix}{name}` - {doc}\n'.format(prefix=prefix, name=name, doc=cmd_doc)
```

What do you think?

marksull:
Certainly the option that you proposed is a valid option. And you are correct that this doesn't solve all potential regex cases, but the help_str is being forcibly wrapped in asterisks which is causing a display issue when an asterisk is in the regex which is the minor problem I am trying to resolve.

If we were looking for a more elegant solution then I don't think enforcing any type of markup on help text (or any text for that matter) is the right direction, as you are assuming every platform uses the same markdown or has implemented markdown to the same level. I have three alternate proposals to solve this:

1) Add to re_botcmd an alt_text parameter (or something of the like) that allows the command to be represented in a more human consumable form in the help output (i.e. in lieu of the regex being displayed, the alt_text is displayed). For example:

```
regex -> my command (\S*) (\S*)

alt_text -> my command [option1|option2|option3] <value>

```

2) Pass the decorating up to the backend bot so that the platform specific implementation can decorate the text in a format suitable to the target platform.

3) Both 1 and 2

zoni:
> Certainly the option that you proposed is a valid option. And you are correct that this doesn't solve all potential regex cases, but the help_str is being forcibly wrapped in asterisks which is causing a display issue when an asterisk is in the regex which is the minor problem I am trying to resolve.

With minor changes I can definitely accept this PR so that the display issue you are running into is fixed. Instead of replacing `*` with `\*` I'd just like to see robust markdown escape taking place as alluded to in my earlier comment.

I had another look and finally [found where that markdown escape function was hiding](https://github.com/errbotio/errbot/blob/076f26fd3a422523d8c4708c37c25e13b53bac4d/errbot/rendering/__init__.py#L87-L91) that I knew we had. :smile: 

So the changes I'd like to see is instead of doing `name=name.replace('*', '\*')` we do `name=md_escape(name)` so that it is fixed for all cases.

> If we were looking for a more elegant solution then I don't think enforcing any type of markup on help text (or any text for that matter) is the right direction, as you are assuming every platform uses the same markdown or has implemented markdown to the same level.

Enforcing markdown on plugin output is actually by design and not going to change. We don't pass markdown through unaltered to the chat network but we render it inside errbot into a format that displays well on the network. For some networks that means we send them markdown too but others may use something different (we render to HTML-IM for XMPP, ASCII with color codes for IRC, etc).

That means plugin authors always write markdown and we go to great lengths to ensure it displays nicely on whatever chat network you run the plugin, without having to do anything for it. :smile: 

I do however think displaying the trigger regex in help output for regex-based commands is not always ideal. Some type of alternate, more helpful output may be desirable there yes. I'd love to discuss potential solutions to that in a separate issue.

gbin:
No activity on this one please re-open if needed.

PR #850, opened 2016-09-14
Update configuration.rst

darkorb:
Correcting `!config PluginExample` to `!plugin config PluginExample`


zoni:
Good catch, thanks!

PR #852, opened 2016-09-18
Add --dark-text CLI option for light backgrounds

sheluchin:


errbot/logs.py
```diff
@@ -20,11 +21,16 @@ def ispydevd():
 isatty = pydev or stream.isatty()  # force isatty if we are under pydev because it supports coloring anyway.
 console_hdlr = logging.StreamHandler(stream)
 
+parser = argparse.ArgumentParser()
+parser.add_argument('-D', '--dark-text', dest="dark_text", action='store_true',
+                    help='force the log output to be dark, for use with light backgrounds')
```
zoni: I don't like how this introduces argument parsing into a module that isn't directly responsible for CLI interaction. I don't directly have a better suggestion, but I think we should find some other approach to pass on this information.

errbot/cli.py
```diff
@@ -131,6 +131,8 @@ def main():
                                 help='force local text backend')
     mode_selection.add_argument('-G', '--graphic', dest="backend", action='store_const', const="Graphic",
                                 help='force local graphical backend')
+    mode_selection.add_argument('-D', '--dark-text', dest="dark_text", action='store_true',
+                                help='force the log output to be dark, for use with light backgrounds')
```
zoni: I think 9 out of 10 people won't use this option, so while I have no objections against `--dark-text` at all, I'd prefer to keep `-D` free, at least for the time being.

zoni:
Perhaps instead of doing this via argument flags, it would be an idea to base this behaviour on an environment variable? Something along the lines of `ERRBOT_LOGGER_THEME` where valid values could be `light`, `dark` and `nocolor`? (this would have the advantage of letting people disable the coloring if they wanted).

sheluchin:
@zoni Would it be practical to save a value to the config inside cli.py and
then retrieve it in logs.py? I looked around for a way to do it, but no
easy approach jumped out at me.

I like the environment variable idea, but would that be introducing a new
convention to Errbot? I'm not aware of any other options that are
configured with environment variables in Errbot.

gbin:
I would not recommend another way to configure errbot beside config.py and flags, this gonna confuse our users.
I would say: one conf parameter can be also used by the Text backend to change its theme btw.
`TEXT_COLOR_THEME` with `light`, `dark` and `nocolor` as value to start with seems perfect to me.

If you want to map it to an env variable, you can always express with something like:
TEXT_COLOR_THEME = os.getenv('ERRBOT_TEXT_COLOR_THEME', 'nocolor')

zoni:
@sheluchin:
> I like the environment variable idea, but would that be introducing a new convention to Errbot? I'm not aware of any other options that are configured with environment variables in Errbot.

@gbin:
> I would not recommend another way to configure errbot beside config.py and flags, this gonna confuse our users.

Fair point, indeed, lets avoid that. I'm in favor of @gbin's suggestion: *"`TEXT_COLOR_THEME` with `light`, `dark` and `nocolor` as value to start with"*.

gbin:
keep it TEXT_COLOR_THEME, I'll take care of the Text backend switch in another PR. (if your console color suck for logs, it will suck for the text backend...)

gbin:
Any progress on this one ?

sheluchin:
Sorry, have been very busy at work the last little while. Would you prefer I close it and re-open when I have more to add?

gbin:
you can continue on this one, the discussion around it is interesting.

sheluchin:
@zoni @gbin I've made some changes according to our discussion. Please let me know if you would like any other modifications.

PR #856, opened 2016-09-29
Allow configurable SSL verification for Hipchat Backend

sivy:
Support for self-signed Hipchat server connections:

When using an alternate/internal Hipchat server, there's a good chance that if it's using SSL the certificate will be self-signed. This change adds a `verify` config that passes through to the hypchat client so that it can connect to servers with self-signed (non-verifiable) SSL certificates.

`verify` defaults to `True` to preserve current behavior.

Tested with Hipchat Server (could not find the version but it's a version or so behind)


gbin:
Awesome, it is possible to add the flag to the doc too ? https://github.com/errbotio/errbot/blob/master/docs/user_guide/configuration/hipchat.rst

Thanks !

sivy:
Added docs for `verify` flag. 😀

gbin:
Thx

PR #862, opened 2016-10-03
Fix testing Telegram user ID against an ACL rule

### LINKED ISSUE #861: Telegram clients cannot be checked against an ACL
Telegram IDs are integer numbers, which means they fail to be used by `fnmatch` directly.

```python
  File "/Users/pavel/.virtualenvs/mybot/lib/python3.5/site-packages/errbot/core.py", line 344, in _process_command_filters
    msg, cmd, args = cmd_filter(msg, cmd, args, dry_run)
  File "/Users/pavel/.virtualenvs/mybot/lib/python3.5/site-packages/errbot/core_plugins/acls.py", line 76, in acls
    if 'allowusers' in acl and not glob(usr, acl['allowusers']):
  File "/Users/pavel/.virtualenvs/mybot/lib/python3.5/site-packages/errbot/core_plugins/acls.py", line 22, in glob
    return any(fnmatch.fnmatchcase(text, pattern) for pattern in patterns)
  File "/Users/pavel/.virtualenvs/mybot/lib/python3.5/site-packages/errbot/core_plugins/acls.py", line 22, in <genexpr>
    return any(fnmatch.fnmatchcase(text, pattern) for pattern in patterns)
  File "/Users/pavel/.pyenv/versions/3.5.1/lib/python3.5/fnmatch.py", line 71, in fnmatchcase
    return match(name) is not None
TypeError: expected string or bytes-like object
```

asfaltboy:
fix #861

### DIFF
--- a/errbot/backends/test.py
+++ b/errbot/backends/test.py
@@ -72,8 +72,8 @@ class TestPerson(Person):
 
     def __unicode__(self):
         if self.client:
-            return self._person + "/" + self._client
-        return self._person
+            return '{}/{}'.format(self._person, self._client)
+        return '{}'.format(self._person)
 
     __str__ = __unicode__
 
--- a/errbot/core_plugins/acls.py
+++ b/errbot/core_plugins/acls.py
@@ -19,7 +19,9 @@ def glob(text, patterns):
     """
     if isinstance(patterns, str):
         patterns = (patterns,)
-    return any(fnmatch.fnmatchcase(text, pattern) for pattern in patterns)
+    if not isinstance(text, str):
+        text = str(text)
+    return any(fnmatch.fnmatchcase(text, str(pattern)) for pattern in patterns)
 
 
 def ciglob(text, patterns):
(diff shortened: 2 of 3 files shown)


asfaltboy:
I've pushed a test for the functionality. It seems that DEBUG level logging would make errbot crash on `errbot/core.py(225)`:

```python
>       log.debug("*** frm = %s" % frm)
E       TypeError: __str__ returned non-string (type int)
```

Since [`__str__` must return a string][1]; I think the issue is with `TestPerson` which does not explicitly return a string. I found [this thread][2] to be a good summary of why `__str__` should return something meaningful.

[1]: https://docs.python.org/3/reference/datamodel.html#object.__str__
[2]: https://mail.python.org/pipermail/tutor/2013-March/094205.html

zoni:
> I've added a test for the functionality. But, it seems that DEBUG level logging would make errbot crash on errbot/core.py(225):

Just to be clear, does it still crash with the changes made in e7421df00443b2e7b97b77cc592b6232cb639e8b?

Regardless, I'm going to go ahead and merge this in to fix the original issue. Thanks!

asfaltboy:
@zoni nope, it does not crash, it output the debug row as a string. Thanks!

zoni:
Awesome!

Issue #726, opened 2016-04-26
init.d script


cweiske:
It would be nice if errbot would ship with a init.d script, so that I can let it run automatically on system startup.

I did read http://errbot.io/en/latest/user_guide/setup.html#starting-the-daemon

> If you are going to run your bot all the time then using some process control system such as supervisor is highly recommended. Installing and configuring such a system is outside the scope of this document however.

It would still be nice to have that pre-configured.

Djiit:
As stated in the docs, this is a bit out of scope but we could provide a minimal supervisor configuration like this: 

```ini
# /etc/supervisor/conf.d/errbot.conf

[program:errbot]
# The root dir of your errbot installation (with a valid config.py)
directory=/path/to/your/errbot/dir
# If your are using a virtualenv, make sure to use the correct errbot
command=errbot
# Start Errbot at Supervisor startup.
autostart=true
# Restart Errbot if it crashes.
autorestart=true
```

rroemhild:
You can try my init.d script https://gist.github.com/rroemhild/2e4d7ee5061f56f3300843c6ee8e67b5 I use on Ubuntu 12.04. Copy this script to `/etc/init.d/errbot` and edit the variables. Finally `chmod ug+x /etc/init.d/errbot` and run `service errbot start`.

rroemhild:
And another example for [systemd](https://gist.github.com/rroemhild/6c2c70b659741ab2ba9ba6f5f8be9f46), tested on debian jessie.

Save this file to /etc/systemd/system/errbot.service, reload systemd `systemctl daemon-reload` and start errbot `systemctl start errbot`. Default config is `/etc/errbot/config.py`.

To use another config file modify `Environment="CONFIGFILE=/etc/errbot/config.py"`.
To run errbot from an virtualenv modify `ExecStart`.

Or overwrite the parts in `/etc/systemd/system/errbot.service.d/errbot.conf`:

```
[Service]
Environment="CONFIGFILE=/home/errbot/config.py"
ExecStart=
ExecStart=/home/errbot/venv/bin/errbot --daemon --config $CONFIGFILE
```

At least enable errbot as a service `systemctl enable errbot`.

cweiske:
Thanks for the systemd example - after my upgrade to debian8 this was the easiest to get going.

I added the errbot user at first:
```
$ useradd --no-create-home --no-user-group -g nogroup -s /bin/false errbot
```
and gave it ownership to the log file and the data directory.

[...]

PR #863, opened 2016-10-04
docs: Add example init daemon configs (#726)

sijis:
An idea to add the example daemon init script to the docs.

Examples are trimmed down from issue #726.

Alternatively, I could add this to a `contrib/` directory in the codebase. I thought, having it in the docs was more visible for new users.

### DIFF
--- a/docs/code_examples/supervisord.conf
+++ b/docs/code_examples/supervisord.conf
@@ -1,0 +1,5 @@
+[program:errbot]
+directory=/etc/errbot
+command=/path/to/my/virtualenv/bin/errbot
+autostart=true
+autorestart=true
--- a/docs/code_examples/supervisord.conf
+++ b/docs/code_examples/supervisord.conf
@@ -1,5 +1,10 @@
 [program:errbot]
-directory=/etc/errbot
-command=/path/to/my/virtualenv/bin/errbot
-autostart=true
-autorestart=true
+command = /path/to/errbot/virtualenv/bin/errbot --config /path/to/errbot/config.py
+user = errbot
+stdout_logfile = /var/log/supervisor/errbot.log
+stderr_logfile = NONE
+redirect_stderr = true
+directory = /path/to/errbot/
+startsecs = 3
+stopsignal = INT
+environment = LC_ALL="en_US.UTF-8"
--- a/docs/code_examples/systemd.service
+++ b/docs/code_examples/systemd.service
@@ -3,4 +3,8 @@ Description=Start Errbot chatbot
 After=network.service
 
 [Service]
-ExecStart=/path/to/my/virtualenv/bin/errbot --daemon --config /etc/errbot/config.py
+Type=forking
+Environment="LC_ALL=en_US.UTF-8"
+ExecStart=/path/to/errbot/virtualenv/bin/errbot --daemon --config /path/to/errbot/config.py
+WorkingDirectory=/path/to/errbot/
+User=errbot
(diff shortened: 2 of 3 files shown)


asfaltboy:
Thought I'd add a few comments, as I just had to config errbot lately.

- I prefer @zoni 's [supervisor example gist][1], since it uses the INT stop signal and is a bit more real life example.
- Not sure about this, but systemd unit should probably not start errbot with `--daemon` option; at least not in `simple` mode (maybe with `Type=forking`). I suggest simply removing the option as systemd will take care of the daemonizing/forking itself.

[1]: https://gist.github.com/zoni/9f5bb3043c91ab8dafd6a049ee1ce47d

sijis:
@asfaltboy Those are all good points. I'll go ahead and update with @zoni's example.
I'll poke around systemd config. I can't say i really tried it as i based it solely on the original issue.

docs/code_examples/supervisord.conf
```diff
@@ -0,0 +1,5 @@
+[program:errbot]
+directory=/etc/errbot
+command=/path/to/my/virtualenv/bin/errbot
+autostart=true
+autorestart=true
```
zoni: As mentioned, I'd use [my version](https://gist.github.com/zoni/9f5bb3043c91ab8dafd6a049ee1ce47d) instead. Most notably it makes sure supervisor sends the correct signal and ensures UTF-8 is used as locale.

docs/code_examples/systemd.service
```diff
@@ -0,0 +1,6 @@
+[Unit]
+Description=Start Errbot chatbot
+After=network.service
+
+[Service]
+ExecStart=/path/to/my/virtualenv/bin/errbot --daemon --config /etc/errbot/config.py
```
zoni: It's not necessary to let errbot daemonize itself. In fact, it is better not to (and won't work like this without also specifying `type=forking`).

It would also be good to incorporate some of the other items present in the supervisor config, namely setting the correct working dir, stop signal and locale.

docs/code_examples/supervisord.conf
sijis: Synced the changes.

docs/code_examples/systemd.service
sijis: I tried to make it similar to supervisord configuration

docs/code_examples/systemd.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Start Errbot chatbot
+After=network.service
+
+[Service]
+Type=forking
+Environment="LC_ALL=en_US.UTF-8"
+ExecStart=/path/to/errbot/virtualenv/bin/errbot --daemon --config /path/to/errbot/config.py
```
zoni: It's better not to daemonize and run in the foreground instead, letting systemd handle all that stuff transparently. That means `--daemon` can be removed here and `Type=forking` is no longer needed.

docs/code_examples/systemd.service
```diff
@@ -0,0 +1,10 @@
...
+WorkingDirectory=/path/to/errbot/
+User=errbot
```
zoni: We should add `Restart=on-failure` (or perhaps even `always`) to ensure systemd restarts the bot should it crash for any reason.

zoni: We should also add `KillSignal=SIGINT` because systemd defaults to SIGTERM.

sijis:
@zoni I've addressed your comments/suggestions.

zoni:
I might change the docs a bit later (moving this part to it's own section with just a simple reference to it) to make the install docs seem less intimidating, but this is excellent, thanks.

PR #868, opened 2016-10-10
Allow admins to mute notifications

sijis:
This allows certain admins to not receive any of the bot admin messages.

Use cases:
- plugin failures on bot startup are "annoying" and some admins want them silenced.
- Too many admin messages being sent from various plugin/bot errors and can get overwhelming
- Admin may want to "mute" all messages when on vacation, leave, holiday, etc

I'm obviously open to feedback / suggestions on this.


sijis:
I should add test cases for this ...

sijis:
Finally added test cases!

This should be good for review now.

errbot/bootstrap.py
```diff
@@ -63,6 +63,8 @@ def bot_config_defaults(config):
         config.TEXT_DEMO_MODE = True
     if not hasattr(config, 'BOT_ADMINS'):
         raise ValueError('BOT_ADMINS missing from config.py.')
+    if not hasattr(config, 'BOT_ADMINS_MUTE_NOTIFICATIONS'):
```
gbin: that notifies everyone ?

gbin:
Thanks for the contrib !

errbot/bootstrap.py
sijis: I see your point on the positive flag.
It makes more sense to make it the list of admins to get notified. It simplifies all the logic too!
A side effect is that non-admins *could* potentially get notifications.

sijis: I can change the behaviour to make it the notification list, as opposed to an exclusion list of admins.

errbot/bootstrap.py
```diff
@@ -64,6 +64,8 @@ def bot_config_defaults(config):
         config.TEXT_DEMO_MODE = True
     if not hasattr(config, 'BOT_ADMINS'):
         raise ValueError('BOT_ADMINS missing from config.py.')
+    if not hasattr(config, 'BOT_ADMINS_NOTIFICATIONS'):
+        config.BOT_ADMINS_NOTIFICATIONS = ()
```
gbin: One last very small thing: can we dafault it to BOT_ADMINS here ? So if someone doesn't touch anything he or she can still get those ?

sijis: I basically do that here:
https://github.com/sijis/err/blob/ffe23db0720d4cb90328e87896b2d5b437ada3c5/errbot/core.py#L565

it probably makes more sense to add it where you mention instead.

sijis:
I've made the suggested changes.

gbin:
thx

PR #869, opened 2016-10-11
WIP Feature: allow markdown by default in telegram_messenger

asfaltboy:
- also allow toggling other send options via config

### DIFF
--- a/errbot/backends/telegram_messenger.py
+++ b/errbot/backends/telegram_messenger.py
@@ -3,7 +3,7 @@ import sys
 
 from errbot.backends.base import RoomError, Identifier, Person, RoomOccupant, ONLINE, Room
 from errbot.core import ErrBot
-from errbot.rendering import text
+from errbot.rendering import md, text
 from errbot.rendering.ansiext import enable_format, TEXT_CHRS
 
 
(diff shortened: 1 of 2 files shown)


asfaltboy:
I found an issue with sending unicode characters to Telegram with the Markdown default

asfaltboy:
I encountered what I first thought was a unicode issue, but ended up being an issue with another  plugin. The plugin sent characters that telegram could not parse as correct markdown (e.g underscores that did not wrap around words, such as `this_stuff is_bad`).

This makes me think that we may want to remove the `parse_mode=Markdown` from default config and simply document that these methods are supported if users require it.

gbin:
Looks cool for me, I don't have a test setup for telegram, maybe @zoni can give it a shot ?

asfaltboy:
@gbin thanks, I did add a test case in ab511fe, however it only tests the custom send options feature I added, not the rest of the backend. I'm also using it in production, I don't however have many other plugins installed (just a few internal ones we've written and [`err-gitwebhook`][1]).

What do you think about changing the default parse_mode? What is the general best practice for `errbot` backends in regards to a default renderer?

[1]: https://github.com/asfaltboy/err-gitwebhook/tree/feature/gitlab-support

errbot/backends/telegram_messenger.py
```diff
@@ -191,9 +191,14 @@ def __init__(self, config):
         self.telegram = None  # Will be initialized in serve_once
         self.bot_instance = None  # Will be set in serve_once
 
-        compact = config.COMPACT_OUTPUT if hasattr(config, 'COMPACT_OUTPUT') else False
-        enable_format('text', TEXT_CHRS, borders=not compact)
-        self.md_converter = text()
+        self.default_send_options = dict(parse_mode='Markdown')
+        self.default_send_options.update(getattr(config, 'TELEGRAM_DEFAULT_SEND_OPTIONS', {}))
```
zoni: I would like to see `TELEGRAM_DEFAULT_SEND_OPTIONS` added to `config-template.py` with a little bit of accompanying documentation of what can go into it, so that this doesn't become a hidden feature.

errbot/backends/telegram_messenger.py
```diff
@@ -191,9 +191,14 @@ def __init__(self, config):
         self.telegram = None  # Will be initialized in serve_once
         self.bot_instance = None  # Will be set in serve_once
 
-        compact = config.COMPACT_OUTPUT if hasattr(config, 'COMPACT_OUTPUT') else False
-        enable_format('text', TEXT_CHRS, borders=not compact)
-        self.md_converter = text()
+        self.default_send_options = dict(parse_mode='Markdown')
```
zoni: I would like to see the default parse mode remain text in order to remain consistent with the current behavior (and because the markdown parsing mode Telegram has is a little iffy.. I deliberately used the text converter because it typically produced better results).

gbin:
Can we move forward on this one ?

asfaltboy:
Hi @gbin sorry completely forgot about this PR, will try to attend to zoni's request later this week.

gbin:
I am closing everything 2016, please reopen a clean PR if you want to resume it.

PR #871, opened 2016-10-13
Increase the efficiency of the Slack backend for busy instances

cniemira:
SlackClient.rtm_read ultimately calls a function that can only return one websocket frame at a a time. Used in this way, errbot will only process one message at a time. With a 1 sec pause between messages, errbot can end up taking a long time before it processes commands on a busy Slack instance.


gbin:
Nice ! Thank you !

PR #875, opened 2016-10-15
Don't sanitize slack file_share event texts.

### LINKED ISSUE #872: Slack backend sanitize_uris() breaks links of shared files
As far as I can tell, it's impossible to get the file URL for a shared file from the slack event.

example from the file_shared notification message:

```
 'text': '<@U18CKQZSM|cc> uploaded a file: '
         '<https://liveintent.slack.com/files/cc/F2PK11R6K/core.py|core.py>',
```

that message text in the event gets turned into:

```
@cc uploaded a file: core.py
```

The whole URL gets deleted by the regex at line 812 in sanitize_uris():

```
text = re.sub(r'<([^\|>]+)\|([^\|>]+)>', r'\2', text)
```

Maybe there should be a way to turn the sanitation off? I'm happy to send a PR with a fix if given some guidance on what the correct fix should be.

capdevc:
Fixes #872

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -423,7 +423,8 @@ class SlackBackend(ErrBot):
 
         text, mentioned = self.process_mentions(text)
 
-        text = self.sanitize_uris(text)
+        if subtype != 'file_share':
+            text = self.sanitize_uris(text)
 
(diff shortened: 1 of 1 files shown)


capdevc:
Issue is slightly more complicated than I thought... this PR doesn't fix things.

PR #878, opened 2016-10-17
Feature: use conftest to register fixture instead of importing

asfaltboy:
This is the de-facto method to register fixtures as far as I know, and should simplify writing tests, errbot's as well as for plugin developers.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,1 @@
+pytest_plugins = ['errbot.backends.test']
--- a/docs/user_guide/plugin_development/testing.rst
+++ b/docs/user_guide/plugin_development/testing.rst
@@ -38,24 +38,23 @@ This does absolutely nothing shocking, but how do you test it?
 We need to interact with the bot somehow, send it `!mycommand` and validate the reply.
 Fortunatly Errbot provides some help.
 
+
+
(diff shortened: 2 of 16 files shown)


zoni:
> This is the de-facto method to register fixtures as far as I know

This is neat. It's actually the first time I've come across it :grinning:

I wonder, how does this affect user of nose, if it affects them any at all?

asfaltboy:
It does not, as users of nose cannot use pytest fixtures at all sadly :( but then again nose is unmaintained and itself recommends [users try alternatives][1]. Which is why I'm happy we are testing with `py.test`! 👍 

Plugins can also be specified in a specific test module, like I did here: fded4e6 . And of course, fixtures defined in the same module do not need to be "registered" at all.

[1]: http://nose.readthedocs.io/en/latest/index.html#note-to-users

zoni:
> It does not, as users of nose cannot use pytest fixtures at all sadly

I was aware of that point, but I have so far ensured that users of nose can still run tests using our test setup. Looking at the code of how `errbot.backends.test.FullStackTest` is implemented though, I don't believe your changes would have any effect on that at all.

Travis is happy too, so I'm going to go ahead and merge this. :smile:

PR #880, opened 2016-10-18
readme: typo

profan:


zoni:
I completely missed this PR but just merged #882 which included this same fix along with a couple more so this one can be closed now.

Thanks for taking the effort however, much appreciated!

PR #881, opened 2016-10-21
Add initial ISSUE_TEMPLATE

sijis:
Create an issues template, so we can get as much data upfront from the reporter.

This should help us reproduce the issue.


zoni:
This has been on my todo list for so long, but I never took time to do it yet.... :smile: Thanks!

PR #886, opened 2016-10-27
Update dynaplugs.rst

andrewthetechie:
Note doesn't relate to the text above it. Appears to be a copy/paste error from another page


sijis:
I do wonder if the intent was to say that it still worked with
`!my_plugin_my_command1` and not just `!my_command1`.

I may be completely wrong and this is just a copy/paste mistake.

zoni:
Judging from the text and context, I do think it was simply a copy&paste mistake so I'm going to merge this one. Thanks Andrew.

PR #888, opened 2016-10-27
Add the packaging metadata to build the errbot snap

elopio:
This is a package for the secure installation of apps that works in most Linux distributions.
Landing it upstream will enable builds for adventurous users.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -18,5 +18,5 @@ parts:
   errbot:
     source: .
     plugin: python
-    build-packages: [libffi-dev, libssl-dev]
+    build-packages: [gcc, libffi-dev, libssl-dev]
     stage-packages: [git]
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -12,7 +12,7 @@ confinement: strict
 apps:
   errbot:
     command: env LC_ALL=C.UTF-8 errbot
-    plugs: [home, network]
+    plugs: [home, network, network-bind]
 
 parts:
   errbot:
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,0 +1,22 @@
+name: errbot
+version: master
+summary: chatbot that brings your tools into the conversation.
+description: |
+  Errbot is a chatbot. It allows you to start scripts interactively from your
+  chatrooms for any reason: random humour, chatops, starting a build, monitoring
+  commits, triggering alerts...
+
+grade: devel
+confinement: strict
+
+apps:
+  errbot:
+    command: env LC_ALL=C.UTF-8 errbot
+    plugs: [home, network]
+
+parts:
+  errbot:
+    source: .
+    plugin: python
+    build-packages: [libffi-dev, libssl-dev]
+    stage-packages: [git]


elopio:
You can find more information about snapcraft here: http://snapcraft.io/

If we upload the snap to the ubuntu store we will get a great way to deliver the software, and get transactional automatic updates. I did some initial testing for the snap [1]. It needs more testing, but it seems to me good enough to publish in the edge channel, which will make it really easy to get more people interested in finding and solving potential problems before the stable release.

If you have any doubts or comments, please let me know.

[1] https://gist.github.com/elopio/45032e7a6a0d23c117a62e44b14b4042

sijis:
I think it is a super interesting and i'm for it. 👍

zoni:
Snaps had already been on my radar so I'm glad to see more interest. :smile: There are some concerns that I think need to be addressed but I don't have time to write up a detailed response right now, so stay tuned, I will try to get back to this soon.

elopio:
I will be here ^_^
Thanks for reviewing.

snapcraft.yaml
```diff
@@ -0,0 +1,22 @@
...
+  errbot:
+    command: env LC_ALL=C.UTF-8 errbot
+    plugs: [home, network]
+
+parts:
+  errbot:
+    source: .
+    plugin: python
+    build-packages: [libffi-dev, libssl-dev]
+    stage-packages: [git]
```
zoni: I think we need to add the build-packages to stage-packages as well as they're needed by pip when installing plugins with extra dependencies. We'll probably also have to add some additional common ones (`libxslt-dev` is another common one which comes to mind.

We should probably add mysql (`libmysqlclient-dev`) and psql (not sure which package provides it) libraries as well so people can install python packages needing to link to those, such as when using our sql storage plugin.

snapcraft.yaml
```diff
@@ -0,0 +1,22 @@
...
+  chatrooms for any reason: random humour, chatops, starting a build, monitoring
+  commits, triggering alerts...
+
+grade: devel
+confinement: strict
+
+apps:
+  errbot:
+    command: env LC_ALL=C.UTF-8 errbot
+    plugs: [home, network]
```
elopio: Done.

snapcraft.yaml
elopio: ok, I can add them to this PR.
I personally prefer to do that as a separate PR, to better link each change with a specific requirement. This change works for a basic bot, I find it nice to keep track of which additions are needed for more complex one.

But, as I said, just let me know and I'll make the change :)

elopio:
Hi @zoni o/
Thanks for your review.

There are a few options to work with plugins. You can install pip as a stage package, and make it accessible to users as an app. Then they can call `$ errbot.pip install`, or you could hack something to make it like `$ errbot.install_plugin`. The modules will end up installed in the snap's writable path. This is good for pip, but we will have to think a little more about debs like libmysqlclient-dev.

Another option is to have an env var or config value for the location of your plugin modules. I haven't fully explored this one yet though. If we want the snap to access the user's PYTHONPATH for example, I think we will need an interface. It's an interesting thing to try, and I'm playing on my free time with an errbot juju charm based on this: https://github.com/1stvamp/juju-errbot I think this is the approach I'll try, and if the python modules are installed in $HOME, we don't need anything in addition to the home plug.

One that we have been experimenting is to have the plugins packaged as snaps, and share resources between the snaps with the content interface [1]. The downside is that of course it would require work from your ecosystem.

And this is my favorite approach: an errbot remote part [2]. You can provide a standalone errbot snap for simple bots and as a demo, or for the people who want to deal with independent plugins. But you can also upload your errbot part so others can easily make fully isolated complex bots. They will just have to add in their yaml "after: [errbot]", and they can pull their database dependencies, all their python modules, and use snap install my-bot to deploy. To make that work, we need better docs for sure.

Sorry for the long reply. Let me know what do you think.

pura vida.

[1] http://mhall119.com/2016/09/sharing-is-caring-with-snaps/
[2] http://snapcraft.io/docs/build-snaps/parts

elopio:
ping @zoni, would you have some time to jump into a hangout or chat? Maybe it will be easier to answer your questions interactively there :)

sijis:
@elopio, we have a gitter channel. I know @zoni is always available there.

zoni:
Yep! The aforementioned gitter chat is a good way to reach me. I also do hang out on freenode (even in `#snappy` :wink:) so you can hit me with a private query there, same nickname (`zoni`). European business hours are probably the best time to reach me, though evenings work too depending on how drained I am at the end of the day. :smile: 

And yes, I'd actually done a bit of experimenting with snaps for errbot based on this PR and have some thoughts on it, I just haven't had time to really write them out yet. But a realtime chat about it is probably more fruitful anyway!

elopio:
Hey @zoni, I finally got the time to play with my bot:
https://github.com/elopio/snappy-m-o/blob/master/snap/snapcraft.yaml

If you land this PR, I can make it simpler by adding errbot to the remote parts.

I'm now trying to add nginx and letsencrypt in a juju charm. The snap works nicely by just installing it, but I think I don't want to bundle all the other fancy things in the snap.

flexiondotorg:
I'd certainly be interested in seeing a snap for errbot.

gbin:
What is the status about this ? Should it be in a separate errbotio/snap-errbot repository or it needs to be in the pypi package ?

elopio:
Hey @zoni, remember me?
I finally got some time to write about the bot that we are using in snapcraft, and how we deploy it through the store:
http://elopio.net/blog/errbot-snap/

gbin:
@elopio this is really cool !

elopio:
Thanks @gbin!

I would like to add a section to the errbot docs about deploying with snaps. Like a much more focused version of my post. What do you think?

gbin:
It would make a lot of sense.

gbin:
Please go ahead with the doc version, I am closing this one.

PR #893, opened 2016-10-30
Playsafe

medecau:
This pull-request contains **security** related changes. One of which is on a certificate generating procedure.

I ask that at least two core developers sign-off on these changes.

---

- replace `random.random()` with `os.urandom()` in a crypto context
- replace `exec()` and `eval()` with `ast.literal_eval()`
- adds `autoescape` to jinja2 environments
- silences some security warnings

### DIFF
--- a/errbot/backends/irc.py
+++ b/errbot/backends/irc.py
@@ -190,7 +190,7 @@ class IRCRoom(Room):
         if username is not None:
             log.debug("Ignored username parameter on join(), it is unsupported on this back-end.")
         if password is None:
-            password = ""
+            password = str()
 
         self.connection.join(self.room, key=password)
         log.info("Joined room {}".format(self.room))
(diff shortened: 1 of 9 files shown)


errbot/backends/irc.py
```diff
@@ -190,7 +190,7 @@ def join(self, username=None, password=None):
         if username is not None:
             log.debug("Ignored username parameter on join(), it is unsupported on this back-end.")
         if password is None:
-            password = ""
+            password = str()
```
zoni: I'm curious, what's the rationale behind this one?

medecau: Not much really.
The options are `str()` OR `""` followed by a `nosec` comment and I'm trying to avoid using those.

gbin:
Thanks for the contribution, I am ok with it, maybe except the empty string, I don't see why it is a security issue, do you have the full linter message on it or any reference regarding the issue ? Has it something to do with object identity somewhere ?

medecau:
@gbin I no longer have the linter on hand but I remember the nature of the message. 

`""` was considered a potential default password. `str()` creates a new `str` object and no longer considered a default password.

Issue #907, opened 2016-11-23
Webhook raises exceptions when used in lambdas


txomon:
In order to let us help you better, please fill out the following fields as best you can:

### I am running...
* Errbot version: master
* OS version: archlinux
* Python version: 3.5.2
* Using a virtual environment: yes

### Issue description

Currently when binding a webhook, it assumes `__name__` attribute is present. This is only true for functions, lambdas don't have such attribute.

### Steps to reproduce

webhook(lambda x: 'Hi')

### Additional info

This can be fixed by using getattr() with a default set in https://github.com/errbotio/errbot/blob/master/errbot/__init__.py#L392

gbin:
Well spotted ! Can you make a quick pull request with a little test ?

txomon:
I was wrong, lambdas do have `__name__` unless you use `functools.wraps()` on them without argument...

Either way, I am submitting the patch because at least we have one more test.

PR #908, opened 2016-11-24
Adding lambda webhook test

txomon:
This is the test about webhook lambda functions in #907. It was my bad, but at least now it will properly log that.


gbin:
thx

Issue #911, opened 2016-11-26
Slack identifiers errors on picking


gbin:
This breaks the backend contract:

```
  File "/home/gbinet/projects/errbot-plugins/err-presence/presence.py", line 8, in activate
    self['blah'] = self.build_identifier('@gbin')
  File "/home/gbinet/projects/errbot/errbot/storage/__init__.py", line 58, in __setitem__
    return self._store.set(key, item)
  File "/home/gbinet/projects/errbot/errbot/storage/shelf.py", line 27, in set
    self.shelf[key] = value
  File "/usr/lib/python3.4/shelve.py", line 124, in __setitem__
    p.dump(value)
_pickle.PicklingError: Can't pickle <function SlackBackend._unpickle_identifier at 0x7fabbe326048>: attribute lookup _unpickle_identifier on yapsy_loaded_plugin_Slack_0 failed
10:33:05 INFO     errbot.core               Error: Presence failed to start: Can't pickle <function SlackBackend._unpickle_identifier at 0x7fabbe326048>: attribute lookup _unpickle_identifier on yapsy_loaded_plugin_Slack_0 failed
```

zoni:
I believe this has been fixed now.

PR #912, opened 2016-11-26
Statified slack_client

gbin:
We had a problem with the Slack Backend because a lot of identifier related
manupulations needs a live slack client but pickling and unpickling
need static functions.

In this PR I push slack_client as a global for the module and cascaded
the change through all the identifier creations and transformations.

This simplifies the code and made the pickling/unpickling code less
finicky at the expense of the structure of the code (basically the whole
backend is a huge static blob now).

This corrects #911.

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -618,7 +618,10 @@ class SlackBackend(ErrBot):
         try:
             if mess.is_group:
                 to_channel_id = mess.to.id
-                to_humanreadable = mess.to.name if mess.to.name else SlackBackend.channelid_to_channelname(to_channel_id)
+                if mess.to.name:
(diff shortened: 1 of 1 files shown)


gbin:
@zoni I know it is ugly but do you have an opinion on this ?

zoni:
I'm really not a fan of that singleton being used all over the place. Although it would work for now, if we ever want to make an architecture where you can be connected to multiple backends at once then this would block that.

Is there no other way we could work around this? Or maybe we need to revisit the rule that identifiers need to be pickle-able? I forsee this being a problem with other backends as well. Maybe we could solve this without making the code ugly by advising people to store the string representation when persisting it anywhere and using `build_identifier()` on that string again when loading it in again. (That actually seems much safer/future-proof anyway).

PR #913, opened 2016-11-26
Fix slack hyperlinks

### LINKED ISSUE #834: Markdown conversion in slack is broken if text start directly with a link
Testing in slack, if the markdown string starts with a link the conversion is not made correctly:

```
echo [a link](http://www.errbot.net)
a link (http://www.errbot.net)
```

If the text start with any other character this works fine:

```
echo link: This is [a link](http://www.errbot.net).
link: This is a link.
```

Looks like the regexp is checking that the link doesn't start with an exclamation mark is causing that. It also expect to find any character just before the link square bracket orelse it doesn't match (slack.py:31)

```
MARKDOWN_LINK_REGEX = re.compile(r'([^!])\[(?P<text>.+?)\]\((?P<uri>[a-zA-Z0-9]+?:\S+?)\)')
```

r3m0t:
Fixes #834 .

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -61,7 +61,7 @@ COLORS = {
     'cyan': '#00FFFF'
 }  # Slack doesn't know its colors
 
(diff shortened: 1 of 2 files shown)


zoni:
Great, thanks!

PR #917, opened 2016-11-27
Slack -- add ts to incoming messages so that bot can send reactions

r3m0t:
On Slack a useful alternative to an "OK" reply is a reaction.


errbot/backends/slack.py
```diff
@@ -838,6 +838,26 @@ def build_reply(self, mess, text=None, private=False):
...
+                                  timestamp=mess.extras['ts'],
+                                  name=reaction)
+        if result['ok']:
+            return
+        elif result.get('error') == 'invalid_name':
+            raise ValueError(result['error'], 'No such emoji', reaction)
+        elif result.get('error') in ('no_reaction', 'already_reacted'):
+            raise LookupError(result['error'])
+        else:
+            raise Exception(result['error'])
```
zoni: Can you raise [SlackAPIResponseError](https://github.com/r3m0t/errbot/blob/64b89797abad251fcd3eee4f7f9526fe71d18e62/errbot/backends/slack.py#L89-L98) here instead?

errbot/backends/slack.py
```diff
@@ -838,6 +838,26 @@ def build_reply(self, mess, text=None, private=False):
...
+        result = self.sc.api_call(api_call,
+                                  channel=mess.to.channelid,
+                                  timestamp=mess.extras['ts'],
+                                  name=reaction)
+        if result['ok']:
+            return
+        elif result.get('error') == 'invalid_name':
+            raise ValueError(result['error'], 'No such emoji', reaction)
+        elif result.get('error') in ('no_reaction', 'already_reacted'):
+            raise LookupError(result['error'])
```
zoni: I think a `ValueError` makes more sense here as well.

errbot/backends/slack.py
```diff
@@ -838,6 +838,26 @@ def build_reply(self, mess, text=None, private=False):
...
+        return self._react('reactions.add', mess, reaction)
+
+    def remove_reaction(self, mess, reaction):
+        return self._react('reactions.remove', mess, reaction)
+
+    def _react(self, api_call, mess, reaction):
+        result = self.sc.api_call(api_call,
+                                  channel=mess.to.channelid,
+                                  timestamp=mess.extras['ts'],
+                                  name=reaction)
```
zoni: I would rather make use of the [self.api_call()](https://github.com/r3m0t/errbot/blob/64b89797abad251fcd3eee4f7f9526fe71d18e62/errbot/backends/slack.py#L303-L334) helper function here.

r3m0t:
I will edit the docs, basically we have a command `!notify [task-name]` that notifies the user privately when the task (that already exists) completes, a few minutes later. It looks odd without an immediate reply, but the replies clog up the channel. So, instead of the `notify` botcmd returning a message, it returns `None` and calls `self._bot.add_reaction(msg, 'ok_hand')`. There are other potential uses as well.

errbot/backends/slack.py
r3m0t: Sure, but I don't understand how plugins can import SlackAPIResponseError in case they want to catch it, given yapsy?

r3m0t:
I couldn't build the docs in my local environment, so there might be an issue there. I also saved `ts` on outgoing messages, but it's almost impossible to use as even if you pass in a `Message` object, it gets replaced multiple times, and doesn't get returned. I'm sure other backends support editing messages as well, so perhaps it would be useful to return `Message` instead of `None` in some places? But, I wouldn't have any use for that feature as of now.

Here's my final test plugin:

```python
from errbot import BotPlugin, botcmd, Message
from time import sleep

class Example(BotPlugin):
    @botcmd
    def longcompute(self, mess, args):
        if self._bot.mode == "slack":
            self._bot.add_reaction(mess, "hourglass")
        else:
            yield "Finding the answer..."
        M = Message(body='hi', to=mess.frm)
        self._bot.send_message(M)
        self._bot.add_reaction(M, "ok_hand")
        sleep(10)
        yield "The answer is: 42"
        if self._bot.mode == "slack":
            self._bot.remove_reaction(mess, "hourglass")
```

r3m0t:
@zoni I added another member to `Message.extras` on received messages which is a permalink to the message. Please re-review?

zoni:
I will try to review this somewhere over the coming few days, currently really swamped which isn't leaving me much time. Please poke me if I end up taking too long to get back to this.

errbot/backends/slack.py
```diff
@@ -505,7 +505,8 @@ def _message_event_handler(self, event):
                 msg.frm = SlackRoomBot(
                     self.sc,
                     bot_id=event['bot_id'],
-                    bot_username=event['username'],
+                    bot_username=event.get('username', 'bot'),  # SlackClient doesn't keep the bot data from the login
```
r3m0t: I will move this unrelated change

errbot/backends/slack.py
```diff
@@ -1014,6 +1021,8 @@ def id(self):
             self._id = self._channel.id
         return self._id
 
+    channelid = id
```
r3m0t: this might be removable

r3m0t:
@zoni happy new year! I probably have time next week to work on this. What would you like done if anything?

r3m0t:
@zoni poke

errbot/backends/slack.py
zoni: I think this can be removed, yes?

errbot/backends/slack.py
zoni: It doesn't look like this is used anywhere, can you see if it's possible to remove it?

rcoup:
Did this get dropped? Or merged in somewhere else?

r3m0t:
I moved my fork into my employer's account. I'll open a new pull request.

Issue #892, opened 2016-10-30
Slack backend should convert @mentions and #channel names in outgoing messages


zoni:
On slack, if a plugin sends a response such as `Hi @zoni, how are you?` the @zoni part is not recognized by slack as a user handle automatically. In order to make this work we must convert @zoni into the `<U123456|zoni>` representation which is basically doing the inverse operation of what we do on incoming messages.

Although I have not verified, I believe the very same applies to channels as well.

---

See also: https://github.com/errbotio/errbot/issues/684#issuecomment-250215872

r3m0t:
Why not just send `link_names=1` on the `chat.postMessage` API call? Also, I have found sending `<U123456>` is left unchanged by the markdown converter and looks correct on the client. It would be useful to have a property giving that text on the `SlackPerson` object.

zoni:
> Why not just send `link_names=1` on the `chat.postMessage` API call?

Huh, I either overlooked that option entirely or it didn't exist back when I implemented the relevant portions of the Slack backend originally. :smile: 

Would you have time to try that out and seeing if it correctly links names/channels and opening a PR then? If not, I will get around to this myself somewhere over the coming couple of weeks for sure.

PR #918, opened 2016-11-29
Slack -- link names

r3m0t:
Fixes https://github.com/errbotio/errbot/issues/892

I have tested this and @ names and # channels link fine, as well as `@here, @everyone` etc. The `<U398232>` and `<!here>` formats still work too, although other formats are mangled by the markdown formatter, as discussed on other cards.

Test plugin:

```python
from errbot import BotPlugin, botcm
import ast

class Example(BotPlugin):
    @botcmd
    def say(self, mess, args):
        result = ast.literal_eval(args.strip('`'))
        yield result
        self.send_card(title='title ' + result,
                body='body ' + result,
                color='blue',
                in_reply_to=mess)
```

Test messages: 

```
!say `'hi #' 'team'`
!say `'hi @' 'here'`
```

The backticks and the space prevent Slack from mangling my message on its way out - smart quotes, highlighting etc.

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -633,6 +633,7 @@ class SlackBackend(ErrBot):
                     'channel': to_channel_id,
                     'text': part,
                     'unfurl_media': 'true',
+                    'link_names': '1',
                     'as_user': 'true',
                 })
         except Exception:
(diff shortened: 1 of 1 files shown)


zoni:
Awesome! Thanks so much.

PR #921, opened 2016-12-02
issues-920 Add setting in the flows so they don't prompt for the next step

### LINKED ISSUE #920: Add setting in the flows so they don't prompt for the next step
### I am...

* [x] Suggesting a new feature

### I am running...

* Errbot version: 4.3.4
* OS version:  MacOS Sierra
* Python version: 3.5 
* Using a virtual environment: yes

### Issue description

While in flow, there is always a prompt to the next step. I would like to add an flag to make this optional

avivl:
Added dismiss_prompt variable (default False).

Fixes #920

### DIFF
--- a/errbot/flow.py
+++ b/errbot/flow.py
@@ -122,17 +122,21 @@ class Flow(object):
     This is a live Flow. It keeps context of the conversation (requestor and context).
     Context is just a python dictionary representing the state of the conversation.
     """
-    def __init__(self, root: FlowRoot, requestor: Identifier, initial_context: Mapping[str, Any]):
+    def __init__(self, root: FlowRoot, requestor: Identifier, initial_context: Mapping[str, Any],
+                 next_step_hinting: int=True):
         """
 
         :param root: the root of this flow.
         :param requestor: the user requesting this flow.
         :param initial_context: any data we already have that could help executing this flow automatically.
+        :param next_step_hinting: toggle display of the "you are now in the flow xxx, you may continue with yyy, zzz"
+        messages after each flow step
         """
         self._root = root
         self._current_step = self._root
         self.ctx = dict(initial_context)
         self.requestor = requestor
+        self.next_step_hinting = next_step_hinting
 
     def next_autosteps(self) -> List[FlowNode]:
         """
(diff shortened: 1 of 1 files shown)


avivl:
@oryband @gbin

errbot/flow.py
```diff
@@ -122,7 +122,7 @@ class Flow(object):
     This is a live Flow. It keeps context of the conversation (requestor and context).
     Context is just a python dictionary representing the state of the conversation.
     """
-    def __init__(self, root: FlowRoot, requestor: Identifier, initial_context: Mapping[str, Any]):
+    def __init__(self, root: FlowRoot, requestor: Identifier, initial_context: Mapping[str, Any], dismiss_prompting=False):
```
zoni: PEP8 error (see Travis build output): `errbot/flow.py:125:121: E501 line too long (123 > 120 characters)`

errbot/flow.py
```diff
@@ -390,7 +391,8 @@ def execute(self, flow: Flow):
                     if syntax_args:
                         syntax += syntax_args
                     possible_next_steps.append("- %s" % syntax)
-                self._bot.send(flow.requestor, "\n".join(possible_next_steps))
+                if not flow.dismiss_prompting:
+                   self._bot.send(flow.requestor, "\n".join(possible_next_steps))
```
zoni: PEP8 error (see Travis build output): `errbot/flow.py:395:20: E111 indentation is not a multiple of four`

errbot/flow.py
zoni: The naming `dismiss_prompting` makes us talk in double negatives (see below, `if not flow.dismiss_prompting`). I think it will read cleaner if we instead name it `next_step_hinting` (defaulting to `True`) so you can initialize a Flow with `Flow(..., next_step_hinting=False)` when you want to disable it.

In my experience this reads better and is easier to reason about.

zoni: We're also trying to have [type hints](https://www.python.org/dev/peps/pep-0484/) on all our (at least user-facing) APIs, so we should define this as such:

```python
def __init__(self, root: FlowRoot, requestor: Identifier, initial_context: Mapping[str, Any],
                    next_step_hinting: int=True):
```

We should also document the parameter in the docstring below. Something like:

`:param next_step_hinting: toggle display of the "you are now in the flow xxx, you may continue with yyy, zzz" messages after each flow step`

avivl:
@zoni comments were addressed

zoni:
Super!

PR #923, opened 2016-12-04
Use literalinclude of code samples in testing docs

sijis:
Trying to simplify the testing docs code samples.

### DIFF
--- a/docs/code_examples/plugin_development/myplugin.plug
+++ b/docs/code_examples/plugin_development/myplugin.plug
@@ -1,0 +1,6 @@
+[Core]
+Name = MyPlugin
+Module = myplugin
+
+[Documentation]
+Description = my plugin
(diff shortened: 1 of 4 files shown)


gbin:
does it work ? can we merge it ?

sijis:
I want to complete the other examples we have. I'd also have to rebase too.
Give me a few days to finalize it.

sijis:
I think this is good to merge.

I'll put in another PR to convert the flow examples.

sijis:
You make some very fair points.
I was looking at this from 2 points of view:
1. It would be great to see the completed myplugin example at the end
2. There quite a bit of duplication of code and I was hoping to simplify that a bit

It makes complete sense that if the code files change, now we'll have to scour the docs to the related line changes that are affected. This is potentially more work than simply copy&pasting or grepping for a specific string to update.

I think that point alone makes me agree with you to not include these changes.

PR #924, opened 2016-12-04
slack -- block on reads on the RTM API

r3m0t:
This reduces delay responding to Slack messages by about half a second on average.

I have tested it and it's noticeably faster, I also tested master with the sleep set to 5 instead of the original 1 and it's very noticeably slower, so I'm confident in this one.


zoni:
I'm going to have total faith in your test and just merge this in now without testing it beforehand myself. :grinning: 

I'll update the slack instance we operate at my employer to run off the latest too to catch any potential problems if this does end up doing something unpredicted.

PR #925, opened 2016-12-07
Adding support for stream requests in test backend

tlee911:


gbin:
This is a good idea !

Can we maybe protect it a little bit ? Like for example what if the plugin uploads a 10MB binary file ? etc.
I think a simple "head" of the file with a filter on printable characters should do the trick...

Something along this ... (untested)
```
import string
self.outgoing_message_queue.put(filter(lambda x: x in string.printable, fsource.read(1024)))
```

tlee911:
> Can we maybe protect it a little bit ? Like for example what if the plugin uploads a 10MB binary file ? etc.
> I think a simple "head" of the file with a filter on printable characters should do the trick...

Hmm... I'm inclined to dump the whole file and let the consumer deal with it, as this is the test backend.  For me I'm just looking at text files, but the 1 kB limit above causes some of my tests to fail already, because I'm not getting the full contents.

Also, will non-printable characters cause problems?  I wonder if this could also be used by someone interested in testing other types of files (say binary) and not constrain it that way.

gbin:
Indeed it is just for the test backend anyway, if the user tests with a huge file, it is his/her problem.

Thanks !

PR #931, opened 2016-12-16
Adding a callback allow plugins to catch unhandled messages

tamarintech:
I would like the ability to define a plugin callback that will catch any unhandled messages.

The goal is to allow a plugin to implement its own fuzzy matching, contextual services, or tasks (ie: user dictionary) where creating multiple plugins would be impractical.

### DIFF
--- a/tests/commandnotfound_plugin/commandnotfound.py
+++ b/tests/commandnotfound_plugin/commandnotfound.py
@@ -2,11 +2,7 @@ from errbot import BotPlugin, cmdfilter
 
 
 class TestCommandNotFoundFilter(BotPlugin):
-    def __init__(self, *args, **kwargs):
-        super(TestCommandNotFoundFilter, self).__init__(*args, **kwargs)
-        self.catch_unprocessed = True
-
-    @cmdfilter
+    @cmdfilter(catch_unprocessed=True)
     def command_not_found(self, msg, cmd, args, dry_run, emptycmd=False):
         if not emptycmd:
             return msg, cmd, args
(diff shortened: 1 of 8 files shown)


gbin:
This is a good idea, thanks for the PR, but I would go a little deeper: why not make the error handling of the command not found of errbot use the same mechanism using a core plugin?
It would allow users to disable the plugin and replace it with their own (ACLs work that way for example).

Also, maybe we can extend the command filters with a way of saying "command does not exist", it was a feature request at one point (a catch all for commands). The advantage of this could be to be able to correct a command on the fly, treat it locally or just respond with an error message.

tamarintech:
To be honest, I'm not entirely sure how to proceed.  I'm not experienced enough with ErrBot internals for that to mean much to me. :)

Did we want to close this PR and we can sort out how to add this feature with a better implementation later?  I'll reopen if I can work out something closer to what @gbin describes, or someone else can open if they beat me to it. :)

tests/commandnotfound_plugin/commandnotfound.py
```diff
@@ -0,0 +1,14 @@
+from errbot import BotPlugin, cmdfilter
+
+
+class TestCommandNotFoundFilter(BotPlugin):
+    def __init__(self, *args, **kwargs):
+        super(TestCommandNotFoundFilter, self).__init__(*args, **kwargs)
+        self.catch_unprocessed = True
+
+    @cmdfilter
```
gbin: Maybe adding a parameter to the decorator is better that a boolean to the plugin like: @cmdfilter(not_found=true) ?

tamarintech: That makes more sense.
I wasn't a big fan of that __init__/__self__ thing I had done originally...

Issue #932, opened 2016-12-16
Issue with uploading file using Slack client


Xorso:
In order to let us help you better, please fill out the following fields as best you can:

### I am...

* [x] Reporting a bug
* [ ] Suggesting a new feature
* [ ] Requesting help with running my bot
* [ ] Requesting help writing plugins
* [ ] Here about something else

### I am running...

* Errbot version: 4.3.4
* OS version: macOS 10.12.1
* Python version: 3.5.0
* Using a virtual environment: yes

### Issue description

Error when posting a file attachment to slack:

Error: Slack API call to files.upload failed: no_file_data

### Steps to reproduce

Submit a BytesIO file to the stream

```python
self.send_stream_request(
                msg.frm,
                BytesIO(r.content),
                name=filename,
            )
```

### Additional info

It looks like the issue may be within the plugin. I wanted to verify before submitting a pull request

Line: 573 in slack.py
```python
            resp = self.api_call('files.upload', data={
                'channels': stream.identifier.channelid,
                'filename': stream.name,
                'file': stream
                })
```

I believe the issue is with `'file': stream` it should be `'file': stream.raw`.

jasedit:
I'm poking around at this to try and help - do you happen to know what the type of `r.content` is when you call `send_stream_request`? If it's a file object, does the code work if you don't wrap it in a `BytesIO` object?

andrewthetechie:
@Xorso we've merged a fix for this issue in #1283 

#closeit

PR #933, opened 2016-12-16
Fix for slack backend send_stream_request.

Xorso:
This is to fix the issue with posting attachments via slack.

Fix for #932

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -648,7 +648,7 @@ class SlackBackend(ErrBot):
             resp = self.api_call('files.upload', data={
                 'channels': stream.identifier.channelid,
                 'filename': stream.name,
-                'file': stream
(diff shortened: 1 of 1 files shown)


zoni:
@jasedit you contributed the Slack stream implementation in #837. Do you have any insight on this?

jasedit:
Digging into the underlying libraries used (this call maps down to the [requests](http://docs.python-requests.org/en/master/) package,) the `file` argument needs to be some type of file object, which `BytesIO` doesn't appear to fulfill. The `raw` accessor isn't an official part of the API (and isn't supported on all platforms,) which presents an issue as well (and making the change at the internal level would break passing any actual file object to the call.)

I would have to see a bit more detail to suggest an alternate route, but I don't believe this is a general fix for the problem.

Issue #928, opened 2016-12-14
slackclient update broken user lookup


rkrambovitis:
### I am...
* [x] Reporting a bug
* [ ] Suggesting a new feature
* [ ] Requesting help with running my bot
* [ ] Requesting help writing plugins
* [ ] Here about something else

### I am running...
* Errbot version: latest
* OS version: ubuntu xenial
* Python version: 3.5
* Using a virtual environment: yes

### Issue description
The following commit from python-slackclient changes slack users from searchList to searchDict.
https://github.com/slackapi/python-slackclient/commit/e13d3a8d67afaac287060a0221a1e1388522ed80
This results in slack.py user lookups being broken. Both username to id and id to username fail.

### Steps to reproduce
Update python-slackclient, try to use bot

### Additional info
The following patch fixes the username_to_userid, however I didn't manage to patch the opposite. My python-foo is pretty basic.
```
@@ -526,10 +527,10 @@
     def username_to_userid(self, name):
         """Convert a Slack user name to their user ID"""
         name = name.lstrip('@')
-        user = [user for user in self.sc.server.users if user.name == name]
+        user = self.sc.server.users.find(name)
         if not user:
             raise UserDoesNotExistError("Cannot find user %s" % name)
-        return user[0].id
+        return user.id

     def channelid_to_channelname(self, id_):
         """Convert a Slack channel ID to its channel name"""
```

### Temporary Workaround
I have updated my dockerfile to use the last good python-slackclient commit while building, rather than latest master.zip.  
https://github.com/slackapi/python-slackclient/archive/15aa80bed5433cdbd54505ce55e7075f8d84d648.zip

zoni:
Lovely, they changed their API in a patch-level release. :disappointed: 

Thanks for diving into this, should not be too terribly difficult to fix this. We'll need some logic to handle both versions, but that shouldn't be too difficult to implement.

This is a pretty big deal so I'll try to work on this soon.

(In the meantime, anyone hitting this can downgrade to a supported version of slackclient using: `pip install slackclient==1.0.2`)

antweiss:
Just hit this too. My first attempt with errbot. Spent an hour debugging. :) Learned a couple of things about slackclient and errbot. Looking forward to having an official fix.

gbin:
I am doing a dot release for this, this gonna hit a lot of people.

PR #936, opened 2016-12-21
Slack: compatibility with slackclient > 1.0.2

zoni:
Slackclient changed their API with the 1.0.2 →  1.0.3 release (see https://github.com/slackapi/python-slackclient/pull/149) which resulted in bug #928. These changes make our slack backend work again with this newer version of slackclient while maintaining compatibility with older versions of the library.


zoni:
> Please make a task for the next release to remove the backward compatibility, thx !

Yes, I'll remove it in a future version. There's another backward compatibility piece related to slackclient in there elsewhere so I'll remove both when that time comes.

PR #941, opened 2017-01-04
Add send_stream_request to Telegram backend

alexsavio:


errbot/backends/telegram_messenger.py
```diff
@@ -339,6 +340,121 @@ def prefix_groupchat_reply(self, message, identifier):
...
+                                          **kwargs)
+        elif msg_type == 'sticker':
+            msg = self.telegram.sendSticker(chat_id=chat_id,
+                                            sticker=content,
+                                            **kwargs)
+        elif msg_type == 'location':
+            msg = self.telegram.sendLocation(chat_id=chat_id,
+                                             latitude=kwargs.pop('latitude', ''),
+                                             longitude=kwargs.pop('longitude', ''),
+                                             **kwargs)
```
zoni: I think we should add a final `else:` clause here which raises an exception, saying an invalid `msg_type` was specified. Otherwise the line below is going to raise a very confusing error because `msg` is not defined.

errbot/backends/telegram_messenger.py
```diff
@@ -339,6 +340,121 @@ def prefix_groupchat_reply(self, message, identifier):
...
+
+    def _telegram_upload_stream(self, stream, **kwargs):
+        """Perform upload defined in a stream."""
+        msg = None
+        try:
+            stream.accept()
+            self._telegram_special_message(chat_id=stream.identifier.id,
+                                           content=stream.raw,
+                                           msg_type=stream.stream_type,
+                                           **kwargs)
```
zoni: Shouldn't the result of `self._telegram_special_message(....)` be assigned to `msg`?

From what I see, the `if msg is None` check 5 lines below will always return True now, result in `stream.error()` being called.

alexsavio:
Sorry, both pushes are the same, one is because of a branch merge.

errbot/backends/telegram_messenger.py
alexsavio: Done

errbot/backends/telegram_messenger.py
alexsavio: Yes. you are right. fixed!

zoni:
Thanks so much for adding this!

PR #945, opened 2017-01-18
Fix a bug in the Telegram send_stream_request

alexsavio:
Hi,

Sorry, I didn't test this in a room. Had to add this fix.


alexsavio:
I don't understand why it's adding all the previous commits. :(

PR #952, opened 2017-01-26
wip: add callback_stream to the Telegram backend

alexsavio:
Hi,

I've been trying to add stream_callback to the Telegram backend.

I am still seeing an error when I try to download the file. The message is:
```
urllib3 HTTPError No host specified.
```
Any idea?

I will keep checking.

### DIFF
--- a/errbot/backends/telegram_messenger.py
+++ b/errbot/backends/telegram_messenger.py
@@ -1,5 +1,8 @@
-import logging
+
+import os
 import sys
+import logging
+import tempfile
 from threadpool import WorkRequest
 
 from errbot.backends.base import RoomError, Identifier, Person, RoomOccupant, Stream, ONLINE, Room
@@ -257,26 +260,35 @@ class TelegramBackend(ErrBot):
             log.warning("Unhandled message type (not a text message) ignored")
             return
 
-        message_instance = self.build_message(message.text)
+        if message.text:
+            message_instance = self.build_message(message.text)
+            message_instance.frm, message_instance.to = self._reply_identifiers(message)
+            self.callback_message(message_instance)
+        else:
+            self._handle_file_transfer(message)
+
+    def _reply_identifiers(self, message):
+        """ Return (frm, to) to build a reply to `message`. """
         if message.chat['type'] == 'private':
-            message_instance.frm = TelegramPerson(
+            frm = TelegramPerson(
                 id=message.from_user.id,
                 first_name=message.from_user.first_name,
                 last_name=message.from_user.last_name,
                 username=message.from_user.username
             )
-            message_instance.to = self.bot_identifier
+            to = self.bot_identifier
         else:
             room = TelegramRoom(id=message.chat.id, title=message.chat.title)
-            message_instance.frm = TelegramMUCOccupant(
+            frm = TelegramMUCOccupant(
                 id=message.from_user.id,
                 room=room,
                 first_name=message.from_user.first_name,
                 last_name=message.from_user.last_name,
                 username=message.from_user.username
             )
-            message_instance.to = room
-        self.callback_message(message_instance)
+            to = room
+
+        return frm, to
 
     def send_message(self, mess):
         super().send_message(mess)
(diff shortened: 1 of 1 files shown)


alexsavio:
Hi, I need some feedback from you for this. I couldn't find any backend that is able to to make the bot retrieve files, i.e., the callback_stream function.  Is there?

I am working on the Telegram backend and my problem is that I don't know how to create a `io.BinaryIO` from the `telegram.bot.getFile` function. The telegram API, later with `file.download()` directly returns me a file. 

I think the simplest solution here would be to create a class with the same interface as Stream, but named, e.g., `TelegramStream`, which holds the `telegram.File` object and calls `file.download` in the `stream.accept` function. 

What do you think?

Thank you!

errbot/backends/telegram_messenger.py
```diff
@@ -369,6 +416,12 @@ def _telegram_special_message(self, chat_id, content, msg_type, **kwargs):
                                              latitude=kwargs.pop('latitude', ''),
                                              longitude=kwargs.pop('longitude', ''),
                                              **kwargs)
+        elif msg_type == 'contact':
```
gbin: Is it linked to file transfers ? if not please split it in another pull request so we can handle the 2 changes in dependently.

errbot/backends/telegram_messenger.py
```diff
@@ -392,7 +445,8 @@ def _telegram_upload_stream(self, stream, **kwargs):
             else:
                 stream.success()
 
-    def send_stream_request(self, identifier, fsource, name='file', size=None, stream_type=None):
+    def send_stream_request(self, identifier, fsource, name='file', size=None,
```
gbin: I don't think it needs reformatting.

errbot/backends/telegram_messenger.py
```diff
@@ -411,13 +465,17 @@ def send_stream_request(self, identifier, fsource, name='file', size=None, strea
             This is only used for debug logging purposes.
 
         :param stream_type: str, optional
-            Type of the stream. Choices: 'document', 'photo', 'audio', 'video', 'sticker', 'location'.
+            Type of the stream. Choices: 'document', 'photo', 'audio', 'video',
+                                         'sticker', 'location', 'contact'.
```
gbin: I don't think it needs reformatting.

errbot/backends/telegram_messenger.py
```diff
@@ -411,13 +465,17 @@ def send_stream_request(self, identifier, fsource, name='file', size=None, strea
...
 
             If 'video', a dict is optional as {'content': fsource, 'duration': str}.
             If 'voice', a dict is optional as {'content': fsource, 'duration': str}.
-            If 'audio', a dict is optional as {'content': fsource, 'duration': str, 'performer': str, 'title': str}.
+            If 'audio', a dict is optional as {'content': fsource, 'duration': str,
+                                               'performer': str, 'title': str}.
```
gbin: same

errbot/backends/telegram_messenger.py
```diff
@@ -446,18 +504,32 @@ def _is_valid_url(url):
...
+            self.thread_pool.putRequest(WorkRequest(self._telegram_upload_stream,
+                                                    args=(stream,)))
 
         return stream
 
+    def callback_message(self, mess):
+        """Processes for commands and dispatches the message to all the plugins."""
+        if self.process_message(mess):
+            # Act only in the backend tells us that this message is OK to broadcast
+            self._dispatch_to_plugins('callback_message', mess)
```
gbin: It should be handled by https://github.com/errbotio/errbot/blob/master/errbot/core.py#L570 the same way

errbot/backends/telegram_messenger.py
```diff
@@ -446,18 +504,32 @@ def _is_valid_url(url):
...
+
+    def callback_stream(self, stream):
+        """Dispatch the stream to all the plugins."""
+        if stream:
+            # Act only in the backend tells us that this message is OK to broadcast
+            self._dispatch_to_plugins('callback_stream', stream)
```
gbin: It is not possible to stream it like https://github.com/errbotio/errbot/blob/master/errbot/core.py#L615 ?

gbin:
The download fonction can get a file-like object so it should work (use the parameter out): 

```
    def download(self, custom_path=None, out=None, timeout=None):
        """
        Download this file. By default, the file is saved in the current working directory with its
        original filename as reported by Telegram. If a ``custom_path`` is supplied, it will be
        saved to that path instead. If ``out`` is defined, the file contents will be saved to that
        object using the ``out.write`` method. ``custom_path`` and ``out`` are mutually exclusive.
        Args:
            custom_path (Optional[str]): Custom path.
            out (Optional[object]): A file-like object. Must be opened in binary mode, if
                applicable.
            timeout (Optional[int|float]): If this value is specified, use it as the read timeout
                from the server (instead of the one specified during creation of the connection
                pool).
```

errbot/backends/telegram_messenger.py
alexsavio: what do you mean? this line is from a previous PR already merged.

alexsavio:
You are looking at the `master` version of the API. The version 5.3.0, which is the one from PyPI, doesn't have the `out` parameter:
https://github.com/python-telegram-bot/python-telegram-bot/blob/v5.3.0/telegram/file.py

errbot/backends/telegram_messenger.py
alexsavio: oh, sorry, okay. it's not exactly linked to file transfer, it works similarly to location and venue. they are just metadata. the API functionality is similar though. I will remove this, for a next PR.

gbin:
Can you make it work with their master ? Once you have it we can ping them, maybe they can release it (the last release was ~2 month ago).

alexsavio:
Ok. I will try. Thanks!

alexsavio:
Hi @gbin, thanks for your help so far. Where should I put the call to the `File.download` function?
I still think I need to create a different Stream class.

alexsavio:
any help with this, please?

alexsavio:
anyone?

gbin:
@zoni maybe ?

gbin:
This gone stale, I'd love to see that implemented, please reopen a PR on the current master version of errbot.

PR #958, opened 2017-02-03
Fix for Render Test

machbio:
Render test does not work because of the missing Markdown file from 'core_plugins' directory


gbin:
thx

PR #959, opened 2017-02-05
Attach the original slack event object to Message

bshelton229:
I set out to create what I believed would be a simple plugin; "detect when a user sends a message  into a specific slack channel, and if 30 minutes or more had passed since the last message appeared in that channel, send a slack card with a specific notice to the channel." This ended up being much more complicated than I thought using the `callback_message` method due to the different message subtypes coming through the streaming API to a channel at random times not initiated by a human in any way I can see.

This passes the original slack event object to the message via the extras dict. My current use case only requires the message subtype. Another option could be to create a `SlackMessage` with a subtype property that subclasses `Message`.

Thanks!

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -485,7 +485,11 @@ class SlackBackend(ErrBot):
 
         msg = Message(
             text,
-            extras={'attachments': event.get('attachments')})
+            extras={
+                'attachments': event.get('attachments'),
+                'slack_event': event,
+            },
+        )
 
         if channel.startswith('D'):
             if subtype == "bot_message":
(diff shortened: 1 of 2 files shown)


zoni:
I can definitely see the use of this so I'm happy to accept this. Any objections @gbin?

gbin:
Maybe we can remove the other one which becomes redundant ?

bshelton229:
I wouldn't mind `attachments` being removed from the top-level of extras, I just didn't want to break the current API in case anyone was relying on it.

Right now I only need `subtype` from the event so I could also live with only adding `subtype` to the top level of extras.

gbin:
indeed, let's not break the backward compatibility for not much.

Thanks.

Issue #949, opened 2017-01-23
Slack dm to errbot causes error


rkrambovitis:
### I am...

* [x] Reporting a bug
* [ ] Suggesting a new feature
* [ ] Requesting help with running my bot
* [ ] Requesting help writing plugins
* [ ] Here about something else

### I am running...

* Errbot version: latest - git
* OS version: Ubuntu 16.04
* Python version: 3.5
* Using a virtual environment: yes (docker)

### Issue description

When issuing command via DM, an exception is thrown.

```
ERROR    errbot.backends.slack     message event handler raised an exception
Traceback (most recent call last):
  File "/err/virtualenv/lib/python3.5/site-packages/errbot/backends/slack.py", line 417, in _dispatch_slack_message
    event_handler(message)
  File "/err/virtualenv/lib/python3.5/site-packages/errbot/backends/slack.py", line 499, in _message_event_handler
    msg.to = SlackPerson(self.sc, self.username_to_userid(self.sc.server.username),
  File "/err/virtualenv/lib/python3.5/site-packages/errbot/backends/slack.py", line 547, in username_to_userid
    raise UserDoesNotExistError("Cannot find user %s" % name)
errbot.backends.base.UserDoesNotExistError: Cannot find user errbot
```

### Steps to reproduce

Use Slack, issue command to errbot using DM.

zoni:
I'm unable to reproduce this. Is this happening for all users in your organization or just one person? Are you able to supply `DEBUG`-level log messages leading up to the exception?

rkrambovitis:
* This affects all users (it was originally reported to me by a user).

The debug log doesn't really add much:
```
16:15:34 DEBUG    errbot.backends.slack     Processing slack event: {'team': 'xxxxxxxxx', 'user': 'Uxxxxxxxx', 'text': '#cds ops', 'channel': 'xxxxxxxxx', 'type': 'message', 'ts': '1485180934.000058'}
16:15:34 DEBUG    errbot.backends.slack     Saw an event: {'channel': 'xxxxxxxxx',
 'team': 'xxxxxxxxx',
 'text': '#cds ops',
 'ts': '1485180934.000058',
 'type': 'message',
 'user': 'Uxxxxxxxx'}
16:15:34 DEBUG    errbot.backends.slack     Escaped IDs event text: #cds ops
16:15:34 ERROR    errbot.backends.slack     message event handler raised an exception
Traceback (most recent call last):
[...]

PR #961, opened 2017-02-06
Slack: compatibility with slackclient 1.0.5

zoni:
Slackclient introduced changes to their underlying client with this release which broke user ID lookups for errbot (see #949). With these changes user ID lookups will work with this new version.

Quick testing shows this also continues to work with 1.0.2 and 1.0.4 but not 1.0.3. Earlier versions were untested.


gbin:
Thanks

PR #962, opened 2017-02-08
fix Slack links with ] in them

r3m0t:


zoni:
Thanks!

errbot/backends/slack.py
```diff
@@ -61,7 +61,7 @@
     'cyan': '#00FFFF'
 }  # Slack doesn't know its colors
 
-MARKDOWN_LINK_REGEX = re.compile(r'(?<!!)\[(?P<text>.+?)\]\((?P<uri>[a-zA-Z0-9]+?:\S+?)\)')
+MARKDOWN_LINK_REGEX = re.compile(r'(?<!!)\[(?P<text>[^\]]+?\]\((?P<uri>[a-zA-Z0-9]+?:\S+?)\)')
```
zoni: Huh, yeah, that door look weird. I overlooked that...

gbin: Was this the intention ?
```
re.compile(r'(?<!!)\[(?P<text>[^\]+?)\]\((?P<uri>[a-zA-Z0-9]+?:\S+?)\)')
```

r3m0t:
```python
MARKDOWN_LINK_REGEX = re.compile(r'(?<!!)\[(?P<text>[^\]]+?)\]\((?P<uri>[a-zA-Z0-9]+?:\S+?)\)')```

r3m0t:
I guess the tests don't run on Travis

zoni:
> I guess the tests don't run on Travis

Indeed, these weren't. The naming of the test file caused `pytest` to not pick it up. I renamed it (`tests/backend_tests/{slack_tests.py => slack_test.py}`) so that it is now being run again which should avoid this in the future.

I also fixed up the regex in master.

Issue #965, opened 2017-02-12
Get rid of threadpool


murinicanor:
### I am...
* [ ] Suggesting a new feature

### I am running...

* Errbot version: 4.3.7
* OS version: any
* Python version: any
* Using a virtual environment: yes

### Issue description

The [description of the threadpool package](https://pypi.python.org/pypi/threadpool) says: 

> This module is OBSOLETE and is only provided on PyPI to support old projects that still use it. Please DO NOT USE IT FOR NEW PROJECTS! Use modern alternatives like the multiprocessing module in the standard library or even an asynchroneous approach with asyncio.

As far as i can see, the module is only used in two places in the errbot code:
- in errbot/flow.py in the FlowExecutor class
- in errbot/core.py in ErrBot's _process_command function

maybe it would be possible, to switch to [multiprocessing](https://docs.python.org/3.5/library/multiprocessing.html) instead? this would remove one dependency

zoni:
This has been on my to-do list for quite a while now, so I'm definitively supportive of this. I just haven't gotten around to it myself yet.

If you feel adventurous, please feel free to dive into the code and open a PR to swap it out. I'll be happy to give some pointers if you get stuck along the way.

intrigeri:
Can this issue be closed, now that #968 was merged?

gbin:
Indeed Thanks

PR #968, opened 2017-02-17
Move from threadpool to multiprocessing.pool. 

murinicanor:
Threadpool is obsolete and recommends to switch to a modern alternative.

as proposed by @zoni in #965 i'm trying to switch from threadpool to multiprocessing.pool. i don't do that much python coding, so i'm not sure if `apply_async` is the correct replacement for putRequest. also i didn't find know how to run the tests. i started the bot with BOT_ASYNC = True and it worked

### DIFF
--- a/errbot/flow.py
+++ b/errbot/flow.py
@@ -2,7 +2,7 @@ import logging
 from threading import RLock
 from typing import Mapping, List, Tuple, Union, Callable, Any
 
-from threadpool import ThreadPool, WorkRequest
+from multiprocessing.pool import ThreadPool
 from yapsy.IPlugin import IPlugin
 
 from errbot import Message
@@ -363,7 +363,7 @@ class FlowExecutor(object):
         with self._lock:
             if flow not in self.in_flight:
                 self.in_flight.append(flow)
-        self._pool.putRequest(WorkRequest(self.execute, args=(flow, )))
+        self._pool.apply_async(self.execute, (flow, ))
 
     def execute(self, flow: Flow):
         """
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -7,7 +7,7 @@ import time
 import sys
 import pprint
 from functools import lru_cache
-from threadpool import WorkRequest
+from multiprocessing.pool import ThreadPool
 
 from markdown import Markdown
 from markdown.extensions.extra import ExtraExtension
@@ -667,7 +667,7 @@ class SlackBackend(ErrBot):
         stream = Stream(identifier, fsource, name, size, stream_type)
         log.debug("Requesting upload of {0} to {1} (size hint: {2}, stream type: {3})".format(name,
                   identifier.channelname, size, stream_type))
-        self.thread_pool.putRequest(WorkRequest(self._slack_upload, args=(stream,)))
+        self.thread_pool.apply_async(self._slack_upload, (stream,))
         return stream
 
     def send_card(self, card: Card):
--- a/errbot/backends/telegram_messenger.py
+++ b/errbot/backends/telegram_messenger.py
@@ -1,6 +1,6 @@
 import logging
 import sys
-from threadpool import WorkRequest
+from multiprocessing.pool import ThreadPool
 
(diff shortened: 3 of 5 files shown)


gbin:
I'd love to be proven wrong but I think it won't work and it will brake Errbot's contract with the plugins as it spawns **processes** it doesn't keep the internal state of anything.

Imagine you write a simple plugin like:
```
@botcmd
def mycmd(self, msg, args):
     self.counter += 1;
     return self.counter;

It would not increase this counter.

**Comment by zoni** [comment]
> I'd love to be proven wrong but I think it won't work and it will brake Errbot's contract with the plugins as it spawns processes it doesn't keep the internal state of anything.

I'm happy to prove you wrong. :grin: See also http://bugs.python.org/issue17140. `multiprocessing.pool.ThreadPool` *is* in fact a thread-based pool. It is sadly a hidden gem of the multiprocessing module as it has no documentation.

It would be nice if we could provide backward-compatibly by still offering the old API as well (which would simply translate the call to the new one and log a deprecation warning). @murinicanor do you think that is something you would be able to implement on your own, or would you need help with that?

File: errbot/flow.py [hunk 1]
```diff
@@ -363,7 +363,7 @@ def _enqueue_flow(self, flow):
         with self._lock:
             if flow not in self.in_flight:
                 self.in_flight.append(flow)
-        self._pool.putRequest(WorkRequest(self.execute, args=(flow, )))
+        self._pool.apply_async(self.execute, (flow, )))
```
zoni: There's one too many closing braces here.

zoni:
@murinicanor 
> also i didn't find know how to run the tests

Running `tox` (install first via `pip install tox`) should be sufficient. Tox will create virtualenvs for the python interpreters available on your system and run the testsuite in them.

This is something we need to better document.

gbin:
@zoni damned this is really confusing sorry :)

murinicanor:
> It would be nice if we could provide backward-compatibly by still offering the old API as well (which would simply translate the call to the new one and log a deprecation warning). @murinicanor do you think that is something you would be able to implement on your own, or would you need help with that?

hi, i'm not sure if i understand correctly: do you mean that there is the possibility of (external) plugins still using threadpool.put_requests (and the WorkRequest object)? So i would have to add the put_requests function to the threadpool objects and pass the arguments to apply_async. not sure how to deal with the WorkRequest object though.
But i can have a look ;)

gbin:
@murinicanor indeed if it is possible to emulate the old API on the new threadpool that would be cool. Adding deprecation warning there would be awesome too.

murinicanor:
hi,

so, i'm a bit stuck at the moment, maybe someone has an idea. in errbot/core.py, line 374 `self.thread_pool.wait()` is called, so all the queued commands are finished, before an admin command is processed.
multiprocessing.pool does only provide a way to wait for a specific command, and not for the queue to be depleted. i've also looked into asyncio and didn't find an equivalent function there.

gbin:
Ok, maybe you can use a signal or something ?

gbin:
Did you fix your problem finally ? Can we test this one ?

murinicanor:
hi,
> Did you fix your problem finally ? Can we test this one ?

i've adapted the code so that the pool is joined, which means waiting for all the worker processes to exit. then the pool is closed and i recreate it and add the admin command to it. all the tox tests are positive, but i've not yet managed to manually test it, because `pip3 install -e .` segfaults on my system and i've not yet found out what the reason.

also i've not looked into the api emulation, also because i don't really understand what the goal is. errbot didn't provide a threading api and if a plugin uses threadpool, shouldn't it import threadpool itself? i mean, providing backwards compatibility for threadpool would mean writing a wrapper around every possible threadpool call, or am wrong? (maybe you have an example where threadpool is used in a plugin)

cheers,
muri
(btw.: my motivation is that i'm working on a debian package and i didn't want to package threadpool ;))

gbin:
Closing and recreating the pool is kind of ugly but I gonna merge this and we will work on a better solution. Thanks for the contribution !

Issue #969, opened 2017-02-22
Slack: Bot message with no username causes exception


rcoup:
In order to let us help you better, please fill out the following fields as best you can:

### I am...

* [X] Reporting a bug

### I am running...

* Errbot version: 4.3.7+ (https://github.com/errbotio/errbot/commit/c95300afe6cee10065e3191c04cbfc368f085af8)
* OS version: Debian Jessie
* Python version: 3.6
* Using a virtual environment: no (Docker)

### Issue description

Exception (`message event handler raised an exception KeyError: 'username'`) from https://github.com/errbotio/errbot/blob/master/errbot/backends/slack.py#L511 because a bot message doesn't have a username. In this case it's from the Github slack integration. 

This is the log message from errbot:
```
Processing slack event:
{'text': '',
'bot_id': 'B0ABCDE1234',
'attachments': [
    {'fallback': "[team/repo] Pull request submitted: '#123 WIP: stuff' https://github.com/team/repo/pull/123 by me",
    'text': '... pr description...',
    'pretext': '[team/repo] Pull request submitted by me',
    'title': '#123 WIP: stuff',
    'id': 1,
    'title_link': 'https://github.com/team/repo/pull/123',
    'color': '6CC644',
    'mrkdwn_in': ['text', 'pretext']
    }
],
'type': 'message',
'subtype': 'bot_message',
'team': 'T0ABCDE1234',
'source_team': 'T0ABCDE1234',
'user_team': 'T0ABCDE1234',
'channel': 'C0ABCDE1234',
'ts': '1487735123.000123'}
```

There's a username set in the Slack configuration for the Github integration.

We're also seeing the same issue with other integrations (eg. Zendesk). We've been seeing since Feb 9th, and >50 times since then.

### Steps to reproduce

Receive a bot message from another Slack integration with the field missing :)

### Additional info

If you have any more information, please specify it here.

sijis:
This issue should be resolved. Could you confirm this issue does not occur in 5.0.0 or higher?

roidayan:
hi,

I hit a related issue where the bot_id doesn't exists.
seems the fix should be the same.
I'm running v5.0.1

```
08:38:31 ERROR    errbot.backends.slack     message event handler raised an exception
Traceback (most recent call last):
  File "/labhome/roid/env01/lib/python3.5/site-packages/errbot/backends/slack.py", line 418, in _dispatch_slack_message
    event_handler(message)
[...]

PR #971, opened 2017-03-02
fix(slack): undefined key ‘username’ when a bot doesn’t have a username

oantonelli:
This fixes the issue https://github.com/errbotio/errbot/issues/969


gbin:
thx

PR #977, opened 2017-03-15
Add support for multiple subcommands

jlowin:
Currently, Errbot only supports a single subcommand. For example, consider these two plugin methods:

```python
def run_subcommand(...):

def run_my_subcommand(...):
```
The first can be run as "!run subcommand \<args\>" but the second must be "!run my_subcommand \<args\>" because only the first "_" is considered when parsing subcommands. This PR allows arbitrarily long subcommands, as long as they match a plugin command. Users could write "!run my subcommand \<args\>" successfully.

Includes a unit test.

### DIFF
--- a/tests/commands_test.py
+++ b/tests/commands_test.py
@@ -335,3 +335,18 @@ def test_callback_no_command(testbot):
     testbot.bot.plugin_manager.update_plugin_places([], extra_plugin_dir)
     testbot.exec_command('!plugin activate TestCommandNotFoundFilter')
     assert expected_str == testbot.exec_command(cmd)
+
+
+def test_subcommands(testbot):
+    # test single subcommand (method is run_subcommands())
+    cmd = '!run subcommands with these args'
+    cmd_underscore = '!run_subcommands with these args'
+    expected_args = 'with these args'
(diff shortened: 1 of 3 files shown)


jlowin:
This should also solve some confusion because "!help <plugin>" shows all commands with spaces, not underscores.

tests/commands_test.py
```diff
@@ -335,3 +335,14 @@ def test_callback_no_command(testbot):
...
+
+def test_subcommands(testbot):
+    # test single subcommand (method is run_subcommands())
+    cmd = '!run subcommands with these args'
+    expected_args = 'with these args'
+    assert expected_args == testbot.exec_command(cmd)
+
+    # test multiple subcmomands (method is run_lots_of_subcommands())
+    cmd = '!run lots of subcommands with these args'
+    assert expected_args == testbot.exec_command(cmd)
```
zoni: Could you also test the same two things, only with underscores, as well? Then we have certainty that `!run_subcommands` and `!run_lots_of_subcommands` also continues to work.

jlowin:
@zoni good call on testing the underscores -- just pushed the changes

zoni:
Very sweet, thanks!

PR #981, opened 2017-03-22
Enable slack formatting for cards

daniloqueiroz:
Enables slack cards formatting


sijis:
This is a good thing, especially if folks want to format the output.
I'm not seeing an issue with this PR. 👍

gbin:
The problem is that we would then just support the *Slack* markdown, and not the markdown supported by Errbot. It would be better to get the input as errbot's markdown and convert it to the Slack one (for whatever it supports)

sijis:
I will double check but isn't sending a card a slack only thing anyway?

If not, then I agree with your argument. Having a single markdown that we translate based on backed is the correct approach.

PR #985, opened 2017-04-03
Slack reactions and Message.extras['url']

r3m0t:
This is #917 again.

On Slack a useful alternative to an "OK" reply is a reaction. We also use it to indicate the result of long-running operations and give immediate ⌛️  feedback.


gbin:
Thanks !

PR #995, opened 2017-04-19
Make hipchat room string representation return the jabber-id

barakschiller:
This is needed to allow usage of a ```HipChatRoom``` in the ```from``` field of messages


gbin:
Thanks for the PR !

Can you check if it was failing with https://github.com/errbotio/err-backend-tester before and fixed after your  PR ? It is supposed to be in the contract of the backend.

gbin:
No activity on this please reopen if this is still an issue and test with the backend-tester

PR #999, opened 2017-05-04
Allow SUPPRESS_CMD_NOT_FOUND=True without crashing

rtxm:
Fixed a typo that made the bot crash when the configuration option SUPPRESS_CMD_NOT_FOUND was set to True.


zoni:
Ouch. Thanks!

Since that line was being edited, I immediately took advantage to fix up a spelling error as well. :)

PR #1005, opened 2017-05-09
callback all unknown commands filters.

gbin:
This changes the behavior of the unknown command filters.
As you can disable the core unknown command plugin, it is better to callback all those and let the user choose which one should be active.


gbin:
CI passing, merging it.

PR #1006, opened 2017-05-10
Add support for use on Slack Enterprise Grid

japz:
Slack introduced an extra identifier to identify users on the Slack
Enterprise Grid. These user IDs start with a W instead of U or B.

See also: https://api.slack.com/types/user


japz:
How do I retrigger the build? This looks like a transient error to me and not a problem with the commit.

gbin:
thx !

Issue #979, opened 2017-03-16
Show user where they are in a flow


ae-dg:
### I am...

* [x] Suggesting a new feature

### I am running...

* Errbot version: 4.3.7
* OS version: Linux 3.16.0-4-amd64 (Debian)
* Python version: 3.4.2
* Using a virtual environment: yes

### Request

When in a flow the `flows list` command shows a list of currently set-up flows but not what steps are available next or what step the user is on.  It would be nice to have a command that shows the "You are in the flow..., you can continue with..." in case they get lost in a flow (the same sort of message they get when a they go into a step).

zoni:
I don't believe #1011 quite covers what is asked here. Note that the `!flows status` command is admin only and displays an overview of active flows for all users, whereas the request in this issue was about showing users themselves some extra information.

aliegha:
I wonder if we could rework this so that if an admin runs the command they
get all statuses but if a regular user runs a command it would allow them
to just return theirs ?

PR #1011, opened 2017-05-23
Updating flows Status to return better information issue 979

aliegha:
This resolves https://github.com/errbotio/errbot/issues/979 : 

Sample response:

image

image

This handles both flows where there are both single and multiple next steps


gbin:
Thanks.

PR #1013, opened 2017-05-25
utf open changes.rst

rkrambovitis:
due to the é character in CHANGES.rst which causes compile time error in my environment.

```
Collecting https://github.com/errbotio/errbot/archive/master.zip
  Downloading https://github.com/errbotio/errbot/archive/master.zip (1.4MB)
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/tmp/pip-8y2onk3m-build/setup.py", line 81, in <module>
        changes = read('CHANGES.rst')
      File "/tmp/pip-8y2onk3m-build/setup.py", line 72, in read
        return open(os.path.join(os.path.dirname(__file__), fname)).read()
      File "/usr/lib/python3.5/encodings/ascii.py", line 26, in decode
        return codecs.ascii_decode(input, self.errors)[0]
    UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 139: ordinal not in range(128)
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -78,7 +78,9 @@ if __name__ == "__main__":
 
     args = set(sys.argv)
 
-    changes = open('CHANGES.rst', 'r', encoding='utf-8').read()
+    changes_file = open('CHANGES.rst', 'r', encoding='utf-8')
+    changes = changes_file.read()
+    changes_file.close()
 
     if changes.find(VERSION) == -1:
         raise Exception('You forgot to put a release note in CHANGES.rst ?!')
--- a/setup.py
+++ b/setup.py
@@ -69,7 +69,7 @@ def read_version():
 
 
 def read(fname, encoding='ascii'):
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -78,7 +78,9 @@ def read(fname):
 
     args = set(sys.argv)
 
-    changes = read('CHANGES.rst')
+    changes_file = open('CHANGES.rst', 'r', encoding='utf-8')
```
gbin: can you use the syntax:
``` 
with open('CHANGES.rst', 'r', encoding='utf-8') as changes_file:
      changes = changes_file.read()
```

rkrambovitis:
I think that's done it.
My python foo it not that great :)

gbin:
Thx

rkrambovitis:
Many thanks :)

PR #1015, opened 2017-05-25
docs/testing.rst: Highligh assertCommand assertion function

meetmangukiya:


meetmangukiya:
@gbin @zoni I think we need to modify this function to actually perform an assert as the function name suggests. Because I noticed that it raises an exception and not perform an actual assert.

In testing, we usually require assertions to even occur in the test results, right?

gbin:
thx

PR #1017, opened 2017-05-25
backends/test: use assert instead of raising an exception

meetmangukiya:
Closes https://github.com/errbotio/errbot/issues/1016


gbin:
Thx

Issue #1019, opened 2017-05-26
Incorrect docstring for TestBot.assertCommandFound


meetmangukiya:
the docstring states 
> """Assert the given command does not exist"""

however, the assert statement verifies if the command exist and so does the function name suggest
> `assert 'not found' not in self.bot.pop_message(timeout)`

https://github.com/errbotio/errbot/blob/master/errbot/backends/test.py#L460

PR #1020, opened 2017-05-26
backends/test: Correct docstring of TestBot.assertCommandFound

meetmangukiya:
Closes https://github.com/errbotio/errbot/issues/1019


zoni:
Good catch, thanks.

PR #1023, opened 2017-05-30
Show version when using git version

SvenNederhoff:
When you are using Errbot by a git checkout the version will always be `9.9.9` when you type `!about`

With this PR we return the latest tag and add `GIT CHECKOUT` to clarify that it is a git checkout.

preferably I want to use git but I don't know the thoughts about an extra python library in the core files only for those who want the git checkout. 

This version is a sort of idea. feel free to edit/improve

### DIFF
--- a/errbot/core_plugins/help.py
+++ b/errbot/core_plugins/help.py
@@ -1,4 +1,5 @@
 import textwrap
+import subprocess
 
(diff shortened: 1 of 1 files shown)


gbin:
thx

PR #1034, opened 2017-06-11
Slack: support for simple threaded responses.

gbin:
By default, if the bot responds from a command in a thread, it will send
its response to the same thread.


gbin:
I will add in another pull request a mean to start a thread with 'self.send' in the API.

gbin:
I have been testing this one for the past few days, it works quite well.

PR #1036, opened 2017-06-14
Show the response and message in the assertion error

meetmangukiya:


errbot/backends/test.py
```diff
@@ -454,7 +454,10 @@ def assertCommand(self, command, response, timeout=5):
         """Assert the given command returns the given response"""
         self.bot.push_message(command)
         msg = self.bot.pop_message(timeout)
-        assert response in msg
+        try:
+            assert response in msg
+        except AssertionError:
+            raise AssertionError("'{}' not in '{}'".format(response, msg))
```
zoni: As this is just going to catch the exception raised by the assert statement, it's more clear and concise to write this simply as:

```python
if response not in msg:
    raise AssertionError("'{}' not in '{}'".format(response, msg))
```

Alternatively, the following should also work (I think I like this best actually):

```python
assert response in msg, "'{}' not in '{}'".format(response, msg)
```

meetmangukiya: yeah, second one is much better :+1:

PR #1049, opened 2017-07-01
FIX: webhooks decorators with no uri but kwargs fail.

### LINKED ISSUE #1043: Cannot initialize webhook decorator with only raw=True parameter

### I am...

* [x] Reporting a bug
* [ ] Requesting help with running my bot
* [ ] Here about something else

### I am running...

* Errbot version: 5.0.1
* OS version: macOS Sierra Version 10.12.5
* Python version: Python 3.6.1
* Using a virtual environment: yes

### Issue description

When setting a webhook as `@webhook(raw=True)` I notice that `*args` is passed to the webhook decorator as an empty tuple, which makes the following line fail with and IndexError exception (errbot/_init.py, L441): `if isinstance(args[0], str): # first param is uri_rule.`

This is the example on the documentarion: http://errbot.io/en/latest/user_guide/plugin_development/webhooks.html#the-raw-request

### Steps to reproduce

1. Create a webhook, defining only the raw argument: `@webhook(raw=True)`. 
2. Use the webhook. It will raise the IndexError exception described above.

gbin:
This fixes #1043

### DIFF
--- a/errbot/__init__.py
+++ b/errbot/__init__.py
@@ -400,6 +400,8 @@ def _tag_webhook(func, uri_rule, methods, form_param, raw):
     func._err_webhook_raw = raw
     return func
 
+def _uri_from_func(func):
+    return r'/' + func.__name__
 
 def webhook(*args,
             methods: Tuple[str]=('POST', 'GET'),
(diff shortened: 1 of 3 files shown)


errbot/__init__.py
```diff
@@ -438,15 +440,22 @@ def a_webhook(self, payload):
...
                                          methods=methods,
                                          form_param=form_param,
                                          raw=raw)
-    return _tag_webhook(args[0],
-                        r'/' + args[0].__name__,
+    return _tag_webhook(args[0],  # naked decorator so the first parameter is a function.
+                        _uri_from_func(args[0]),
                         methods=methods,
                         form_param=form_param,
                         raw=raw)
```
zoni: Lots of duplication going on here (only the first 1 or 2 arguments differ each time). Can we avoid this somehow?

gbin: we can use kwargs but then it will make the decorator less explicit. Otherwise I don't see how... any idea ?

zoni: Yeah, fair point. i'm don't see a nice way either.

PR #1053, opened 2017-07-03
add: Function `delay_call` for `BotPlugin`

mrshu:
* Add function `delay_call` which delays the execution of a provided
  method for a given number of seconds to `BotPlugin`

* Add a bit of documentation on the newly added `delay_call` function.

### DIFF
--- a/docs/user_guide/plugin_development/scheduling.rst
+++ b/docs/user_guide/plugin_development/scheduling.rst
@@ -24,3 +24,22 @@ minute when your plugin gets activated:
         def activate(self):
             super().activate()
             self.start_poller(60, self.my_callback)
+
+Delaying a function call
+------------------------
+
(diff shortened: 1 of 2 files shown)


mrshu:
@gbin as I mentioned on Gitter, I did miss this function in err's `BotPlugin`. However, the actual implementation is basically a oneliner, so it may not make sense to pollute the `BotPlugin`'s API with a function like this.

The final decision is yours =)

gbin:
You could implement it with an optional `times` parameter in start_poller default None (as infinite number of times), if you put 1, it is basically what you are doing here.
What do you think ?

mrshu:
@gbin makes a lot of sense, thanks for the suggestion!

I did that in #1059 and this PR can therefore be closed.

PR #1055, opened 2017-07-04
Simple command autocomplete in text mode.

gbin:
Press [TAB] to complete any command.

### DIFF
--- a/errbot/backends/text.py
+++ b/errbot/backends/text.py
@@ -226,14 +226,6 @@ class TextBackend(ErrBot):
         super().__init__(config)
         log.debug("Text Backend Init.")
 
-        try:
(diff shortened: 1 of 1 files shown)


errbot/backends/text.py
```diff
@@ -113,6 +105,24 @@ def __init__(self, config):
...
+                    return options[state]
+
+            readline.parse_and_bind("tab: complete")
+            readline.set_completer(completer)
+
+        except ImportError:
+            # Readline is Unix-only
+            log.debug("Python readline module is not available")
+            pass
+
```
zoni: Can we move this block into its own function so it doesn't pollute `serve_forever()` so much?

gbin: done

PR #1057, opened 2017-07-04
This adds threading support officially.

gbin:
It is only implemented on Slack at the moment.

- adds a parent field to Message signaling to backends that it is a threaded
response.
- retrofit in_reply_to as a signal to thread by putting this message as
parent without changing the frm/to behavior.
- adds an optional threaded boolean for the build_reply & co that needs
to be implemented by the backends (similar category as private).

### DIFF
--- a/errbot/backends/base.py
+++ b/errbot/backends/base.py
@@ -227,28 +227,31 @@ class Message(object):
     A chat message.
 
     This class represents chat messages that are sent or received by
-    the bot. It is modeled after XMPP messages so not all methods
-    make sense in the context of other back-ends.
+    the bot.
     """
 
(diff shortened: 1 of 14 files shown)


sijis:
I'll look at this more closely over the weekend. I'm not 100% familiar with the threading model, however, if you can show a couple of example usages that would be great!

gbin:
@sijis It allows you to start message threads like this:
image

sijis:
Ahhh! I was thinking python threads.
I like that this feature being added. Again I'll try and poke at this and provide any feedback.

gbin:
I timeouted on this, merged it.

PR #1059, opened 2017-07-05
add: `times` parameter to `start_poller`

mrshu:
* Add a `times` parameter to `start_poller` which denotes how many times
  a method should be polled.

* Add a bit of documentation on the new `times` parameter.

### DIFF
--- a/errbot/botplugin.py
+++ b/errbot/botplugin.py
@@ -201,12 +201,16 @@ class BotPluginBase(StoreMixin):
     def start_poller(self,
                      interval: float,
                      method: Callable[..., None],
+                     times: int=None,
                      args: Tuple=None,
                      kwargs: Mapping=None):
         """ Starts a poller that will be called at a regular interval
 
         :param interval: interval in seconds
         :param method: targetted method
+        :param times:
+            number of times polling should happen (defaults to``None`` which
+            causes the polling to happen indefinitely)
         :param args: args for the targetted method
         :param kwargs: kwargs for the targetting method
         """
@@ -221,12 +225,13 @@ class BotPluginBase(StoreMixin):
         # noinspection PyBroadException
         try:
             self.current_pollers.append((method, args, kwargs))
-            self.program_next_poll(interval, method, args, kwargs)
+            self.program_next_poll(interval, method, times, args, kwargs)
         except Exception:
             log.exception('failed')
 
     def stop_poller(self,
                     method: Callable[..., None],
+                    times: int=None,
                     args: Tuple=None,
                     kwargs: Mapping=None):
         if not kwargs:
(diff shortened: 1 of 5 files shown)


errbot/botplugin.py
```diff
@@ -239,10 +244,15 @@ def stop_poller(self,
     def program_next_poll(self,
                           interval: float,
                           method: Callable[..., None],
+                          times: int=None,
                           args: Tuple=None,
                           kwargs: Mapping=None):
+        if type(times) in [int, float] and times <= 0:
```
gbin: don't check the type, if it is a string there between crashing and failing silently, it is way better that it crashes ! :)

just: 
```python
if times is not None and times <= 0:
```

errbot/botplugin.py
```diff
@@ -264,7 +275,11 @@ def poller(self,
                 method(*args, **kwargs)
             except Exception:
                 log.exception('A poller crashed')
-            self.program_next_poll(interval, method, args, kwargs)
+
+            if type(times) in [int, float]:
```
gbin:
```python
if times is not None:
```

errbot/botplugin.py
zoni: An assertion makes a lot of sense for such things. Though since we have type hinting, I'd argue even that is unnecessary now as `mypy` could be used to verify arguments of the correct type are provided.

mrshu: Thanks @zoni and @gbin, makes a lot of sense!

mrshu:
@gbin @zoni I tried to setup a test for this, but It was not really successful. I admit that I was not able to run the tests locally, so I am not sure what exactly is wrong and how to put together a test for something like this.

Any help would be greatly appreciated =)

mrshu:
@gbin @zoni after some experimentation I've managed to get the tests to pass. Please let me know if you find the current implementation acceptable.

Thanks!

tests/poller_plugin/poller_plugin.py
```diff
@@ -0,0 +1,15 @@
...
+
+class PollerPlugin(BotPlugin):
+
+    def delayed_hello(self, frm):
+        self.send(frm, 'Hello world! was sent 5 seconds ago')
+
+    @botcmd
+    def hello(self, msg, args):
+        """Say hello to the world."""
+        self.start_poller(5.0, self.delayed_hello, times=1,
```
gbin: put that ridiculously low: 0.1

tests/poller_test.py
```diff
@@ -0,0 +1,12 @@
+from os import path
+import time
+
+CURRENT_FILE_DIR = path.dirname(path.realpath(__file__))
+extra_plugin_dir = path.join(CURRENT_FILE_DIR, 'poller_plugin')
+
+
+def test_delayed_hello(testbot):
+    assert 'Hello, world!' in testbot.exec_command('!hello')
+    time.sleep(5)
```
gbin: sleep 1 s

tests/poller_test.py
```diff
@@ -0,0 +1,12 @@
...
+    delayed_msg = 'Hello world! was sent 5 seconds ago'
+    assert delayed_msg in testbot.pop_message(timeout=1)
```
gbin: and check if only 1 message has been enqueued.

mrshu:
@gbin Thanks for the suggestions! 

If they indeed were the last improvements, this should be ready for the final review.

gbin:
Thank you !

PR #1062, opened 2017-07-08
Makes hinting on flows useable

gbin:
It needs to be attached to a specific node so it will be set at
initialization phase after the fact with a 'hints' parameter.


gbin:
@zoni a quick look ? Thanks !

gbin:
Thx @zoni

PR #1073, opened 2017-08-03
slack: Fix possible event without bot_id

### LINKED ISSUE #969: Slack: Bot message with no username causes exception

roidayan:
Fixes #969

Signed-off-by: Roi Dayan <roid@mellanox.com>

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -496,7 +496,7 @@ class SlackBackend(ErrBot):
             if subtype == "bot_message":
                 msg.frm = SlackBot(
                     self.sc,
-                    bot_id=event['bot_id'],
+                    bot_id=event.get('bot_id', ''),
                     bot_username=event.get('username', '')
                 )
             else:
@@ -508,7 +508,7 @@ class SlackBackend(ErrBot):
             if subtype == "bot_message":
                 msg.frm = SlackRoomBot(
                     self.sc,
-                    bot_id=event['bot_id'],
+                    bot_id=event.get('bot_id', ''),
                     bot_username=event.get('username', ''),
                     channelid=event['channel'],
                     bot=self


zoni:
This will certainly avoid the KeyError exception, but I wonder how the bot actually behaves with such a message where the `bot_id` is now an empty string. I'm worried it may cause a different error further down the line.

Do you have any info on this?

roidayan:
I dont info. It was reproduced while playing with plugins. I'll try to get the exact step to reproduce this and see how it goes..

roidayan:
I couldn't reproduce the exception.
Also according to here https://api.slack.com/events/message/bot_message
it doesn't say if it's possible bot_id is missing unlike username for example.

i'll close this pull request and reopen it will ever reproduce and i'll have a way to reproduce it again.

roidayan:
from the bot log the issue does reproduce. 

2017-09-19 05:01:56,873 ERROR    errbot.backends.slack     message event handler raised an exception
  File "/root/venv2/lib/python3.6/site-packages/errbot/backends/slack.py", line 422, in _dispatch_slack_message
    event_handler(message)
  File "/root/venv2/lib/python3.6/site-packages/errbot/backends/slack.py", line 503, in _message_event_handler
    bot_id=event['bot_id'],
KeyError: 'bot_id'
2017-09-19 05:01:56,987 ERROR    errbot.backends.slack     message event handler raised an exception
  File "/root/venv2/lib/python3.6/site-packages/errbot/backends/slack.py", line 422, in _dispatch_slack_message
    event_handler(message)
  File "/root/venv2/lib/python3.6/site-packages/errbot/backends/slack.py", line 503, in _message_event_handler
    bot_id=event['bot_id'],
KeyError: 'bot_id'


I still don't have definite answer for what can happen with empty bot_id.
i'll try to test this.

roidayan:
this issue could have potentially hit before, in line ~477 at this block:
```
        if 'message' in event:
            text = event['message'].get('text', '')
            user = event['message'].get('user', event.get('bot_id'))
        else:
            text = event.get('text', '')
            user = event.get('user', event.get('bot_id'))
```
but here we get bot_id using get method which if key doesn't exists by default returns None.
so user is is None in this case.

I looked in the commit that added this code (SlackBot msgs)
0abc248 Slack: Process messages from webhook users

it looks like before adding SlackBot msg we only used user which can be the bot_id or None.
-            msg.frm = SlackPerson(self.sc, user, event['channel'])

so if everything was ok with None before it should be ok now as well. it just uses another
layer SlackBot which inherits from SlackPerson and passing botid as user to SlackPerson.

I'll run with this change for few days and see if any new issue pops up.

roidayan:
The issue seems to reproduce when the bot sends private msg to me. I noticed the issue shows in the log at the same times I get private msg from the bot about a new version (VersionChecker plugin)

PR #1075, opened 2017-08-04
Support Hipchat server JIDs in callback_message

sivy:
When the bot receives a command in a Hipchat Server room, the room name is returned as a JID, and during while processing ACLs for `callback_message` (using Hypchat) cannot find the room via the hipchat API (`v2/room/1_room@conf.btf.hipchat.com NOT FOUND`).

This introduces two changes:

- changes `callback_room` to use `query_room` to check if the room passed was a JID
- changes `query_room` to check for the Server-style JID.


errbot/core_plugins/chatRoom.py
```diff
@@ -235,6 +235,8 @@ def callback_message(self, msg):
             elif msg.is_group:
                 fr = msg.frm
                 chat_room = str(fr.room)
+                # Hipchat Server, this might be a JID, look it up just in case
+                chat_room = self.query_room(chat_room)
```
gbin: It looks super hackish, what is really going on ? why fr.room is not a room from the hipchat backend ? (the bug is in the backend, not core I guess ?)

kincl:
Just adding the check for `@conf.btf.hipchat.com` was enough for me to get past this bug.

gbin:
ok, can you fix the PR to only include the check for @conf.btf.hipchat.com so I can merge it? Thanks

kincl:
Thats going to have to be @sivy

anttipalsola:
For me just including the check for @conf.btf.hipchat.com fixed the problem. I also made another pull request  (#1121) including only that change. (Sorry, I should have checked this pull request first...) Anyway without this fix Errbot is basicly unusable with HipChat Server.

zoni:
I just merged #1121 which seems to resolve this, according to reports by @anttipalsola and @kincl. Please open a new issue and/or PR if that change was insufficient.

PR #1083, opened 2017-08-15
slack: Add markdown support for text and pretext

### LINKED ISSUE #839: Slack: send_card() does not support markdown
Using [`send_card()`](http://errbot.io/en/latest/_modules/errbot/botplugin.html#BotPlugin.send_card) with the Slack backend does not render the given string to markdown prior to passing it to the underlying API.

For example:
```
self.send_card(                         
    in_reply_to=message,                
    body="`some text`",
)                                       
```
Results in the sent card actually containing
```
`some text`
```
rather than image in the usual Slack inline code style. 

This applies to all of the markdown syntax that Slack supports.

The same issue also exists when using the `fields` argument of `send_card()`.

roidayan:
Add markdown support for text and pretext in slack cards.
Fixes #839

Signed-off-by: Roi Dayan <roi.dayan@gmail.com>

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -721,6 +721,8 @@ class SlackBackend(ErrBot):
         if card.fields:
             attachment['fields'] = [{'title': key, 'value': value, 'short': True} for key, value in card.fields]
 
+        attachment['mrkdwn_in'] = ['text', 'pretext']
+
         data = {
             'text': ' ',
             'channel': to_channel_id,


roidayan:
not sure why the travis failure

zoni:
Tests have been flaky again lately. I restarted the offending job so it's green across the board.

@gbin do we feel okay with just rending the card body as markdown by default? I think that's what we want, right?

gbin:
It is probably confusing for the user: it is markdown as *slack* markdown not *errbot* markdown if I understand correctly. We need to add a conversion in the middle like we do for normal messages no?

zoni:
> It is probably confusing for the user: it is markdown as slack markdown not errbot markdown

In this specific instance, I'm not sure. Slack has a dialect for *message formatting* that's not markdown (but somewhat markdown inspired) which is why we absolutely need to conversion there.

For cards, I think they do actual markdown format on those. You're right that's probably vanilla markdown though, not the extended markdown support errbot understands.

Extra datapoint: Our Gitter backend also converts from our markdown format to regular markdown when sending messages (Gitter supports markdown natively).

You're probably right in that we should convert explicitly for cards too. That'll complicate the code though. @roidayan would you feel up for that challenge?

roidayan:
sure. just to clarify.. you want me to do as in Gitter backend, to convert from errbot markdown to regular markdown the text and pretext fields? so the errbot extended markdown will be supported in plugins?

gbin:
correct

roidayan:
so I looked in the code and it seems also send_message strips markdown from the messages.
verified with echo plugin ```!echo *text*```

the current code seems to have duplication:
```self.md = slack_markdown_converter(compact)```
and slack_markdown_converter() is actually duplicated code of imtext().
we can delete slack_markdown_converter() and do this instead:
```
from errbot.rendering import imtext
self.md = imtext()
```

anyway it strips also *bold* so something is wrong with it.

I did what Gitter backend does and just use md as is:
```
from errbot.rendering import md
...
self.md = md()
```

and in send_card added md.convert calls like in send_message
```attachment['pretext'] = self.md.convert(card.summary)```
```attachment['text'] = self.md.convert(card.body)```

and now messages and cards works as should with markdown.
I haven't try what will happen with markdown not supported by Slack. 
I guess there msg will contain those special characters but users of Slack will just not use this markdown?

gbin:
can you rebase and retest? Thanks.

roidayan:
sure. I currently have an issue with my bot not connecting and not sure why. when it is resolved i'll be able to test.

gbin:
If you have some time to rebase retest, please do and reopen thanks!

PR #1085, opened 2017-08-17
Fixes HipChat user ACLs when in a room

kincl:
When in a room, the ACL plugin tries to compare ACLs against the MUC
from JID which does not directly correlate to the actual JID of the user.
Here I am using the aclattr property to set the user's JID which the
ACL plugin will use to match against.

### DIFF
--- a/errbot/backends/hipchat.py
+++ b/errbot/backends/hipchat.py
@@ -103,6 +103,7 @@ class HipChatRoomOccupant(XMPPRoomOccupant):
     def aclattr(self):
         return self._aclattr
 
+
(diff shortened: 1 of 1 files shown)


kincl:
Hey @gbin when you get a chance can you review this PR? Thanks!

errbot/backends/hipchat.py
```diff
@@ -404,6 +411,14 @@ def create_connection(self):
             verify=self.api_verify,
         )
 
+    def _build_room_occupant(self, txtrep):
```
gbin: Where do you call this one?

PR #1094, opened 2017-09-01
Remove smart quotes for argument parsing

jasedit:
Slack replaces all quotation marks with smart quotation marks, which interferes with parsing commands via the argparse interface. This PR fixes this by stripping smart quotation marks from the arguments.


jasedit:
I had started to write this up using regex, but wasn't sure if it would be considered too significant a change for two extra character types. If you'd prefer a regex version, I can put together a quick PR for comparison/consideration.

zoni:
I'd certainly be interested in seeing it for comparison. If you wouldn't mind coming up with a PR for that, that'd be great.

PR #1110, opened 2017-10-02
Make all config directory strings raw strings.

derAnfaenger:
This fixes issues on Windows, where backslashes
are used in directory paths.

Specifically, I received this error when trynig to run errbot on windows:
```
C:\Users\Robert\errbot>errbot
Starting errbot with a default system encoding other than 'utf-8' might cause you a heap of troubles. Your current encoding
 is set at 'cp1252'
I could not import your config from C:\Users\Robert\errbot\config.py, please check the error below...
  File "c:\users\robert\appdata\local\programs\python\python35-32\lib\site-packages\errbot\cli.py", line 77, in get_config
    config = __import__(path.splitext(path.basename(config_fullpath))[0])
  File "C:\Users\Robert\errbot\config.py", line 11
    BOT_EXTRA_PLUGIN_DIR = 'C:\Users\Robert\errbot\plugins'
                          ^
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 2-3: truncated \UXXXXXXXX escape
```

### DIFF
--- a/errbot/templates/initdir/config.py.tmpl
+++ b/errbot/templates/initdir/config.py.tmpl
@@ -8,7 +8,7 @@ import logging
 BACKEND = 'Text'  # Errbot will start in text mode (console only mode) and will answer commands from there.
 
 BOT_DATA_DIR = r'{{ data_dir }}'
-BOT_EXTRA_PLUGIN_DIR = '{{ extra_plugin_dir }}'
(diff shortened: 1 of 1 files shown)


zoni:
Windows obviously doesn't receive any testing from us (:sweat_smile:) so I'm glad you took the time to dive into the cause and open a PR. Thanks!

derAnfaenger:
Thanks for merging! I'm a little late on the Zulip backend, but started working on it  :)

PR #1111, opened 2017-10-03
fix(repo_manager): Use utc timestamps (#1)

Jake-Shadle:
In a fresh instance of errbot with no previous storage data, the `LAST_UPDATE` time is set to 0 to indicate it's never been checked. However, when you search or install a new repo, it checks to see if should do an update, but since it was using `datetime.fromtimestamp` it would fail (at least on Windows, this might be OS specific) since 0 is an invalid value. Now we just call `self.index_update()` when we don't have any previous state to avoid the issue.

### DIFF
--- a/errbot/repo_manager.py
+++ b/errbot/repo_manager.py
@@ -106,12 +106,12 @@ class BotRepoManager(StoreMixin):
             log.info('No repo index, creating it.')
             self[REPO_INDEX] = {LAST_UPDATE: 0}
 
-        if datetime.fromtimestamp(self[REPO_INDEX][LAST_UPDATE]) < datetime.now() - REPO_INDEXES_CHECK_INTERVAL:
+        if datetime.utcfromtimestamp(self[REPO_INDEX][LAST_UPDATE]) < datetime.utcnow() - REPO_INDEXES_CHECK_INTERVAL:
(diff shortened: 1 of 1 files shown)


errbot/repo_manager.py
```diff
@@ -106,12 +106,12 @@ def check_for_index_update(self):
             log.info('No repo index, creating it.')
             self[REPO_INDEX] = {LAST_UPDATE: 0}
```
zoni: I think it might be better to directly call `self.index_update()` here instead of relying on the calculation below. That should also fix the issue you are experiencing.

Jake-Shadle:
Ahh yes, that is a much simpler fix, thanks!

errbot/repo_manager.py
```diff
@@ -104,7 +104,8 @@ def shutdown(self):
     def check_for_index_update(self):
         if REPO_INDEX not in self:
             log.info('No repo index, creating it.')
-            self[REPO_INDEX] = {LAST_UPDATE: 0}
+            self.index_update()
+            return
```
zoni: I added an early return here, otherwise it could happen that `self.index_update()` fails and causes the check below to fire it again. Otherwise, looks good. I'll merge this in, thanks!

PR #1112, opened 2017-10-03
slack: new get_user_profile function for slack backend

alex-p-uk:
I could not figure out how to get slack user's email for integration with another service, so created this small function to fetch user profile with users.info API call. Not sure if it's the best or right way, but hoping to get something workable in the end to avoid keeping local patches.

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -601,7 +601,7 @@ class SlackBackend(ErrBot):
                 return None
             else:
                 raise e
-                
+
     def get_user_profile(self, userid):
         """Get slack user profile"""
         try:
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -601,6 +601,20 @@ class SlackBackend(ErrBot):
                 return None
             else:
                 raise e
+                
+    def get_user_profile(self, userid):
+        """Get slack user profile"""
+        try:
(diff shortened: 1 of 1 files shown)


andrewthetechie:
@gbin @zoni any chance this could get added to a release in the near future?

sijis:
I'm not seeing an issue with this. It does seem like a good helper method.

zoni:
After giving this quite some thought, I've come to the conclusion I'm not in favor of adding this into errbot directly. While I absolutely understand the convenience of having it in errbot core, it also adds additional burden on us as maintainers as people expect it to work and keep on working if Slack were to change something in their API. Based on the this I'm of the opinion that we're currently too low on resources to support such utility functions.

That said, I think some community-contributed docs/guides on doing these types of backend-specific things would be tremendously valuable and I would happily work with anyone who'd like to work on such a write-up.

zoni:
(I'm closing the MR to prevent it from lingering, but feel free to continue discussion on this)

zoni:
One more thing. Utility functions like these could easily go into their own class and be distributed independently. You could then use these as a mixin in your own plugins, which avoids the need for local patches and/or duplication across multiple plugins. A very good example of this principle is the `CrontabMixin` provided by [errcron](https://pypi.python.org/pypi/errcron).

PR #1113, opened 2017-10-04
Cache global user list for HipChat to avoid large amounts of API calls

nulfox:
Fix to resolve issue where large amount of calls to _find_user() was resulting in users() being called, which creates multiple calls to API.
After upgrading to latest version, I started hitting the throttle limit on every start while making a very large amount of repeated calls to users(). We have about 6 rooms with 10-20 participants each. 

Downside is that any modifications to users would no be picked up until the bot is restarted. Ideally, this would be time based cached that would expire, leading to periodic updates of user list. Can add this into request if desired.

### DIFF
--- a/errbot/backends/hipchat.py
+++ b/errbot/backends/hipchat.py
@@ -154,7 +154,7 @@ class HipChatRoom(Room):
         return "<HipChatMUCRoom('{}')>".format(self.name)
 
     def __str__(self):
-        return self._name
+        return self.room['xmpp_jid']
 
     def join(self, username=None, password=None):
         """
--- a/errbot/backends/hipchat.py
+++ b/errbot/backends/hipchat.py
@@ -346,6 +346,7 @@ class HipchatClient(XMPPConnection):
     def __init__(self, *args, **kwargs):
         self.token = kwargs.pop('token')
         self.endpoint = kwargs.pop('endpoint')
+        self._cached_users = None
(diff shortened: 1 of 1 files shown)


errbot/backends/hipchat.py
```diff
@@ -346,6 +346,7 @@ class HipchatClient(XMPPConnection):
     def __init__(self, *args, **kwargs):
         self.token = kwargs.pop('token')
         self.endpoint = kwargs.pop('endpoint')
+        self._cached_users = None
```
gbin: We have LRU cache we can use, there should be no need to implement it ourself, can you check why it doesn't work?
`@lru_cache(1024)`

nulfox: @gbin It does work, but not well enough to avoid a bunch of calls on start up. Reason for this is that each room participant is a unique user, even if it's same organizational user, so with 10 users per room, with 5 rooms, you would get 50 uncached calls. Each uncached call results in 2-3 API calls. users() gets all users, and since it's paged, this can (and does for us) results in multiple calls, depending on size of organization. It's not an enormous amount of calls, but large enough to trigger throttling on every start up.

barakschiller:
This issue had be bugging me for a while, thank you for taking the time to sort it out!

jereze:
Same here. Thanks!

PR #1117, opened 2017-10-10
Remove the extra spaces in the identity config template

elopio:
I'm a little pep8 obsessive, and the inconsistency in this file was driving me crazy ;)


zoni:
:smile:

PR #1124, opened 2017-10-19
Docs: Add note about shelve module restriction

asfaltboy:
When using ShelfStorage plugin (default), the user may not be aware that he needs to stop `errbot` to access it outside the bot, in the cli helper commands.


zoni:
This is an excellent warning, that's a good thing to document indeed. Thanks for making the change.

I took the liberty of making some small editorial changes for consistency with other areas of the docs, which I hope you're okay with. :smile:

PR #1136, opened 2017-11-23
Add config option to set a custom Sentry Transport

GenPage:
Ran into an issue the other day where I needed to set a custom Sentry Transport and was unable to. This is a first pass at how this might work. The implementation is a bit complex due to the need to dynamically import the transport class from raven, based on the configured transport provided by the user. 

List of transports and their classes can be found here: https://docs.sentry.io/clients/python/transports/

Example config option would be:

`SENTRY_TRANSPORT = ('RequestsHTTPTransport', 'raven.transport.requests')`

### DIFF
--- a/docs/user_guide/sentry.rst
+++ b/docs/user_guide/sentry.rst
@@ -49,6 +49,9 @@ To setup Errbot with Sentry:
 * Open up your bot's config.py
 * Set **BOT_LOG_SENTRY** to *True* and fill in **SENTRY_DSN** with the DNS value obtained previously
 * Optionally adjust **SENTRY_LOGLEVEL** to the desired level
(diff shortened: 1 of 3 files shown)


GenPage:
@zoni @gbin Any thoughts on this? Is this a viable solution?

andrewthetechie:
+1 on needing this. We'll have to fork errbot to work with our internal sentry without this change

gbin:
Awesome, can you add it to the config sample and the doc at the same time so people can discover that more easily? Thanks

GenPage:
@gbin Added docs and config-sample for your approval/merge.

GenPage:
@gbin-argo I am unable to merge, not sure if that is intentional or if you were just staging changes

`Only those with write access to this repository can merge pull requests.`

PR #1140, opened 2017-11-25
Make !repo search works out of the box

maxandersen:
Why:

 *  https://repos.errbot.io/repos.json has no valid certificate
    thus out-of-box install just fails using !repo search

This change addreses the need by:

 * update default to point to less secure but at least working index
   http://version.errbot.io/repos.json

### DIFF
--- a/errbot/bootstrap.py
+++ b/errbot/bootstrap.py
@@ -17,7 +17,7 @@ HERE = path.dirname(path.abspath(__file__))
 CORE_BACKENDS = path.join(HERE, 'backends')
 CORE_STORAGE = path.join(HERE, 'storage')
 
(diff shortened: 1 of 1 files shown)


maxandersen:
I'm sure this is not the optimal fix but since it makes errbot not working out of the box I thought it was worth to at least try and fix it.

btw. error in the logs when using default is:

`ssl.SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed`

gbin:
I let the certificate expire yesterday, I am working on fixing this. Sorry
about that.

maxandersen:
ah - so that explains why I couldn't find someone else have the problem :)

sijis:
This should probably be closed as the certificate is now valid.

PR #1147, opened 2017-12-27
Adding parameter in config config.MESSAGE_SIZE_LIMIT

fernand0:
Following Telegram backend, I'm proposing changing the way limits are managed and also the name of the variable.


gbin:
But you always overwrite the limit at init, I am confused. Do you want it to take a global setting from the bot?

fernand0:
I'm sorry, I proposed this some time ago. You are true, maybe it should go in the initialization part?
But in the existing code you can change it everytime, anyway:

    limit = min(self.bot_config.MESSAGE_SIZE_LIMIT, SLACK_MESSAGE_LIMIT)

PR #1151, opened 2018-01-05
Updating flows status functionality

aliegha:
This is in reference to issue #979 and the purpose is to allow all users to see the status of their current flows.

Logic changes:
1. removed admin only flag from flows_status botcmd
2. added check user function and logic to check and see if the requesting user either has current flows they started, or is an BOT ADMIN. If the user does not have any active flows, and is not an admin they will receive an 'No Flow started for current user'
3. I moved the if args statement under an if else statement that checks to see if there are any flows in flight (currently if you !flows status <arg> and there were no flows you  get no response, now you get a No Flow Started).

I have tested this command with both an admin and non-admin user:
Admin:
image

non-admin user that did not start the flow:
image

Result:

Admins will see all flows in flight
non-admins will only see their flows(if any)


errbot/core_plugins/flows.py
```diff
@@ -127,3 +137,13 @@ def flows_kill(self, _, user, flow_name):
         if flow:
             return flow.name + ' killed.'
         return 'Flow not found.'
+
+    def check_user(self, msg, flow):
+        """Checks to make sure that either the user started the flow, or is a bot admin
+        """
+        authorized = False
+        if glob(get_acl_usr(msg), self.bot_config.BOT_ADMINS):
+            authorized = True
```
gbin: return True

gbin: return True

errbot/core_plugins/flows.py
```diff
@@ -127,3 +137,13 @@ def flows_kill(self, _, user, flow_name):
...
+        elif glob(get_acl_usr(msg), flow.requestor.person):
+            authorized = True
```
gbin: return True

errbot/core_plugins/flows.py
```diff
@@ -127,3 +137,13 @@ def flows_kill(self, _, user, flow_name):
         if flow:
             return flow.name + ' killed.'
         return 'Flow not found.'
+
+    def check_user(self, msg, flow):
+        """Checks to make sure that either the user started the flow, or is a bot admin
+        """
+        authorized = False
```
gbin: remove

errbot/core_plugins/flows.py
```diff
@@ -127,3 +137,13 @@ def flows_kill(self, _, user, flow_name):
...
+        if glob(get_acl_usr(msg), self.bot_config.BOT_ADMINS):
+            authorized = True
+        elif glob(get_acl_usr(msg), flow.requestor.person):
+            authorized = True
+        return authorized
```
gbin: return False

errbot/core_plugins/flows.py
```diff
@@ -72,28 +73,37 @@ def flows_start(self, msg, args):
         self._bot.flow_executor.start_flow(flow_name, msg.frm, context)
         return 'Flow **%s** started ...' % flow_name
 
-    @botcmd(admin_only=True)
+    @botcmd()
     def flows_status(self, _, args):
```
gbin: if you use msg, name it:

def flows_status(self, msg, args):

errbot/core_plugins/flows.py
```diff
@@ -72,28 +73,37 @@ def flows_start(self, msg, args):
...
-                        template = '\>>> {} is using flow \*{}\* on step \*{}\*\nNext Step(s): \n{}'
-                        text = template.format(str(flow.requestor),
-                                               flow.name,
-                                               str(flow.current_step),
-                                               '\n'.join(next_steps))
-                        response.write(text)
+                    if args:
+                        for flow in self._bot.flow_executor.in_flight:
+                            if self.check_user(_, flow):
+                                if flow.name == args:
```
gbin: this is and:
if self.check_user(msg, flow) and flow.name == args:

aliegha:
"as an admin how can I check all the states?"

Are you meaning check the states of all inflight flows ?

If the user that requests !flows status is in the admin group,  they will receive a status response for all in flight flows. If they are not than they will only receive a status for a flows that they started. If the user has not started any flows than they will receive the no flow started for {user}. 

As far as I can tell the original code would only give back info on inflight flows as well...

aliegha:
Changes have been made as  requested. I am not sure why the build failed. I just retested this branch by running tox on my own machine and all environments succeeded. @gbin Anything else that I need to change?

PR #1160, opened 2018-02-14
Add support for specifying a custom log formatter

oz-linden:
Our deployment environment uses logs formatted as json, so to use this I needed to add the ability to specify a custom formatter.

### DIFF
--- a/errbot/config-template.py
+++ b/errbot/config-template.py
@@ -109,6 +109,9 @@ PLUGINS_CALLBACK_ORDER = (None, )
 # the user running Err.
 #AUTOINSTALL_DEPS = True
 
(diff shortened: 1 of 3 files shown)


errbot/config-template.py
```diff
@@ -109,6 +109,9 @@
 # the user running Err.
 #AUTOINSTALL_DEPS = True
 
+# To use your own custom log formatter, uncomment and set BOT_LOG_FORMATTER
+# BOT_LOG_FORMATTER =
```
zoni: Could include a link to the Python docs which explain the formatter syntax here? This would help tremendously with people wanting to use their own format who aren't sure how to specify it and/or where to find a reference of what is supported.

zoni:
Much appreciated, thanks!

PR #1162, opened 2018-02-21
Fix mentions in text backend

sijis:
When text appears with a `@` character, it would kill the bot.

```
[@saviles ➡ @zombie] >>> Breaking @backend
  File "/home/saviles/.local/share/virtualenvs/errbot-6RMKkNNT/bin/errbot", line 11, in <module>
    load_entry_point('errbot', 'console_scripts', 'errbot')()
  File "/home/saviles/data/git/github/errbot/errbot/cli.py", line 297, in main
    bootstrap(backend, root_logger, config, restore)
  File "/home/saviles/data/git/github/errbot/errbot/bootstrap.py", line 225, in bootstrap
    bot.serve_forever()
  File "/home/saviles/data/git/github/errbot/errbot/backends/text.py", line 289, in serve_forever
    mentioned = [self.build_identifier(word[1:]) for word in re.findall(r"@[\w']+", entry)
  File "/home/saviles/data/git/github/errbot/errbot/backends/text.py", line 290, in <listcomp>
    if word.startswith('@')]
  File "/home/saviles/data/git/github/errbot/errbot/backends/text.py", line 374, in build_identifier
    raise ValueError('An identifier for the Text backend needs to start with # for a room or @ for a person.')
ValueError: An identifier for the Text backend needs to start with # for a room or @ for a person.
```
This fixes it so it now it does not break:
```
[@saviles ➡ @zombie] >>> breaking @backend

[@saviles ➡ @zombie] >>> 

``

### DIFF
--- a/errbot/backends/text.py
+++ b/errbot/backends/text.py
@@ -286,8 +286,7 @@ class TextBackend(ErrBot):
 
                 self.callback_message(msg)
 
-                mentioned = [self.build_identifier(word) for word in re.findall(r"@[\w']+", entry)
-                             if word.startswith('@')]
+                mentioned = [self.build_identifier(word) for word in re.findall(r'(?<=\s)@[\w]+', entry)]
                 if mentioned:
                     self.callback_mention(msg, mentioned)
 
--- a/errbot/backends/text.py
+++ b/errbot/backends/text.py
@@ -286,7 +286,7 @@ class TextBackend(ErrBot):
 
                 self.callback_message(msg)
 
-                mentioned = [self.build_identifier(word[1:]) for word in re.findall(r"@[\w']+", entry)
+                mentioned = [self.build_identifier(word) for word in re.findall(r"@[\w']+", entry)
                              if word.startswith('@')]
                 if mentioned:
                     self.callback_mention(msg, mentioned)

### CODE REVIEW

File: errbot/backends/text.py [hunk 1]
```diff
@@ -286,8 +286,7 @@ def serve_forever(self):
 
                 self.callback_message(msg)
 
-                mentioned = [self.build_identifier(word[1:]) for word in re.findall(r"@[\w']+", entry)
-                             if word.startswith('@')]
+                mentioned = [self.build_identifier(word) for word in re.findall(r"^@[\w']+", entry)]
```
zoni: Previously that regex was unachored, now it's anchored to the start of the string. Doesn't this break mention detection? (I imagine previously it matched `"hello @person1 @person2"`, I'm not sure it still would with this change).

sijis: There was also a condition i removed, which had `if word.startswith('@')`, so anchoring in the regex should have the same effect. I also removed the stripping of the 1st character sent to `build_identifier()`, so it sends `@saviles`, instead of `saviles` (for example).

Additionally, `build_identifier()` is also checking if the line begins with `@`. When it matches in there, it processes it as a 'mention'.
https://github.com/errbotio/errbot/blob/master/errbot/backends/text.py#L373

zoni: I think the anchor is wrong though, with it, it won't match anything:

```
>>> entry = "Hi there, said @nick to @sijis"
>>> re.findall(r"^@[\w']+", entry)
[]
>>> re.findall(r"@[\w']+", entry)
['@nick', '@sijis']
```

Though that's not quite right either:

```
>>> entry = "Hi @somebody, you can email me at person@example.com."
>>> re.findall(r"@[\w']+", entry)
['@somebody', '@example']
```

A regex like this may be more appropriate:

```
>>> entry = "Hi @somebody, you can email me at person@example.com."
>>> re.findall(r"(?<=\s)@[\w']+", entry)
['@somebody']
```

sijis: The question, for me, is what are we considering a 'mention'? Is it *any* `@user` in string or must it be only in the beginning (like `@user ping`)?

zoni: If I'm not mistaken, it is/was intended as a username being mentioned anywhere in the message, not necessarily at the start.

(For people wanting to specifically only get mentions at the start of message, other mentions could easily be filtered out inside `callback_mention`)

zoni: Looking at the signature of [callback_mention](http://errbot.io/en/latest/user_guide/plugin_development/mentions.html?highlight=callback_mention), it talks about `people` (plural) as well.

sijis: I discussing with a few guys, it makes sense to match all `@username`. The only question regarding the regex, is why is the single quote there? It tends to match stuff like: `@'First_Name', @'First`. Is that expected?

zoni: I noticed that as well and it doesn't make a lot of sense to me either. I'm thinking it might have been an oversight that slipped in somehow.

sijis: Fixed it all up using your regex and removing the single quote.

PR #1164, opened 2018-02-25
Slack backend member_joined_channel event handler

lanthos:
Added event handler to the Slack backend for the member_joined_channel event (reference https://api.slack.com/events/member_joined_channel).

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -534,7 +534,9 @@ class SlackBackend(ErrBot):
 
     def _member_joined_channel_event_handler(self, event):
         """Event handler for the 'member_joined_channel' event"""
-        self.callback_room_joined(SlackRoom(channelid=event['channel'], bot=self))
+        user = event['user']
+        if user == self.bot_identifier:
+            self.callback_room_joined(SlackRoom(channelid=event['channel'], bot=self))
 
     def userid_to_username(self, id_):
         """Convert a Slack user ID to their user name"""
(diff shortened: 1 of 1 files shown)


errbot/backends/slack.py
```diff
@@ -531,6 +532,12 @@ def _message_event_handler(self, event):
         if mentioned:
             self.callback_mention(msg, mentioned)
 
+    def _member_joined_channel_event_handler(self, event):
+        """Event handler for the 'member_joined_channel' event"""
+        user = event['user']
+        if user == self.bot_identifier:
```
gbin: does it work?
`user` is a string and `self.bot_identifier` is a SlackPerson. Maybe reconstruct the SlackPerson like for the presence event handler and compare? `SlackPerson(self.sc, event['user'])`

lanthos: Good catch!  I tested that it didn't kick off when a person joined the channel but forgot to also check that it worked when the bot joined the channel.  I've updated the code and tested both instances and it's working as expected.  Thank you!

lanthos:
Hi @gbin I made the change as requested but I'm not sure how to mark it as having been done?  I did a push to the PR and it shows the latest code when I look at it.  Please let me know if there is anything else you need on this!  For what it's worth, I'm running with these changes on the errbot that we use for Slack and it's working wonderfully.

zoni:
Related historic discussion on presence may be found in https://github.com/errbotio/errbot/pull/244
and https://github.com/errbotio/errbot/pull/278

lanthos:
@zoni That's probably a bad choice of wording by calling the branch member_joined_channel (I named it after the API call from Slack, since they don't have a different API call for bots or general members).  This was created as a way to be able to detect when the bot joined a channel since callback_room_joined wasn't setup for the Slack backend.  Sorry if there was some confusion there!  The code as written won't react to non-bot users.

PR #1170, opened 2018-03-14
Sort help commands

feiming:
Current !help commands is unsorted. Sort before appending to usage variable.

### DIFF
--- a/errbot/core_plugins/help.py
+++ b/errbot/core_plugins/help.py
@@ -112,7 +112,7 @@ class Help(BotPlugin):
                     doc=cls.__errdoc__.strip() or '',
                 )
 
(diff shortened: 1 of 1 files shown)


sijis:
I'm trying out the patch and i'm not seeing a difference in the `!help` output locally.

Could you provide a screenshot of your changes before and after?

Lastly, what version of Python and Errbot are you using?

Thank you for taking the time and sending a patch.

feiming:
Python 3.4.3 and Errbot 5.1.2 with slack
Before
before

After
after

sijis:
@zoni I have no issues with the patch. When I tested it locally, everything was ordered alphabetically even without the patch.

PR #1173, opened 2018-03-20
Make it possible to use `/command@botname`

SvenNederhoff:
as documented by telegram, bots must be capable to receive bot messages
https://core.telegram.org/bots#commands

### DIFF
--- a/errbot/backends/telegram_messenger.py
+++ b/errbot/backends/telegram_messenger.py
@@ -1,5 +1,6 @@
 import logging
 import sys
+import re
 
 from errbot.backends.base import RoomError, Identifier, Person, RoomOccupant, Stream, ONLINE, Room
 from errbot.core import ErrBot
@@ -182,6 +183,7 @@ class TelegramBackend(ErrBot):
 
         identity = config.BOT_IDENTITY
         self.token = identity.get('token', None)
+        self.botname = identity.get('username', None)
         if not self.token:
             log.fatal(
                 "You need to supply a token for me to use. You can obtain "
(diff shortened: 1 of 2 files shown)


errbot/backends/telegram_messenger.py
```diff
@@ -182,6 +183,7 @@ def __init__(self, config):
 
         identity = config.BOT_IDENTITY
         self.token = identity.get('token', None)
+        self.botname = identity.get('username', None)
```
zoni: I believe you could get the bot's username through the [getMe](https://core.telegram.org/bots/api#getme) API call, so there would be no need to require the user to set it manually.

errbot/backends/telegram_messenger.py
```diff
@@ -256,6 +258,8 @@ def _handle_message(self, message):
             log.warning("Unhandled message type (not a text message) ignored")
             return
 
+        if re.match('\/.[^ ]*@{}'.format(self.botname), message.text):
```
zoni: `/` is not a special character in PCRE, there should be no need for escaping. Regexes are typically also entered using a raw litereal to avoid newline escapes and such: `re.match(r"/.[^ ]*@{}".format(self.botname), message.text)`

errbot/backends/telegram_messenger.py
```diff
@@ -256,6 +258,8 @@ def _handle_message(self, message):
...
+            message.text = message.text.strip("@{}".format(self.botname))
```
zoni: I don't see how this could ever work. Strip removes characters from beginning/end of string, it can't remove a substring in the middle of a string.

zoni: I think you can drop `re.match` and use `re.sub` instead for this:

```python
>>> import re                                       
>>> message = "/echo@errbot Hello world!"
>>> re.sub(r"^(/[^ ]+?)@errbot ", r"\1 ", message)                                                                                                                                                                 
'/echo Hello world!' 
```

gbin:
@SvenNederhoff ping?

SvenNederhoff:
@gbin Im here but looking for some time to develop a better consept. i'm chatting with @zoni about it

gbin:
It went stale, please reopen when ready

PR #1176, opened 2018-03-21
Make Systemd example file auto startable

SvenNederhoff:
Without the `[Install]` section the service won't auto start after reboot.


zoni:
Thanks!

PR #1177, opened 2018-03-21
Add environment path tho example file

SvenNederhoff:
Without this addition Errbot cant find the `git` command which makes it impossible to use `!repo install`

### DIFF
--- a/docs/code_examples/systemd.service
+++ b/docs/code_examples/systemd.service
@@ -4,6 +4,7 @@ After=network.service
 
 [Service]
 Environment="LC_ALL=en_US.UTF-8"
+Environment="PATH=/usr/bin:/path/to/errbot/virtualenv/bin"
 ExecStart=/path/to/errbot/virtualenv/bin/errbot --config /path/to/errbot/config.py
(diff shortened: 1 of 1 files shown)


docs/code_examples/systemd.service
```diff
@@ -4,6 +4,7 @@ After=network.service
 
 [Service]
 Environment="LC_ALL=en_US.UTF-8"
+Environment="PATH=/usr/bin:/path/to/errbot/virtualenv/bin"
```
zoni: We should probably set `/bin/` if we're setting `$PATH` here.

SvenNederhoff: maybe add the full default path ? like `/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin`

zoni: The term "default" is a bit subjective and OS-vendor dependent, but that seems sensible yes.

SvenNederhoff:
>The term "default" is a bit subjective and OS-vendor dependent, but that seems sensible yes.

the term Common use would be more aplicible

Issue #1179, opened 2018-03-26
Enable proxy configuration for SlackClient


deferato:
### I am...

* [*] Suggesting a new feature

### I am running...

* Errbot version: 5.1.3
* OS version: Docker-AmazonLinux
* Python version: 3.6
* Using a virtual environment: no

### Issue description

Please enable a specific config for Slack enabling set the proxy.
The SlackClient package allow to set a proxy when you start the client
If you allow to pass this option on the BOT_TOKEN, and then pass to the SlackClient would work.

### Additional info
https://github.com/slackapi/python-slackclient/blob/07cce9bf15bee2e5151e88374f700fab46381d3d/slackclient/server.py#L156

sijis:
I *think* you should be able to get that token and put it into the `config.py`.

So it'd look something like this:
```
import logging
import os

BACKEND = 'Slack'
BOT_IDENTITY = {
    'username': 'mybot',
    'token': os.environ['BOT_TOKEN'],
}
# all your other config options
```

sijis:
Sorry. I misread your issue.

PR #1180, opened 2018-03-27
Including the option to set a proxy when connecting using the SlackClient

deferato:
We are trying to use the Errbot, but in our infrastructure we use Proxy and to bypass this, I've that SlackClient already enables Proxy when starting the connection.

I included this option on the config and the serve_once code to be able to connect using proxy too.

Issue #1179


GenPage:
@deferato Nice! Thank you for including documentation for this as well. LGTM

gbin:
thx

PR #1186, opened 2018-04-12
Added email property for slack backend SlackPerson class

aaugustinas:
It's useful to return slack user email together with other slack user information. In enterprise slack edition company users register in slack with company email but can change other attributes freely. Email is the only user identification which can be mapped to user account.

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -153,6 +153,15 @@ class SlackPerson(Person):
         return f'@{self.username}'
 
     @property
+    def email(self):
+        """Convert a Slack user ID to their user email"""
+        user = self._sc.server.users.find(self._userid)
+        if user is None:
+            log.error("Cannot find user with ID %s" % self._userid)
+            return "<%s>" % self._userid
+        return user.email
+
+    @property
     def fullname(self):
         """Convert a Slack user ID to their user name"""
         user = self._sc.server.users.find(self._userid)
--- a/errbot/backends/base.py
+++ b/errbot/backends/base.py
@@ -69,6 +69,16 @@ class Person(Identifier):
         """
         pass
 
+    @property
+    @abstractmethod
+    def email(self) -> str:
+        """
+        Some backends have the email of a user.
+
+        :return: the email of this user if available.
+        """
+        pass
+
 
 class RoomOccupant(Identifier):
     @property
--- a/errbot/backends/test.py
+++ b/errbot/backends/test.py
@@ -40,11 +40,12 @@ class TestPerson(Person):
     methods exposed by this class.
     """
 
-    def __init__(self, person, client=None, nick=None, fullname=None):
+    def __init__(self, person, client=None, nick=None, fullname=None, email=None):
         self._person = person
         self._client = client
         self._nick = nick
         self._fullname = fullname
+        self._email = email
 
     @property
     def person(self):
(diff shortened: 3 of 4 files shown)


GenPage:
Hey @folijosvanagas! Thanks for the PR, this would be really great as we also use Enterprise Slack as well.

The tests are failing because `TestPerson` extends the abstract `Person` and needs an implementation of the email property

https://github.com/folijosvanagas/errbot/blob/master/errbot/backends/test.py

folijosvanagas:
Thanks. Updated `TestPerson` with email as you suggested.

folijosvanagas:
Please check code changes.
CI is failing due to codestyle mistakes in files which I haven't touched.

GenPage:
LGTM! @gbin or @zoni this should be good to merge.

errbot/backends/slack.py
```diff
@@ -157,6 +157,15 @@ def aclattr(self):
         # an incorrect format from SlackMUCOccupant.
         return "@%s" % self.username
 
+    @property
+    def email(self):
+        """Convert a Slack user ID to their user email"""
+        user = self._sc.server.users.find(self._userid)
+        if user is None:
+            log.error("Cannot find user with ID %s" % self._userid)
+            return "<%s>" % self._userid
```
gbin: this is a strange return, how the calling code knows it could not find an email?

folijosvanagas: What do you think should be returned?
By the way, it's very unlikely for a slack user not to have an email :)

GenPage: @gbin any update on this?

ngrayson-amfam: @gbin You still working on this?

sijis: I suppose the real question is... does all user objects return an email address? Is there a possibility where a user is found but the `user.email` attribute is `None` or not valid?

GenPage: The bot will need a special scope to be able to access them if I understand this correctly.

>If you want access to email addresses, you'll need the new OAuth permission scope, users:read.email. It provides an explicit, additive way to request access to team email addresses.

>Additionally, the bot scope will no longer grant bot user tokens access to email addresses. Bot users must utilize a user token and the users:read.email scope instead.

https://api.slack.com/changelog/2017-04-narrowing-email-access

GenPage: I wonder if this is just for the Web API methods and how this affects RTM

gbin: My question was a contract question...
If the user is not found, the property should be None and not "<id>".

Imagine you want to parse the email...

GenPage: @gbin Very good point, I would say merge this and we can address/clean-up the concern with no user found and trying to return an id with another PR as there are other methods that follow this return format.

lethonomia:
Looks like this is completed, anything holding up the merge? Having emails accessible would be a huge win for command verification.

drahamim:
I'm very interested in this feature. Is there no way to do this with the current 6.1.4 bot?

frozenfoxx:
Same here, anything holding this up? Looks like the requested changes were made.

drahamim:
@sijis  is there a reason this cannot be merged?

sijis:
I'll have review that it mergeable as this was created a few years back and also ensure it still works as expected.

I'll target to look at this change this upcoming weekend.

sijis:
Rebased and fixed up an output issue i noticed.

PR #1194, opened 2018-04-27
Use sys.executable explicitly instead of just 'pip' to install plugin requirements

nicoddemus:
'pip' might not be available on PATH: for example when installing errbot on a
virtualenv and starting it with systemclt pointing directly to the executable:

    [Service]
    ExecStart=/home/errbot/.env/bin/errbot

'python -m pip' is semantically the same, and for example recommended by
conda-forge to build packages: https://conda-forge.org/docs/meta.html#use-pip


zoni:
Good call, thanks!

Issue #1191, opened 2018-04-25
warn_admins throws exception in Slack if an admin has notifications Snoozed


oz-linden:
In order to let us help you better, please fill out the following fields as best you can:

### I am...

* [X] Reporting a bug
* [ ] Suggesting a new feature
* [ ] Requesting help with running my bot
* [ ] Requesting help writing plugins
* [ ] Here about something else

### I am running...

* Errbot version: 5.2
* OS version: Debian Jessie
* Python version: 3.5.5
* Using a virtual environment: yes

### Issue description

In one of my plugins, I attempt to alert the admin whenever the bot connects by:

    def callback_connect(self):
        self.warn_admins("Connected " + self.version)

If this happens when an admin has suppressed notifications (Slack Snooze), an exception is thrown in the Slack backend (see below).

### Steps to reproduce

1. Add a callback_connect as above.
2. Snooze notifications in Slack, but remain connected.
3. Start the bot.

### Additional info

{"error_type": "KeyError", "error_message": "'bot_id'", "msg": "message event handler raised an exception", "level": "ERROR", "time": "2018-04-25T05:23:54.682733Z", "traceback": "KeyError
Python 3.5.5: /Users/oz/_config/virtualenv/botll/bin/python3.5
Wed Apr 25 05:23:54 2018

A problem occurred in a Python script.  Here is the sequence of
function calls leading up to the error, in the order they occurred.

 /Users/oz/_config/virtualenv/botll/lib/python3.5/site-packages/errbot/backends/slack.py in _dispatch_slack_message(self=<yapsy_loaded_plugin_Slack_0.SlackBackend object>, message={'channel': 'D597B3NMR', 'dnd_suppressed': True, 'event_ts': '1524648234.000340', 'is_ephemeral': True, 'subtype': 'bot_message', 'text': 'oz has snoozed notifications. Send one anyway? <...966959978/1524648234.000379|Send notification>\\u200b\\u200b\\u200b', 'ts': '1524648234.000379', 'type': 'message', 'user': 'USLACKBOT', 'username': 'slackbot'})
  418         try:
  419             log.debug(\"Processing slack event: %s\" % message)
  420             event_handler(message)
  421         except Exception:
  422             log.exception(\"%s event handler raised an exception\" % event_type)
event_handler = <bound method SlackBackend._message_event_handle...yapsy_loaded_plugin_Slack_0.SlackBackend object>>
[...]

PR #1195, opened 2018-04-29
Fixing #1191 by changing to event.get method

andrewthetechie:
This should fix https://github.com/errbotio/errbot/issues/1191

After this change, sending an admin warning to DnD'd slack admins does this:

[dnd_slack_admin.txt](https://github.com/errbotio/errbot/files/1958305/dnd_slack_admin.txt)

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -504,7 +504,7 @@ class SlackBackend(ErrBot):
             if subtype == "bot_message":
                 msg.frm = SlackBot(
                     self.sc,
-                    bot_id=event['bot_id'],
+                    bot_id=event.get('bot_id'),
                     bot_username=event.get('username', '')
                 )
             else:
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -516,7 +516,7 @@ class SlackBackend(ErrBot):
             if subtype == "bot_message":
                 msg.frm = SlackRoomBot(
                     self.sc,
-                    bot_id=event['bot_id'],
+                    bot_id=event.get('bot_id'),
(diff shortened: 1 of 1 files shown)


errbot/backends/slack.py
```diff
@@ -504,7 +504,7 @@ def _message_event_handler(self, event):
             if subtype == "bot_message":
                 msg.frm = SlackBot(
                     self.sc,
-                    bot_id=event['bot_id'],
+                    bot_id=event.get('bot_id'),
```
zoni: I'm a little bit worried what kind of effects this change could have where `bot_id` ends up being `None` in cases, but let's see how it goes. :smile: 

That said, if this is changed here, the same should be applied to the initialization of `SlackRoomBot` a little further down (line 519 currently).

andrewthetechie: Made that change.

GenPage:
@andrewthetechie @zoni I recommend reviewing #1073 as it seems these are both similar.

gbin:
I merged #1073

gbin:
Thx

zoni:
I think I prefer this approach over the one inf #1073 though. Returning `None` seems like a better API design over returning empty strings when these values can't be found.

@gbin what're your thoughts on that?

gbin:
@zoni you have a good point, I assumed the empty string was there to no break something else.

zoni:
From my understanding, the immediate problem is a KeyError exception on trying to access values that don't exist in the slack event. I suspect if it breaks further down the line, it'll currently break whether they're empty strings or None values.

gbin:
directly patched on master. We'll see where it goes.

PR #1197, opened 2018-05-01
XMPPRoomOccupant: add real_jid property

globin:
This adds the real JID of the room occupant they used to login, if the
errbot is a moderator of the MUC or the MUC is not anonymous.

Bikeshedding on a better name for the property is welcome.


zoni:
Nice, thanks! I can't think of a better property name so I'm happy with it. :smile:

PR #1198, opened 2018-05-01
split_string_after: fix empty string handling

globin:
This would otherwise not yield anything for an empty string, i.e.
causing send_card to not send anything on the Slack backend with an
empty body.


zoni:
Returning the original input (an empty string) in this case seems very sensible to me. :+1: Thanks.

PR #1200, opened 2018-05-01
card template: remove broken colour rendering

globin:
This gets rendered verbatim in Jabber at least for me, if this is useful somewhere it might be better to only change it for XMPP. I could also not manage to understand how this should work properly.


gbin:
So it looks like we have to implement the card rendering for XMPP, here you will brake the rendering for Text etc..

Issue #1204, opened 2018-05-05
Travis CI: Master Branch Broken due to codestyle 


nvzard:
In order to let us help you better, please fill out the following fields as best you can:

### I am...

Reporting a bug

### Issue description

the most problematic/prolific is E252 'missing whitespace around parameter equals' , which is a new rule enforcing a pep8 style guidance for parameters with type annotations & defaults. disabling that rule atm might be best , as it will break (i.e. cause merge conflicts) a lot of pending patches

jayvdb:
Sort of replying to https://github.com/errbotio/errbot/pull/1205#issuecomment-386862712

Just fix the problems.

Split it into separate commits so the maintainer can easily choose which ones they like, and discard the ones they dont like.

The regex problems are *real* problems which will become warnings in Python 3.7, so they *must* be fixed.

zoni:
I'd say #1205 was merged a little too greedily by Guillaume. :wink: I now have a follow-up to this: https://github.com/errbotio/errbot/pull/1207

PR #1205, opened 2018-05-05
PyCodestyle: Ignore E252, W605

nvzard:
Ignoring these rules is the best at the moment
as they'll break a lot of pending patches.

Fixes https://github.com/errbotio/errbot/issues/1204


jayvdb:
Replied on the issue https://github.com/errbotio/errbot/issues/1204

PR #1207, opened 2018-05-10
Format codebase according to modern style rules

zoni:
This fixes a number of warning and errors that were being reported by
`pycodestyle`, making CI happy again. It also formats all code according
to the rules of [black], simplifying future code-style formatting.
Black is also added to the CI pipeline, enforcing this codestyle going
forward.

Black is opinionated which may initially cause some friction, however I
personally believe this to be a good thing. It ensures a consistent look
and bypasses any ambiguity and pointless bikeshedding over personal
style preferences.

I've personally had good responses to enforcing consistent code style
from similar approaches in Go (via gofmt) and Elixir (mix format) in
other projects.

It should be noted that black is currently calling itself alpha-level
software, however after reviewing the auto-formatted results from this
commit I believe it to be quite reliable already. Furthermore, CI will
run in check-mode only so worst-case, it fails a build which should not
be getting failed.

[black]: https://pypi.org/project/black/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ nohup.out
 /.coverage
 /data
 /config.py.*
+.pytest_cache
 
 # tools
 /tools/Home.md
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ sudo: false
 language: python
 matrix:
   include:
+    - python: 3.6
+      env: TOXENV=black
     - python: 3.4
       env: TOXENV=py34
     - python: 3.5
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,16 +4,18 @@ matrix:
   include:
     - python: 3.6
       env: TOXENV=black
+    - python: 3.6
+      env: TOXENV=codestyle
+    - python: 3.6
+      env: TOXENV=isort
+    - python: 3.6
+      env: TOXENV=pypi-lint
     - python: 3.4
       env: TOXENV=py34
     - python: 3.5
       env: TOXENV=py35
     - python: 3.6
       env: TOXENV=py36
-    - python: 3.6
-      env: TOXENV=pypi-lint
-    - python: 3.6
-      env: TOXENV=codestyle
 
 install: pip install tox
 before_script: cp tests/config-travisci.py config.py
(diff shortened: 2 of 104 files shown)


zoni:
Cc. @gbin @GenPage @sijis @jamesob

I'd love to hear how you folks would feel about enforcing black in our CI pipeline. Yay? Nay?

gbin-argo:
I am sad for all the ' flipped to ", can we set the preference to ' ?

zoni:
> I am sad for all the ' flipped to ", can we set the preference to ' ?

I'm afraid that's not possible. Black wants double-quotes for strings except for some special cases (like allowing `'` to avoid having to escape a `"` inside the string).

> Also it look like it is pretty aggressive at splitting lines: 2 lines can get split into 8, is it really set on 120 chrs lines?

I thought it was, as that's specified in places like `setup.cfg` for flake8 which black's documentation on "Line length" led me to believe would be honored. Turns out it wasn't, so I just updated the PR with line-length set to 120.

> I'd probably also include isort if that's not already part of the ci checks.

I didn't know about isort. I'm about to head to bed but I'll investigate this tomorrow.

sijis:
We do this with [isort](https://github.com/timothycrosley/isort) in our `tox` config:
`isort --check-only --recursive --diff src/`.

If its a non-zero exit code, something was not isorted.

gbin:
what about https://github.com/hhatto/autopep8 ?

zoni:
@sijis thanks for the tip on isort, I've added it in this branch as well. It was a bit finicky though, getting it to play nice with ordering of mock/pytest between the virtualenv managed by tox and my own development virtualenv. We'll have to see whether it's not too much of a hassle to contributors.

@gbin 
> what about https://github.com/hhatto/autopep8 ?

I greatly value the work put into autopep8 but have witnessed some people having issues with it which has slowed adoption. There seems to be some buzz around black at the moment so I've picked that one in order to lift on that buzz, hoping it becomes more widespread.

> I also forked black, the patch to prefer ' is trivial

I'd argue the whole point of black is to remove these personal style preferences from projects and converge on a shared consensus in the broader python ecosystem, which in my experience prefers `"` over `'` for strings by default.

gbin:
o I gonna go ahead and push the simple fix for the formatting issue + a bunch of refactors on the very long list completions that were too hard to fix as is. We should still continue the black discussion, I think it is a good idea that I would like to mature before jumping ship.

PR #1214, opened 2018-05-14
Replace back Bottle & Rocket by Flask

gbin-argo:
This is an initial port of the errbot webhook feature to Flask.

Rocket is not only abandoned but we had to maintain a lot of patches out of the official tree. It is time to come back to something more maintained and standard: Flask.

Flask can replace both Bottle and Rocket so let's remove one dependency.

### DIFF
--- a/errbot/core_plugins/webserver.py
+++ b/errbot/core_plugins/webserver.py
@@ -90,9 +90,6 @@ class Webserver(BotPlugin):
             self.log.info('Webserver is not configured. Forbid activation')
             return
 
-        #self.webserver = Rocket(interfaces=interfaces,
-        #                        app_info={'wsgi_app': bottle_app}, )
-
         if self.server_thread:
             raise Exception('Invalid state, you should not have a webserver already running.')
         self.server_thread = Thread(target=self.run_server, name='Webserver Thread')
--- a/errbot/core_plugins/webserver.py
+++ b/errbot/core_plugins/webserver.py
@@ -114,7 +114,8 @@ class Webserver(BotPlugin):
             ssl = self.config['SSL']
             self.log.info('Starting the webserver on %s:%i' % (host, port))
             ssl_context = (ssl['certificate'], ssl['key']) if ssl['enabled'] else None
-            self.server = ThreadedWSGIServer(host, ssl['port'] if ssl_context else port, flask_app, ssl_context=ssl_context)
+            self.server = ThreadedWSGIServer(host, ssl['port'] if ssl_context else port, flask_app,
+                                             ssl_context=ssl_context)
             self.server.serve_forever()
             self.log.debug('Webserver stopped')
         except KeyboardInterrupt as _:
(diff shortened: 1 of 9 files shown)


zoni:
The PR description is a little misleading/factually incorrect. Bottle the web framework is still active and maintained. We do not have any custom patches for it.

You're confusing this with rocket, the embedded webserver we settled on. That is indeed unmaintained and carrying around our own fork with a few patches.

That said though, we initially switched from flask to bottle due to missing python 3 support by flask at the time. Seeing as how python 3 support has matured on flask, and given it's more widespread use/popularity, I have no objections to switching back now.

If we do though, the next errbot release should be a major version release (hitting 6.0.0). We will also need to update some of the documentation, as found at http://errbot.io/en/latest/user_guide/plugin_development/webhooks.html, which currently lists bottle-specific code.

(Haven't had time to look at the actual code changes yet)

errbot/core_plugins/webserver.py
```diff
@@ -74,48 +76,61 @@ def get_configuration_template(self):
...
     def deactivate(self):
-        if self.webserver is not None:
-            self.log.debug('Sending signal to stop the webserver')
-            self.webserver.stop()
+        if self.server is not None:
+            self.log.info('Shutting down the internal webserver.')
+            self.server.shutdown()
+            self.log.info('Waiting for the webserver thread to quit.')
+            self.server_thread.join()
+            self.log.info('Webserver shut down correcly.')
```
zoni: Typo: correcly -> correctly

errbot/core_plugins/webserver.py
```diff
@@ -74,48 +76,61 @@ def get_configuration_template(self):
...
+            host = self.config['HOST']
+            port = self.config['PORT']
+            ssl = self.config['SSL']
+            self.log.info('Starting the webserver on %s:%i' % (host, port))
+            ssl_context = (ssl['certificate'], ssl['key']) if ssl['enabled'] else None
+            self.server = ThreadedWSGIServer(host, ssl['port'] if ssl_context else port, flask_app,
+                                             ssl_context=ssl_context)
+            self.server.serve_forever()
+            self.log.debug('Webserver stopped')
+        except KeyboardInterrupt as _:
```
zoni: Maybe do `except KeyboardInterrupt:` (omitting the `as _` bit) since the exception object is not used?

errbot/core_plugins/webserver.py
```diff
@@ -74,48 +76,61 @@ def get_configuration_template(self):
...
+            self.log.exception('Keyboard interrupt, request a global shutdown.')
```
zoni: I think this should be logged at info level only.

errbot/core_plugins/webserver.py
```diff
@@ -74,48 +76,61 @@ def get_configuration_template(self):
...
+            self.log.info('webserver is ThreadedWSGIServer')
```
zoni: Can we remove this log message, it doesn't seem useful in this context? Otherwise I would move it to debug only.

gbin:
Thanks for the review! I fixed all the comments. Gonna merge it but if you find something after the fact I'll fix it.

PR #1219, opened 2018-05-20
Yapsictomy - part 1

gbin:
This is dealing with plugins and flows.

The added code is all modern, typed, Python 3.7 with backported dataclasses for 3.6-. 

There are still a bit of wobbling between path as str and Path, name vs PluginInfo etc... I expect future small refactoring cleanup PRs to simplify those easily.

Note: The backends are still loaded from Yapsy. This will be addressed in another PR.


andrewthetechie:
LGTM

gbin-argo:
This is 6.0. we will have a beta for a while ;)

PR #1229, opened 2018-05-31
Recover webserver functionality

Alvarord:
The code from the tutorial was not working. Adapt to the configuration.

Change functionality using only one port for the webserver.

Show the configuration that needs to be added once the certificate is
created.


gbin:
It is tested part of https://github.com/errbotio/errbot/blob/master/tests/webhooks_test.py

Alvarord:
Sorry, I missunderstood I will proof it like that then.

Thanks for the guide.

PR #1232, opened 2018-06-25
bootstrap.py: Create data/ dir at runtime if missing.

nvzard:


gbin:
Nope we don't want to do that: if a user makes a mistake we don't want to pollute a random directory out of it: it needs to fail and say the directory doesn't exist.

PR #1236, opened 2018-07-03
sanitize comma quotation marks too

pznamensky:
Slack client converts apostrophes into comma quotation marks. This PR adds them to the list of characters to be sanitized.

### DIFF
--- a/errbot/__init__.py
+++ b/errbot/__init__.py
@@ -28,7 +28,7 @@ webroute = route  # this allows plugins to expose dynamic webpages on Errbot emb
 # Same happens with quotations marks, which are required for parsing
 # complex strings in arguments
 # Map of characters to sanitized equivalents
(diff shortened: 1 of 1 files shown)


sijis:
Thank you for putting this together. It seems like an easy fix.

To better understand, were you seeing an error/issue when using the`@arg_botcmd()` in a plugin? Thanks.

pznamensky:
@sijis, I'm not sure I understand your question :)
In our case, we're using a homegrown plugin and sometimes we have to pass arguments in the following format:
`--some-param "some args with 'quotation marks' "`
which were converted to
`--some-param  “some args with  ‘quotation marks’ ”`
And with this fix, `@arg_botcmd()` works without errors for us.
As far as I remember, without this fix, it works fine too, but variable `some_param` contained unexpected `‘` or `’` instead of `'`.

sijis:
@pznamensky You answered my question. I was trying to gather what was the behavior you noticed and how your change fixes it.

I'm good with this. :+1:

PR #1240, opened 2018-07-10
Add email attribute to SlackPerson

ngrayson-amfam:
Reference to slackclient: https://github.com/slackapi/python-slackclient/blob/master/slackclient/user.py#L8

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -143,6 +143,14 @@ class SlackPerson(Person):
     def domain(self):
         return self._sc.server.domain
 
+    @property
+    def email(self):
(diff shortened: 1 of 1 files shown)


savilesupgrade:
Is the expectation to use this property within a plugin? Is that why this property needs to be exposed?

savilesupgrade:
This also seems related to PR https://github.com/errbotio/errbot/pull/1186/.

ngrayson-amfam:
@savilesupgrade Yep. I want to use the slack email to set an ec2 tag. You are right that #1186 will fix this. I didn't see that when I looked for it.

closing in favor of #1186

PR #1245, opened 2018-08-10
Add docker support for installing package dependencies

GenPage:
When setting up errbot in a docker container, setting up a virtualenv and activating it so the bot can properly install dependencies is a big hurdle. To make this more seamless, I propose an additional check to see if errbot is running inside of a docker container which allows direct package installation instead of user.

I used a separate `elif` even though the same command is called since its a different case.

### DIFF
--- a/errbot/plugin_manager.py
+++ b/errbot/plugin_manager.py
@@ -63,6 +63,10 @@ def install_packages(req_path: Path):
 
         Return an exc_info if it fails otherwise None.
     """
+    def is_docker():
+        with open('/proc/1/cgroup') as d:
+            return 'docker' in d.read()
+
     log.info('Installing packages from "%s".', req_path)
     # use sys.executable explicitly instead of just 'pip' because depending on how the bot is deployed
     # 'pip' might not be available on PATH: for example when installing errbot on a virtualenv and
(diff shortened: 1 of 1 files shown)


errbot/plugin_manager.py
```diff
@@ -75,6 +79,9 @@ def install_packages(req_path: Path):
         if hasattr(sys, 'real_prefix') or (hasattr(sys, 'base_prefix') and (sys.base_prefix != sys.prefix)):
```
sijis: Curious, why not add `is_docker()` to this condition? My first thought was
```
any([hasattr(sys, 'real_prefix'), is_docker()]) or (hasattr(sys, 'base_prefix') and (sys.base_prefix != sys.prefix))
```

errbot/plugin_manager.py
```diff
@@ -63,6 +63,10 @@ def install_packages(req_path: Path):
 
         Return an exc_info if it fails otherwise None.
     """
+    def is_docker():
+        with open('/proc/1/cgroup') as d:
+            return 'docker' in d.read()
```
sijis: In poking at this `/proc/1/cgroup` file, it seems (to me) probably safer to check for `/docker/` vs just `docker`.
I don't know the inner-workings of a container or docker but I just wondering if there is a scenario, like docker in docker or mounting a disk/cgroup where this could be a false positive.

GenPage: @sijis Solid concern. Let me do some testing. The /proc/1 pid is always the `init` process so it should always be the base filesystem. In this case, a non-docker container would be `/` so I'm not too worried about mounting a disk/cgroup but will double check docker in docker.

xcompass:
Hiting this issue right now. What's the workaround right now? Hopefully this will be merged soon!

sijis:
@xcompass My only thought of a current workaround is disable `AUTOINSTALL_DEPS`.
https://github.com/errbotio/errbot/blob/master/errbot/config-template.py#L110

PR #1248, opened 2018-08-23
Fix/backend import error messages

samueldg:
Fix small typos/omissions in the error messages when using a backend without the appropriate dependencies installed.

Originally found this issue with Slack, but figured I'd clean up the other while at it.

### DIFF
--- a/errbot/backends/hipchat.py
+++ b/errbot/backends/hipchat.py
@@ -23,7 +23,7 @@ try:
 except ImportError:
     log.exception("Could not start the HipChat backend")
     log.fatal(
-        "You need to install the hipchat support in order to use the HipChat.\n "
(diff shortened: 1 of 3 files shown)


sijis:
I'm wondering if the text, for all, should be something like. Thoughts?
```
    log.fatal(
        "You need to install the required package in order to use the $backend backend.\n"
        "You should be able to install this package using:\n"
        "pip install errbot[$backend]"
    )
```

samueldg:
@sijis This makes sense to me. I wonder if that specific template will work for all (since it requires all backends to be declared as "extras", which I don't know about), but I agree it would be cleaner to have only one format for these messages.

sijis:
The 'extras' are defined here: https://github.com/errbotio/errbot/blob/master/setup.py#L117-L122
If its in core, we should have an extra for it.

Issue #1249, opened 2018-08-27
Certificate of https://repos.errbot.io/ expired. Unable to install new plugins.


malt3:
In order to let us help you better, please fill out the following fields as best you can:

### I am...

* [x] Reporting a bug
* [ ] Suggesting a new feature
* [ ] Requesting help with running my bot
* [ ] Requesting help writing plugins
* [ ] Here about something else

### I am running...

* Errbot version: 5.2.0
* OS version: Arch Linux 4.17.6-1-ARCH
* Python version: 3.7.0
* Using a virtual environment: yes

### Issue description

The certificate of https://repos.errbot.io/  expired on 24th of August 2018. This server provides the plugin repo that is loaded whenever a new package is installed from https://repos.errbot.io/repos.json 

### Steps to reproduce

Install a repo using the following command:

```
!repos install [repo_name]
```
e.g.:
```
< !repos install https://github.com/drsm79/err-xkcd
> Installing https://github.com/drsm79/err-xkcd...
> Computer says nooo. See logs for details:
> b'repo_index'
```

### Additional info

Output of errbot:
```
ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: certificate has expired (_ssl.c:1045)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/malte/.virtualenvs/errbot-ve/lib/python3.7/site-packages/errbot/repo_manager.py", line 120, in index_update
    with urlopen(url=source, timeout=10) as request:  # nosec
  File "/usr/lib64/python3.7/urllib/request.py", line 222, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib64/python3.7/urllib/request.py", line 525, in open
    response = self._open(req, data)
  File "/usr/lib64/python3.7/urllib/request.py", line 543, in _open
    '_open', req)
  File "/usr/lib64/python3.7/urllib/request.py", line 503, in _call_chain
    result = func(*args)
  File "/usr/lib64/python3.7/urllib/request.py", line 1360, in https_open
    context=self._context, check_hostname=self._check_hostname)
  File "/usr/lib64/python3.7/urllib/request.py", line 1319, in do_open
    raise URLError(err)
urllib.error.URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: certificate has expired (_ssl.c:1045)>
[...]

PR #1251, opened 2018-08-31
Load plugins deterministically. Fixes #1010.

### LINKED ISSUE #1010: Clash renames are non-deterministic

### I am...

* [x] Reporting a bug
* [x] Requesting help with running my bot
* [ ] Here about something else

### I am running...

* Errbot version: 5.0.1
* OS version: Debian Jessie
* Python version: 3.4.2
* Using a virtual environment: Docker (https://hub.docker.com/r/rroemhild/errbot/)

### Issue description

I have a plugin with a `status` command defined. This of course clashes with the core `status` command. It is unknown until start-up if my `status` gets renamed to `myplugin-status` or the core to `health-status`.

### Steps to reproduce

Just create a 'helloworld' plugin but call the function `status`...

    $ errbot -T
    ...
    HelloWorld.status clashes with Health.status so it has been renamed hellowworld-status
    ...

Run it a few times and eventually it will respond differently

    $ errbot -T
    ...
    Health.status clashes with HelloWorld.status so it has been renamed health-status
    ...
    

### Additional info

Is there a way for my plugin to ensure it gets the preferred command name?
For now, I just use a different command name :)

KimNorgaard:


KimNorgaard:
Test error is caused by #1249.

GenPage:
Hopefully @gbin will get #1249 addressed soon

errbot/utils.py
```diff
@@ -146,7 +147,7 @@ def find_roots(path, file_sig='*.plug'):
             the file pattern to look for
        :return: a set of paths
     """
-    roots = set()  # you can have several .plug per directory.
```
gbin-argo: No, you have ignored the comment, you need to setup them right?

KimNorgaard: I am not sure I understand your question, but let me try to get it right.

Before, a `set` was used to return a unique list of paths in which to look for plugins. The problem with using a `set` is that it is unordered. I replaced the set with a `list`. This means that the paths will be added multiple times (as stated by the comment).

In order to make the list unique again and at the same same preserve order, I use an `OrderedDict`. I moved the original comment to the `return` statement to indicate this:

```python
    return list(collections.OrderedDict.fromkeys(roots))  # you can have several .plug per directory
```

If the purpose of the original `set` was to ensure the list was unique, then this PR should not change that. It just assures the order is always the same.

Am I missing something?

gbin-argo: Doh sorry about, thanks for the patch!

errbot/utils.py
```diff
@@ -160,8 +161,8 @@ def find_roots(path, file_sig='*.plug'):
                     log.debug('Ignore %s.', dir_to_add)
                     break
             else:
-                roots.add(dir_to_add)
-    return roots
+                roots.append(dir_to_add)
+    return list(collections.OrderedDict.fromkeys(roots))  # you can have several .plug per directory.s
```
sijis: nitpick. There's a typo in the comment `directory.s`.

errbot/utils.py
```diff
@@ -175,13 +176,13 @@ def collect_roots(base_paths, file_sig='*.plug'):
             the file pattern to look for
        :return: a set of paths
```
sijis: We should update the method signature.

errbot/utils.py
```diff
@@ -146,7 +147,7 @@ def find_roots(path, file_sig='*.plug'):
...
```
sijis: We should update the method docstring.

sijis:
Do you think we should update this line too, to send a list?
https://github.com/errbotio/errbot/blob/8c3d74995e316a14870527901ce89f2733691c53/errbot/backend_plugin_manager.py#L28

Its the only reference i see for `collect_roots()` used outside of `utils.py`.

sijis:
Otherwise, I think this PR looks great! Awesome! :+1:

GenPage:
@gbin-argo Need your approval now as well before merging

sijis:
I'm closing this in favor of https://github.com/errbotio/errbot/pull/1282. Its based off this work but it has additional feedback from other contributors.

Thank you for initiating these fixes.

Issue #860, opened 2016-10-02
re clone missing repos at startup


gbin:
If a plugin is marked as installed in the db but not present on disk, try to clone it.

It is useful because errbot deployments can be "static" with this feature (ie. a docker image that doesn't change for example) with no need to map external volumes etc.

PR #1256, opened 2018-10-01
Install missing repo's

demonbob:
This is useful for bootstraping, when you do not want to use the extra plugs directory. On deploy I insert the repo manually into repomgr.db as installed, then run repos update and it will download what is missing.

### DIFF
--- a/errbot/core_plugins/plugins.py
+++ b/errbot/core_plugins/plugins.py
@@ -121,11 +121,9 @@ class Plugins(BotPlugin):
         for d, success, feedback in results:
             if success:
                 yield f'Update of {d} succeeded...\n\n{feedback}\n\n'
-            else:
-                yield f'Update of {d} failed...\n\n{feedback}'
 
-            for plugin in self._bot.plugin_manager.getAllPlugins():
-                if plugin.path.startswith(d) and hasattr(plugin, 'is_activated') and plugin.is_activated:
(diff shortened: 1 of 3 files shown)


demonbob:
solves #860

demonbob:
BotRepoManager subclasses BotPluginManager, this gives it access direct access to plugin commands while retaining normal functions. This makes more sense to me as they are technically closely related in functions and this will allow for easier bootstraping of repo's and/or auto repair. 

* BotRepoManager is now subclasses BotPluginManager
* Fix for repo update to activate plugins when done.
* Legacy broken getAllPlugins() function call removed (only one left in the codebase)
** Replaced with get_plugin_by_path() function as repo update outputs filesystem paths
* Fix for remove_plugins_from_path() if repo is listed but plugin/files don not exist
* check_dependencies() moved back under BotPluginManager as both BotRepoManager and BotPluginManager needed, subclassing BotPluginManager solves the need for a single import from repo_manger
* Fix for repo uninstall

gbin:
I like the feature but I am not a fan of the major refactoring to just add this feature. For example semantically having BotRepoManager inherit BotPluginManager doesn't make sense. The repo manager manages repos and the other one the plugins. A plugin can exist without a repo etc..

sijis:
I think we conceptually like what this change will do. Please reopen once the concerns raised are addressed. Thank you.

PR #1259, opened 2018-10-19
Fix missing string formatting in "Command not found" output

Tenzer:
95bb61f0 attempted to switch the string formatting in `unknown_command()` to use
f-strings instead of string concatenation, but failed to add the `f` prefix to
the string, meaning the "Did you mean" part of the output when a nonexisting
command is attempted is outputted literally as:

> Did you mean "{self.bot_config.BOT_PREFIX}{alternatives}" ?

This fixes that so the variables will be replaced in the string as intended.


gbin-argo:
Doh ;)

gbin-argo:
Thx

Issue #1206, opened 2018-05-07
Errbot should fail if git binary is not present


sijis:
In order to let us help you better, please fill out the following fields as best you can:

### I am...

* [x] Reporting a bug

### I am running...

* Errbot version: Latest
* OS version: Fedora 28
* Python version:  3
* Using a virtual environment: yes

### Issue description

Installing a plugin fails silently if `git` binary is not present on the system.

### Steps to reproduce
* Install errbot
* ensure git is not installed
* run `errbot -T`, then try to install a plugin via `!repos install x`

### Additional info

This was brought up on gitter.im.

andrewthetechie:
I'll take a look at this

sijis:
This is no longer needed.

PR #1260, opened 2018-10-22
Fixing #1206, bot now fails to start on missing git

andrewthetechie:
https://github.com/errbotio/errbot/issues/1206

This moves the which method into utils and uses it to find git in the path on bot startup. If not found, the bot exits with an error.

Tested this in docker with python:3.6.6-slim which does not have git installed
```root@6cce6e09947f:~# errbot --init
Your Errbot directory has been correctly initialized !
Just do "errbot" and it should start in text/development mode.
root@6cce6e09947f:~# errbot -v
Errbot version 9.9.9
root@6cce6e09947f:~# errbot -T
git is either not installed or not accessible in the path. Please install it or add it to your path
```


errbot/bootstrap.py
```diff
@@ -74,6 +75,12 @@ def bot_config_defaults(config):
 
 
 def setup_bot(backend_name: str, logger, config, restore=None) -> ErrBot:
+    # Check if git is installed on the system and in the path. If not, error out immediately
+    if which("git") is None:
+        log.exception("git is either not installed or not accessible in the path. Please install it or add it to your "
```
sijis: I think something like this may read better. `"The required 'git' package is either not installed or accessible in your system path. Please install 'git' or ensure it's added to your system path."`

sijis:
I thought we had a docs page that mentioned requirements and 'git' was listed. Could you also validate and ensure that is the case?

I think this is a needed change for better user experience. Thank you for taking the time in putting this together.

andrewthetechie:
@sijis Updated the error message and added two other quick catches where git is actually executed:
* return code 126 == not executable
* return code 127 == doesn't exist

I also added a note under Setup/pre-requisites that the system has to have git installed and executable by the user that runs Errbot

GenPage:
LGTM, thank you for updating the docs as well

sijis:
I like everything about this. I just have 1 reservation/question. When I try this on a docker container, i'm seeing this:
```
[root@fd9f0b7700fb tmp]# errbot
The required 'git' package is either not installed or accessible in your system path. Please install 'git' or ensure it's added to your system path.
NoneType: None
[root@fd9f0b7700fb tmp]# echo $?
255
[root@fd9f0b7700fb tmp]# command -v errbot
/usr/local/bin/errbot
[root@fd9f0b7700fb tmp]# command -v git   
[root@fd9f0b7700fb tmp]# echo $?
1
[root@fd9f0b7700fb tmp]# 

```

The odd thing about this response is the `NoneType: None` output. Where is that coming from?

sijis:
@andrewthetechie There's 1 open item to address and that's the `NoneType: None` output.

andrewthetechie:
@sijis hey I just haven't had time to get back to this.

andrewthetechie:
@sijis I'm going to close this PR and take a different tactic at fixing this bug. I'll open a new PR in a bit

PR #1189, opened 2018-04-21
Add some slack functionality

TheJokersThief:
I submitted an issue/feature suggestion (https://github.com/errbotio/errbot/issues/1181) a while ago asking about interest in adding some slack functionality.

This pull request:

* Creates a new method to transform an `event` into a `Message`
* Adds ability to update/delete messages
* Adds ability to send multiple cards at once

These are things we've found useful at [HostedGraphite](https://www.hostedgraphite.com) for our ChatOps bot.

I hope this is okay but if it doesn't feel like it would fit in with the bot overall, I totally understand 😊 I've really enjoyed working with the codebase and it's made development much easier for us so I was hoping to contribute some of our home-grown improvements upstream.

Sidenote: I tried running the tox tests on OSX but they ran into an error about being unable to create a new thread and I assume codestyle tests that fail but aren't related to this PR are okay.

Issue #1181, opened 2018-03-28
Feature Feedback: Slack Backend Additions


TheJokersThief:
In order to let us help you better, please fill out the following fields as best you can:

### I am...

* [x] Suggesting a new feature

### I am running...

* Errbot version: 5.1.3
* OS version: Ubuntu 16.04
* Python version: 3.6
* Using a virtual environment: yes

### Issue description

I've been working on a ChatOps bot for work and we've made a couple improvements to the Slack backend that currently sit as their own little utils package. I've made this ticket to determine whether you'd be okay with me contributing them upstream via PR before I spend time working them into the core. So if you'd like me to put something together as a base to take suggestions, please let me know.

The specific features:
* Update a message/card
* Send multiple cards at once
* A side-effect of integrating both of the above is that sending messages returns a `Message` object which I've found useful for keeping track of responses (and for updating text)

If these don't fit the ethos or style of the bot, no worries, just thought I'd ask before I committed time to the PR :)

GenPage:
@TheJokersThief Go ahead and open a PR and we can take it from there :) Sounds interesting

TheJokersThief:
Awesome, I'll get a PR sorted and see what ye think :)

PR #1265, opened 2018-11-24
(#1181) Add unsupported delete message to the plugin API

TheJokersThief:
From #1189, adds a method for deleting messages to the base API that should be overridden by any backends that support it.

Not sure if there's anything I should do beyond the below, so please let me know if there's unit tests or anything else that needs to be added. Would also appreciate some advice/structure help if there is :)


andrewthetechie:
I really like this and see tons of places it could be useful.

Once https://github.com/errbotio/errbot/issues/1249 is fixed, can you rebase on master to get CI to pass?

PR #1267, opened 2018-11-25
Add deletion functionality to Slack backend

TheJokersThief:
From #1189, adds a method for deleting messages to the slack API.


andrewthetechie:
I like this a lot and could use this functionality myself. I'm currently just manually calling the API to do it.

Once https://github.com/errbotio/errbot/issues/1249 is fixed, can you rebase on master to get passing CI?

sijis:
This is quite interesting feature. It will require a rebase though.

PR #1268, opened 2018-11-25
Add unsupported update message to the plugin API

TheJokersThief:
From #1189, adds a method for updating messages to the base API that should be overridden by any backends that support it.

Uses the exception defined in #1265


andrewthetechie:
I like this and can see it being very useful. One comment: if a backend implemented update and then called super().update_message() wouldn't it throw an Unsupported Exception? Maybe instead of throwing an exception, log an error and do nothing?

Once #1249 is fixed, can you rebase on master to get CI to pass?

PR #1270, opened 2018-12-03
FIX msg.delayed value in XMPP backend

jlejeune:
In case of a delayed value defined in message is equal to localtime, it means there is no delay

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,7 @@ features:
 
 fixes:
 
+- backend/xmpp: include message delayed for send/recieved messages (#1270)
 - backend/xmpp: "unexpected keyword argument 'wait'" when connecting (#1507)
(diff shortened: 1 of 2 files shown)


jlejeune:
Any update ?

andrewthetechie:
This change looks good to me just reviewing the code. I don't have an XMPP server setup to functionally tests these changes though.

Anyone have XMPP setup that can validate this change?

sijis:
I'm not familiar with xmpp but what exactly is this fixing?

jlejeune:
It appears that some xmpp servers force to fill a delay parameter in each sent/received messages even if it's not necessary. This parameter should only be used to determine if message has been delivered with a delay, but it's not the case if this value is equal to localtime. That's the check I added. I was having this problem with Cisco Jabber and this patch fix the issue and don't break the other cases.

PR #1277, opened 2018-12-27
Fix check_configuration documentation

andrewthetechie:
errbot.utils.ValidationException has been moved to errbot.ValidationException as of 5.0.0. 

Update this docstring so we can update the docs as well.


GenPage:
LGTM!

PR #1282, opened 2019-01-09
Make clashes deterministic

andrewthetechie:
Fixes https://github.com/errbotio/errbot/issues/1010

### DIFF
--- a/errbot/utils.py
+++ b/errbot/utils.py
@@ -1,5 +1,4 @@
-from typing import List
-
+import collections
 import fnmatch
 import inspect
 import logging
@@ -9,6 +8,7 @@ import sys
 import time
 from platform import system
 from functools import wraps
+from typing import List
 
 
 from dulwich import porcelain
@@ -142,16 +142,16 @@ def split_string_after(str_, n):
         yield str_[start:start + n]
 
 
-def find_roots(path, file_sig='*.plug'):
+def find_roots(path: str, file_sig: str = '*.plug') -> List:
     """Collects all the paths from path recursively that contains files of type `file_sig`.
 
        :param path:
             a base path to walk from
        :param file_sig:
             the file pattern to look for
-       :return: a set of paths
+       :return: a list of paths
     """
-    roots = set()  # you can have several .plug per directory.
+    roots = list()  # you can have several .plug per directory.
     for root, dirnames, filenames in os.walk(path, followlinks=True):
         for filename in fnmatch.filter(filenames, file_sig):
             dir_to_add = os.path.dirname(os.path.join(root, filename))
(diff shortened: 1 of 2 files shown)


andrewthetechie:
Tests are failing due to #1249

sijis:
The cert has been fixed.

Would you be able to rebase and review the new errors?

Thanks.

andrewthetechie:
@sijis ty for reminding me about this pr. Should be all green now. 

I also realized my original description was a bit brief. I've tested this fairly thoroughly.

What this will do:
CorePlugins take precedence. Any plugin that duplicates the name of a core plugin botcmd will get renamed
After that, plugins are ordered alphabetically.

So, using the botcmd "status" (in core plugins) and 2 plugins "Example" and "Alpha" you'd end up with 3 status botcmds:

status (from core plugins)
alpha-status
example-status

If two plugins compete, Renaming will be ordered alphabetically. So "Example" and "Alpha" plugin both have a "foobar" botcmd:
foobar (From Alpha)
example-foobar (From Alpha)

errbot/utils.py
```diff
@@ -160,11 +162,11 @@ def find_roots(path, file_sig='*.plug'):
                     log.debug('Ignore %s.', dir_to_add)
                     break
             else:
-                roots.add(dir_to_add)
-    return roots
+                roots.append(dir_to_add)
+    return list(collections.OrderedDict.fromkeys(roots))
```
gbin: What about letting the set in the implementation and just return the sorted list at the end?

andrewthetechie: I'm not sure what you mean here, could you explain a bit more?

sijis:
I've rebased the branch so i could actually test it.

This is something that's close to merge in.

errbot/utils.py
sijis: I think he's tying to say that this maybe could be simplified by doing the mutation / sorting at the return line, instead of changing the type of `roots`. My guess is that a set ensures we have unique entries in the resolution of `roots`.

Maybe something like this would work:
```
roots = set()
for ...
return sorted(roots)
```

On another note, i never realized this but it seems like a `set()` already sorts entries?
```
In [17]: roots = {'g', 'c', 'b', 'd', 'a'}                                                                                                                                   

In [18]: roots                                                                                                                                                               
Out[18]: {'a', 'b', 'c', 'd', 'g'}

In [19]: roots = {'/g', '/c', '/b', '/d', '/a'}                                                                                                                              

In [20]: roots                                                                                                                                                               
Out[20]: {'/a', '/b', '/c', '/d', '/g'}

```

sijis: Either way, your changes does make the clashing deterministic.

PR #1284, opened 2019-01-09
Slack Backend: Add markdown support to card fields

andrewthetechie:
Addresses https://github.com/errbotio/errbot/issues/839

Slack supports markdown in three fields in a message: the pretext, the text, and the attachments. In the backend Card this maps to card.summary, card.body, and card.fields.

We split card.body to get around message size limits, making it very hard to handle markdown on that field. However, summary and attachment are not split and should be easy to handle markdown on. This PR adds that feature.

I tested this with a plugin like this:
```    @botcmd 
    def tryme(self, msg, args):  
        self.send_card(to=msg.to, summary="*Test* Card", title="Test Card Title", body="This is a body", fields=(("_Field 1_", "Test"), ("Field 2", "```code block```")))
        return
```

That produced this output via slack:
image

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -697,13 +697,27 @@ class SlackBackend(ErrBot):
         self.thread_pool.apply_async(self._slack_upload, (stream,))
         return stream
 
-    def send_card(self, card: Card):
+    def send_card(self, card: Card) -> None:
+        """
+        Sends a card via the slack api. Cards are rendered out as attachments
+        https://api.slack.com/docs/message-attachments
+
+        This card is sent with markdown enabled on the pretext and fields. Text (card.body) does not support markdown
+        :param card: (Card) the card to send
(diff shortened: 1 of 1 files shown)


andrewthetechie:
Tests are failing due to #1249

sijis:
Are there any docs that may need to be updated to include notes on this new functionality?

andrewthetechie:
@sijis I couldn't even really find any docs on Markdown in Errbot at all, besides a quick mention of "You can use Markdown!" on one page.

Is there a doc in particular you think that this could go in?

andrewthetechie:
Closing due to inactivity. I no longer work on errbot very frequently so I can't guarantee these changes will work with  recent changes

sijis:
@andrewthetechie I get ya.
@nzlosh maybe this something that could be useful on the slack-backend.

PR #1286, opened 2019-01-12
Room selector

iamkirkbater:
Hi!

Added the ability to switch rooms in the Text Backend.  It was a feature I noticed was lacking, as we have certain commands that should only be available for certain channels, or commands that have different functionality based on what channel that it's being run in.

I'm admittedly new to the errbot core, so I'm not exactly sure how to test this (I'm about to go double check to see if there's any documentation) and would not mind any pointers or improvements if you think they're necessary.

You should be able to ping me either directly here or through the gitter chat.

Thanks for looking!
Kirk


sijis:
This is interesting and good feature. I'll test and check it out in a couple of days and provide any feedback (if any).

PR #1295, opened 2019-02-04
dnspython3 is not used

greut:
~Otherwise, it's stuck on dnspython==1.15.0 when 1.16.0 is available.~

dnspython is not used.

### DIFF
--- a/errbot/bootstrap.py
+++ b/errbot/bootstrap.py
@@ -103,12 +103,12 @@ def setup_bot(backend_name: str, logger, config, restore=None) -> ErrBot:
 
         try:
             if hasattr(config, 'SENTRY_TRANSPORT') and isinstance(config.SENTRY_TRANSPORT, tuple):
-                    mod = importlib.import_module(config.SENTRY_TRANSPORT[1])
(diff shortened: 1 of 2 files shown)


sijis:
Wow. I wasn't even aware. Quick digging this is going back a few years. https://groups.google.com/forum/m/#!topic/dnspython-announce/Q0CWlESFSNg

I'll do a quick sanity check on this tomorrow.

Thank you.

andrewthetechie:
This LGTM

sijis:
So I was looking into this and it appears there are no dependencies in errbot on `dnspython` in the codebase.

greut:
@sijis aha, right. I've revised the PR accordingly.

sijis:
I wanted to go through and figure out why this was needed.

From what i can tell, it was a [requirement for xmpp](https://github.com/errbotio/errbot/commit/44b7b5c03346358b7b6177910f7170f08768d6be).

The current package supporting xmpp backend is sleekxmpp and that package already includes `dnspython` as a [package requirement](https://github.com/fritzy/SleekXMPP/blob/develop/setup.py#L155).

Needless to say, we should not need to include this as a core dependency.

Issue #1281, opened 2019-01-09
Investigate moving to Python git implementation to remove requirement of system python


andrewthetechie:
### I am...

* [ ] Suggesting a new feature

### I am running...

n/a 
### Issue description

Currently, Errbot uses system git for its cloning of plugins. This lead to #1206. 

There are python implementations of git that could be used without requiring system git. https://github.com/FriendCode/gittle is a possible candidate. It is based on https://www.dulwich.io/

sijis:
Removing that binary dependency, I think, would be a great actually.

In quickly looking, dulwitch appears more active than little.

On a separate note, I've always thought having plugins available via pypi would be awesome and avoid the whole git thing altogether. It's obvious a separate conversation.

andrewthetechie:
There's another issue about trying to come up with a solution for packaging plugins via pypi I'm looking into.

andrewthetechie:
I opened up #1296 to move us over to using dulwich rather than system git.

andrewthetechie:
Closed now that #1296 is merged

PR #1296, opened 2019-02-17
Move to python git rather than system git

andrewthetechie:
https://github.com/errbotio/errbot/issues/1281

This uses dulwich rather than shelling out to the system git binary.


sijis:
I'm going to look review this one we have a release, which hopefully is within the week.
I don't want to introduce significant changes all at once, as we haven't had a release in a while.

Issue #1307, opened 2019-03-07
Errbot doesn't treat new lines as command seperators


sinkingpoint:
In order to let us help you better, please fill out the following fields as best you can:

### I am...

* [x] Reporting a bug
* [ ] Suggesting a new feature
* [ ] Requesting help with running my bot
* [ ] Requesting help writing plugins
* [ ] Here about something else

### I am running...

* Errbot version: 

```
nobody@box:/$ errbot --version
Errbot version 5.2.0
```

* OS version: Debian 9.5
* Python version: Python 3.6.5
* Using a virtual environment: yes

### Issue description

When scanning for a command to run, errbot splits the command by spaces (https://github.com/errbotio/errbot/blob/master/errbot/core.py#L269). In the event that a message has been sent with a newline separating the command from the arguments, this fails and errbot (IMO) incorrectly identifies the command as not found.

### Steps to reproduce

Consider the following bot command:
```
@botcmd
def hello(self, msg, args):
  return 'Hello! You said: ' + args
```

When called as such:

```
@errbot hello
Line 1
```

In this case, errbot looks for commands: `["hello\nline_1", "hello\nline"]` - as it does not treat the \n as a whitespace character, it does not attempt to look for a command 'hello', and thus fails the command lookup. What I would _expect_ is that errbot correctly calls the `hello` function with the argument 'Line1' (Correctly trimming the whitespace)

### Additional info

If you have any more information, please specify it here.

sijis:
Closed via #1496.

PR #1308, opened 2019-03-07
Split the command on _all_ whitespace when matching commands

sinkingpoint:
Solves #1307 by treating all whitespace as things that could potentially
separate a command from its arguments

### DIFF
--- a/errbot/core.py
+++ b/errbot/core.py
@@ -266,7 +266,7 @@ class ErrBot(Backend, StoreMixin):
             prefixed = True
 
         text = text.strip()
-        text_split = text.split(' ')
+        text_split = text.split()
         cmd = None
         command = None
         args = ''


sijis:
A quick test this looks like the same
```
Python 3.7.2 (default, Jan 16 2019, 19:49:22) 
[GCC 8.2.1 20181215 (Red Hat 8.2.1-6)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> str = 'this is a test.'
>>> str1 = str.split(' ')
>>> str2 = str.split()
>>> str1 == str2
True
>>> str1
['this', 'is', 'a', 'test.']
>>> str2
['this', 'is', 'a', 'test.']
>>> 
```

Then i tried with a newline:
```
>>> str = 'this is a test\n of something new.'
>>> str
'this is a test\n of something new.'
>>> str.split()
['this', 'is', 'a', 'test', 'of', 'something', 'new.']
>>> str.split(" ")
['this', 'is', 'a', 'test\n', 'of', 'something', 'new.']
```

Lastly, it made me think to try a few other characters
```
>>> str = 'this is a test\n of something new\r\n whatever dude\t yipeeee.'
>>> str.split()
['this', 'is', 'a', 'test', 'of', 'something', 'new', 'whatever', 'dude', 'yipeeee.']
>>> str.split(" ")
['this', 'is', 'a', 'test\n', 'of', 'something', 'new\r\n', 'whatever', 'dude\t', 'yipeeee.']
>>> 
```

Using `split()` w/out passing an empty string seems to split things, probably as intended.

My only real concern, is do we have any adverse effects from doing this, as its likely we have other code that then tries to handle the other characters (eg \n, \r, \t, etc..).

sinkingpoint:
@sijis I've had a cursory glance at the code and can't see any extra handling of whitespace. What are the next steps here?

sijis:
I'm looking at merging a few things in the next few days (more likely know weekend).

This is on my list.

gbin:
I also fear some breakages here.

sijis:
I just thought of something.
Using `split(' ')` will do what's needed but the after effect is we are left with extra characters. Maybe we then re-filter through `strip()`

```
>>> str = 'this is a test\n of something new\r\n whatever dude\t yipeeee.'
>>> [i.strip() for i in str.split(' ')]
['this', 'is', 'a', 'test', 'of', 'something', 'new', 'whatever', 'dude', 'yipeeee.']
```

oddly enough, the result is the same output as just using `str.split()` initially.
```
>>> [i.strip() for i in str.split(' ')] == str.split()
True
```

sijis:
In looking over this again, i do not think this will cause much of an impact as I originally thought.
@gbin Any objections in merging this in?b

PR #1314, opened 2019-04-02
test: Add security linter with bandit

sijis:
This is adding a security linter named [bandit](https://pypi.org/project/bandit/) to our testing process.

I thought it would be educational to see what potential security issues we may be introducing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ matrix:
       dist: xenial
       sudo: true
       env: TOXENV=codestyle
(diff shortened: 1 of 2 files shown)


gbin:
This is pretty cool can we leave it as warning so we can spead the load and fix those?

sijis:
Yeah, initially i would just ignore the output and slowly work through any issues it brings up.

sijis:
I set the command to ignore the output.

PR #1317, opened 2019-04-02
fix: Remove waffle

sijis:
Service is being discontinued


sijis:
Its a simple doc update, so merging.

PR #1318, opened 2019-04-08
fix double threading in slack backend if DIVERT_TO_THREAD is used

maxim-nazarenko:
Current code tries to start a new thread (`if threaded:` statement) regardless of the current state.
We should check if we are in a thread before setting `response.parent`.

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -892,12 +892,12 @@ class SlackBackend(ErrBot):
     def build_reply(self, msg, text=None, private=False, threaded=False):
         response = self.build_message(text)
 
-        if threaded:
-            response.parent = msg
-
-        elif 'thread_ts' in msg.extras['slack_event']:
+        if 'thread_ts' in msg.extras['slack_event']:
             # If we reply to a threaded message, keep it in the thread.
             response.extras['thread_ts'] = msg.extras['slack_event']['thread_ts']
+        elif threaded:
+            # otherwise check if we should start a new thread
+            response.parent = msg
 
         response.frm = self.bot_identifier
         if private:


lawrence-law:
@gbin and @sijis - any chance of reviewing this one? I've also experienced the same issue and identified the same bugfix.

iamkirkbater:
Does this fix the issue where if you enter a second command in the first command's thread it will split out the second command into a top level thread and then add the processing to that thread?

If so: yeessssssssss.

maxim-nazarenko:
> Does this fix the issue where if you enter a second command in the first command's thread it will split out the second command into a top level thread and then add the processing to that thread?
> 
> If so: yeessssssssss.

yes, it's that one

sijis:
Maybe it's late but I'm not understanding what this is fixing. Is there a way you could post plugin code (or steps) that reproduces the problem? It would be helpful. Thanks.

maxim-nazarenko:
> Maybe it's late but I'm not understanding what this is fixing. Is there a way you could post plugin code (or steps) that reproduces the problem? It would be helpful. Thanks.

Steps to reproduce:
1) Configure the Slack backend
2) update the config.py with "DIVERT_TO_THREAD = ('help', 'about', 'status')"
3) issue the `help` command (the response should go to a new thread)
3) in the new thread issue `help` command again.

Expected behavior: help text appears in the same thread where you issued the command
Actual behavior: the response goes to the main thread aka room (the room where you issued the first `help` command)

lawrence-law:
> Maybe it's late but I'm not understanding what this is fixing. Is there a way you could post plugin code (or steps) that reproduces the problem? It would be helpful. Thanks.

@sijis - thanks for responding. Here's how you can reproduce the problem:

1. Configure a particular command as `DIVERT_TO_THREAD` in Errbot config
2. Post any message in a Slack room
3. Start a message thread from the above message and use the command that was configured in step 1

What then happens is that the command that you sent in step 3 somehow appears in the chatroom as such (my command is `!ping`):

image

The bot actually does manage to respond in the thread, but you actually have to change context (e.g. close the thread window, change to another chatroom and back in) before you can see the bot's reply inside the thread.

iamkirkbater:
@sijis - One of the potential use cases for this threaded behavior is flows.  We have a command in our bot in which kicks off a flow and then has a `!verify` or `!quit` step, which allows us to stop and have to confirm the action the original command is taking.  Being able to have those flows threaded is just cleaner.

sijis:
Thanks for explaining this to me. This makes sense now.

PR #1321, opened 2019-04-09
docs: Auto-create CHANGES from git

sijis:
This dynamically creates `CHANGES.rst` using [gitchangelog](https://github.com/vaab/gitchangelog).

Since this file is shared with the website, a generated output looks like below:

errbot-auto-changelog-01


sheluchin:
It would be nice to use something like https://github.com/c4milo/github-release as well to give https://github.com/errbotio/errbot/releases a bit more meaning and to include commit hashes next to each change. Maybe gitchangelog can be configured to include the commit hash? I think it would make it easier to understand each change.

PR #1331, opened 2019-04-25
Doc updates

garmann:
While i was reading the documentation, i noticed some things to improve imho. I compiled the doc with the instructions from the sphinx/make page and it looked nice. Let me know if i should change my commits. I did all changes in the browser, so there are single commits.


sijis:
There errors are related to something outside of this PR.

PR #1333, opened 2019-04-29
Increase Slack message size limit.  Fixes #1325

### LINKED ISSUE #1325: 5.2+: Slack messages are being split too early
This issue was introduced in version 5.2. Version 5.1.3 and prior work fine.

The offending change seems to be within this pull request:
https://github.com/errbotio/errbot/pull/1118

The message limit is set to "4096" and you cannot override this behavior because it was programmed to select the minimum of your setting or that number, meaning you can set a smaller limit but not a higher one.

However, [Slack supports up to 40,000 characters since August 2018](https://api.slack.com/changelog/2018-04-truncating-really-long-messages). If not by default, then it should at least be possible to configure it so that messages are not split before they reach this length.

This is truly a very annoying bug, as it takes a nicely formatted single message that gets trimmed with a "see more" button (which is great for our use case) to 5-10+ messages.

nzlosh:
As per the information provided in issue #1325 the slack message limit has been increased from the empirical 4000 characters to the documented 40000 (https://api.slack.com/changelog/2018-04-truncating-really-long-messages).  The original limit of 4000 can always to achieved by setting the `MESSAGE_SIZE_LIMIT=4000` in the bot configuration.

Documentation about this feature has also been added.

### DIFF
--- a/docs/user_guide/configuration/slack.rst
+++ b/docs/user_guide/configuration/slack.rst
@@ -76,3 +76,13 @@ then you can set `CHATROOM_PRESENCE` to a list of channels and groups to join.
 
     You may leave the value for `CHATROOM_FN` at its default
     as it is ignored by this backend.
+
+
+Message size limit
+------------------
+
+As of the 12th August 2018 the Slack API has a message limit size of 40,000 characters.  Messages 
+larger than 40,000 will be truncated by Slack's API.  Errbot includes the functionality to split 
+messages larger than 40,000 characters into multiple parts.  To reduce the message limit size, set the 
+`MESSAGE_SIZE_LIMIT` variable in the configuration file.  Errbot will use the smallest value between 
+the default 40,000 and `MESSAGE_SIZE_LIMIT`.
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -36,8 +36,9 @@ except ImportError:
 # token matching this regex.
 SLACK_CLIENT_CHANNEL_HYPERLINK = re.compile(r'^<#(?P<id>(C|G)[0-9A-Z]+)>$')
 
-# Empirically determined message size limit.
-SLACK_MESSAGE_LIMIT = 4096
+# https://api.slack.com/changelog/2018-04-truncating-really-long-messages
+# On August 12, 2018 we started truncating messages containing more than 40,000 characters.
+SLACK_MESSAGE_LIMIT = 40000
 
 USER_IS_BOT_HELPTEXT = (
     "Connected to Slack using a bot account, which cannot manage "


sijis:
I believe the default should continue to be 4096 and have a config option that will allow someone to set that value to something higher.

nzlosh:
I don't know the reason why the original author used `SLACK_MESSAGE_LIMIT` as a hard limit, allowing it to be override with lower values by setting `MESSAGE_SIZE_LIMIT`, rather than allowing `MESSAGE_SIZE_LIMIT` to any arbitrary size.

Perhaps to avoid negative user experience with truncated messages due to misconfiguration?  This may not be a problem any more since the limit is actually documented on Slack's website.

It's possible to keep the same behaviour as the slack backend has now by adding a third variable `HARD_MESSAGE_LIMIT` set to 40000.  Keep `SLACK_MESSAGE_LIMIT` at 4000 and allow `MESSAGE_SIZE_LIMIT` to override the value to any value lower than hard message limit.  Would this be a better option?

sijis:
By avoiding the negative experience, i was referring to folks potentially working around the current limit set at 4000 in their plugins.

My quick thought was doing the following:
* Use the `MESSAGE_SIZE_LIMIT` config (https://github.com/errbotio/errbot/blob/master/errbot/config-template.py#L334).
* If that value is set, use it, if not, its defaulted to existing 4000 value.

This should allow folks to adjust it to the slack limit, as they chose.

nzlosh:
The problem is more complicated than I initially thought.  There are actually 3 variables at play.  `bootstrap.py` sets `MESSAGE_SIZE_LIMIT` to 10000.
https://github.com/errbotio/errbot/blob/2d8fe13de42fed20b9472a63222e02e4ba63bedd/errbot/bootstrap.py#L47
This makes reasoning about user supplied value versus bootstrap value versus backend default value more difficult because the backend code is unable to know if `config.MESSAGE_SIZE_LIMIT` is in fact user supplied or set by the bootstrap code.

It seems to me the backend message size limit is the true hard upper limit that a message can be.  Bootstrap is used if the user configuration is not set and is currently 10000 (the comment suggests it's a hipchat limit).

As far as I can tell each backend implements slightly different limit strategies.
- irc and telegram overwrite the message limit with the backend value.
https://github.com/errbotio/errbot/blob/a1543e13ea7ca6b6fe27f283a8abb2901e3566f4/errbot/backends/telegram_messenger.py#L179
https://github.com/errbotio/errbot/blob/8c3d74995e316a14870527901ce89f2733691c53/errbot/backends/irc.py#L663
- gitter overwrites the message limit only if it's larger than the backend value.
https://github.com/errbotio/err-backend-gitter/blob/14a428bdd0597b473605264c822855da979bd916/gitter.py#L293
- mattermost and slack use the smaller value between the message limit and backend value.
https://github.com/Vaelor/errbot-mattermost-backend/blob/1c3e6b603c9540346bce32ae38e07be93c5c914b/mattermost.py#L359
https://github.com/errbotio/errbot/blob/03ec94afcbde91ae40bc284f9ff0c1f800d6b41a/errbot/backends/slack.py#L652
- discord sets its own default value and doesn't use message limit at all.
https://github.com/gbin/err-backend-discord/blob/57c02cdc99ece946dd18bfa332cf4e2480f4a555/discordb.py#L253
- rocketchat doesn't set any limits.

I'd like to propose that the bootstrap code doesn't set a default value and each backend + `core.py` https://github.com/errbotio/errbot/blob/fe64c2514cdd274b7ee14c80ea47598f88db70e8/errbot/core.py#L159 use their backend upper limit when the user doesn't configure a value.  If this seems acceptable, I'll work on a PR.

nzlosh:
The solution I'd like to propose is a new method `set_message_size_limit` in `Errbot` class which can be overridden by each backend that inherits from this class.

If the method isn't overridden, the default message size of 10000 is used, as is currently the situation when a user doesn't set `MESSAGE_SIZE_LIMIT`.

In the case a user sets this variable, it is used to split messages that are larger than this limit.

In the case a user sets the variable higher than the backends hard limit, a warning is logged to notify the user that message delivery issues may be encountered.

The patch allows the default value of 4096 chars by default for Slack, while setting a hard limit of 40000 characters for warning purposes.  User defined values take precedence of the backend limit.

This change will potentially impact 3rd party backends and they will need to be updated to use this new message size facility.  I have already identified a few of them which I will offer patches to make them compatible.

@sijis @gbin Is this an acceptable solution for you?

sijis:
That sounds reasonable to me.

The changes should give some sane default, in your case 1000, if the method isn't added on a 3rd party backend.

nzlosh:
Thanks for the review.

I agree `message_size_limit` appears to be a property. The way I've implemented the call in `ErrBot.__init__` to override the parent variables by the backend's variables without requiring the backend auther to explicitly call the `message_size_limit` function would break if the method was converted to a decorated property.

Would wrapping the configuration variable as a decorated property be worth while and acceptable compromise in your opinion?

```
    @property
    def message_size_limit(self):
        return self.bot_config.MESSAGE_SIZE_LIMIT
```

sijis:
Ya. I figured it would be a decorated property.

nzlosh:
Any thoughts on merging this PR for the next point release?

PR #1336, opened 2019-05-14
Remove empty "text" body on Slack send_card

ascandella:
When "text" is provided, Slack uses this empty message as the body when
sharing (cross-posting) messages from one room to another.

If you remove this "text" field, it uses the content of the attachments.

## Before the patch

Original message:
image

Share dialog (blank preview):
image

Cross-posted:
image

## After the patch

Original message:
image

Share dialog:
image

Cross-posted:
image

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -748,7 +748,6 @@ class SlackBackend(ErrBot):
                 attachment['footer'] = f'{footer} [{i + 1}/{part_count}]'
             attachment['text'] = parts[i]
             data = {
-                'text': ' ',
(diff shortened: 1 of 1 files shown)


sijis:
What does this look like with the patch?

sectioneight:
@sijis sorry, that would have been very helpful to include originally. I've updated the PR description to show the issue before and after.

sijis:
Thank you that was extremely helpful.

I don't see an issue with this. I just need to look into why this field was added originally and set to ` `. Probably not an issue I'd just like to double-check.

PR #1337, opened 2019-05-15
Implementing support for updating topics in XMPP-groupchats

zaughon:
I've added support for updating the topic in XMPP groupchats, and it is working for me, but i'm not 100% sure whether the code might cause issues for other backends, as i don't have any other backends to test with.   
The tests are passing, aside from a webhooks test that sometimes passes, sometimes fails. I didn't touch that file at all though. 

`errbot/backends/base.py`:
- `class Message` can now contain a `msubject`-string

`errbot/backends/xmpp.py`:
- Changed `send_message` to include msubject. Will either be None or string
- Fixed `@topic.setter` so it sends a message to update the topic

`errbot/botplugin.py`
- `send()` now accepts msubject bool. If True, `text` will be used as new topic by `send()` in `errbot/core.py`

`errbot/core.py`
- `send()` now accepts msubject bool. If True, `text` will be used as new topic

`errbot/core_plugins/chatRoom.py`
- New topic will now be `" ".join(args[1:])`. Otherwise only the first word of the new topic was used

### DIFF
--- a/errbot/backends/base.py
+++ b/errbot/backends/base.py
@@ -263,7 +263,7 @@ class Message(object):
     def clone(self):
         return Message(body=self._body, frm=self._from, to=self._to, parent=self._parent,
                        delayed=self._delayed, partial=self._partial, extras=self._extras, flow=self._flow,
-                       msubject = self._msubject)
+                       msubject=self._msubject)
 
     @property
     def to(self) -> Identifier:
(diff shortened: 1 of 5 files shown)


gbin:
As for all the features that change Errbot's API, is there at least another backend capable of this?
If not, you might want to implement it only for the XMPP backend.

also for the name msubject with m representing "message", it is better named "subject" because it is already a member of the message class (smurf anti-pattern)

sijis:
Closing.
* Older PR
* various conflicts

PR #1341, opened 2019-05-21
add clearer error message for slackclient>=2.0

maxim-nazarenko:


sijis:
I think this is a good thing. As I ponder this, this change would still require a new version to be noticed (for folks using pypi).

If that's the case, we are better off pinning the version in `setup.py`. then we may not need this message at all.

maxim-nazarenko:
> I think this is a good thing. As I ponder this, this change would still require a new version to be noticed (for folks using pypi).
> 
> If that's the case, we are better off pinning the version in `setup.py`. then we may not need this message at all.

@sijis goog point. Probably, pinning the version would be sufficient.

sijis:
This is not needed any longer. We released 6.1.0 which includes the pinned slackclient version.

PR #1342, opened 2019-05-24
chore: Move all tox environments to use py37

sijis:
This makes tox use python 3.7.


sijis:
weird. This works locally for me on py 3.7.3. :/

PR #1343, opened 2019-06-11
Prepare for 6.1

sijis:
I'm merging in `master` branch changes onto `6.0` branch.

This is to prepare for a 6.1 release.

It would include the following changes:
- docs:
    - various documentation fixes
    - update storage backend
    - rename Spark to Webex Teams
- fixes:
    - `errbot -l` error
    - fix: slack backend eror by pinning slackclient to a pre 2.0 version
- feature:
    - use python git implementation instead of system git binary

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,7 +9,6 @@ features:
 - Adds ability to go back to a previous room
 - Pass telegram message id to the callback
 
-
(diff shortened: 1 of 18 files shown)


sijis:
Merging. We have that slackclient pin which is very important to do.

PR #1345, opened 2019-06-16
docs: Add 6.1.0 changelog

sijis:
This adds the changelog for the 6.1.0 release.

Depends on #1343.


sijis:
Merging. We have that slackclient pin which is very important to do.

Issue #1368, opened 2019-07-23
Blacklisted plugins causes AttributeError on startup


gaggle:
In order to let us help you better, please fill out the following fields as best you can:

### I am...

* [x] Reporting a bug

### I am running...

* Errbot version: errbot==6.1.1
* OS version: Ubuntu 18.04
* Python version: Python 3.6.8
* Using a virtual environment: yes

### Issue description

Blacklisting plugins results in exceptions during bot startup, the exception is highly visible in bot log output **and** the error is emitted by the bot in Slack which is highly confusing.

The exception is:
`.../lib/python3.6/site-packages/errbot/plugin_manager.py`:
```
                if self.is_plugin_blacklisted(name):
                    errors += f'Notice: {plugin.name} is blacklisted, ' \
                              f'use {self.bot.prefix}plugin unblacklist {name} to unblacklist it.\n'
                    continue
```

But there is no `self.bot`, so `self.bot.prefix` throws exception, which causes the bot to emit errors.

Quick glance at git history looks to me like this was introduced May 28th, 2018 in commit [21c282356892596cf18b356cfe2435c4abe41b70](https://github.com/errbotio/errbot/commit/21c282356892596cf18b356cfe2435c4abe41b70#diff-86fb3ee3298c4b1416b1da174131201cL140) where `self.bot` property was removed from `plugin_manager.py` (L140) as part of "removed circular dependency between pm and errbot "

### Steps to reproduce

1. Create `backup.py` with content:
```
## This file is not executable on its own. use errbot -r FILE to restore your bot.

bot.plugin_manager["configs"] = {}
bot.plugin_manager["bl_plugins"] = ['Flows']
if "installed_repos" in bot.repo_manager:
  for repo in bot.repo_manager["installed_repos"]:
    log.error(bot.repo_manager.install_repo(repo))
log.info("Restoring plugins data.")
bot.plugin_manager.update_plugin_places(bot.repo_manager.get_all_repos_paths())
pobj = bot.plugin_manager.plugins["Example"]
pobj.init_storage()
pobj.close_storage()
pobj = bot.plugin_manager.plugins["Plugins"]
pobj.init_storage()
pobj.close_storage()
pobj = bot.plugin_manager.plugins["CommandNotFoundFilter"]
pobj.init_storage()
pobj.close_storage()
pobj = bot.plugin_manager.plugins["Help"]
pobj.init_storage()
pobj.close_storage()
pobj = bot.plugin_manager.plugins["Backup"]
[...]

PR #1369, opened 2019-07-23
Remove self.bot.prefix

gaggle:
`self.bot` doesn't exist

This is a simple proposal for #1368. Maybe there's a better way to do it?

### DIFF
--- a/errbot/plugin_manager.py
+++ b/errbot/plugin_manager.py
@@ -358,7 +358,7 @@ class BotPluginManager(StoreMixin):
             try:
                 if self.is_plugin_blacklisted(name):
                     errors += f'Notice: {plugin.name} is blacklisted, ' \
-                              f'use plugin unblacklist {name} to unblacklist it.\n'
+                              f'use "{self.plugins["Help"]._bot.prefix}plugin unblacklist {name}" to unblacklist it.\n'
                     continue
                 if not plugin.is_activated:
                     log.info('Activate plugin: %s.', name)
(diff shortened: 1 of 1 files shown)


gaggle:
Not immediately recognizing the failure, if anyone wants to guide me a bit I'm happy to iterate on the solution

sijis:
I think the prefix should be included in the output.

I don't know the path offhand but finding an example usage of self.bot.prefix should indicate how you can get acces to the attributes.

sijis:
I was finally able to poke around a little.
It seems that the way to get the same behaviour is doing: `self.plugins['Help']._bot.prefix`. This is not ideal but it accomplishes the same intended behaviour.

errbot/plugin_manager.py
```diff
@@ -356,7 +356,7 @@ def activate_non_started_plugins(self):
             try:
                 if self.is_plugin_blacklisted(name):
                     errors += f'Notice: {plugin.name} is blacklisted, ' \
-                              f'use {self.bot.prefix}plugin unblacklist {name} to unblacklist it.\n'
+                              f'use plugin unblacklist {name} to unblacklist it.\n'
```
sijis: Based on earlier findings, this should work.
```suggestion
                              f'use "{self.plugins["Help"]._bot.prefix}plugin unblacklist {name}" to unblacklist it.\n'
```

gaggle:
Sorry @sijis I forgot this PR, thanks for your suggestion! I've implemented it, and rebased to latest upstream master. I get test-errors in `commands_test` but I'm not seeing their relation to my change. But open for feedback..

craigds:
Will there be a release for this fix soon?

sijis:
@craigds @gaggle The latest 6.1.7 release should include this fix.

PR #1364, opened 2019-07-21
Google Plus does not exist anymore

fernand0:
Suggesting change to show gitter community

PR #1374, opened 2019-08-19
Fix: Remove google+ references

sijis:
Google+ no longer exists, so we should remove references for it.

This should superseed #1364.


sjsajj:
the 'open chat' button seems to be missing.

sijis:
Bottom right corner on the website.

sjsajj:
The button was blocked on the corporate network. After opening the website on my home network I see it.

PR #1386, opened 2019-09-12
make the TestBot start() timeout parameterized

neunkasulle:
For my current CI setup a 2 second timeout is too short. So a parameter
would be helpful to be able to adjust that


sijis:
Thanks!

Issue #1383, opened 2019-09-01
`/proc/1/cgroup`: errbot is mandatorily checking for this file for non-docker builds too


RO-29:
In order to let us help you better, please fill out the following fields as best you can:

### I am...

* [ ✅  ] Reporting a bug
* [ ] Suggesting a new feature
* [ ] Requesting help with running my bot
* [ ] Requesting help writing plugins
* [ ] Here about something else

### I am running...

* Errbot version: 5.1.3
* OS version: macOS Sierra: 10.12.6
* Python version: 3.6.0
* Using a virtual environment: no

### Issue description

errbot text mode is throwing up trying to find `/proc/1/cgroup` even for non docker builds
```
in is_docker
    with open('/proc/1/cgroup') as d:
FileNotFoundError: [Errno 2] No such file or directory: '/proc/1/cgroup'
```
### Steps to reproduce

jut run errbot -T

RO-29:
I would like to submit a PR for this to fix this issue

sijis:
Could you try with the latest version of errbot or master branch?

RO-29:
The bug still exists in the master 

```
Some plugins failed to start during bot startup:
 <class 'FileNotFoundError'>: [Errno 2] No such file or directory: '/proc/1/cgroup'
  File "/Users/rohit/.pyenv/versions/3.6.0/lib/python3.6/site-packages/errbot/plugin_manager.py", line 82, in install_packages
    elif is_docker():
  File "/Users/rohit/.pyenv/versions/3.6.0/lib/python3.6/site-packages/errbot/plugin_manager.py", line 67, in is_docker
    with open('/proc/1/cgroup') as d:
 <class 'FileNotFoundError'>: [Errno 2] No such file or directory: '/proc/1/cgroup'
  File "/Users/rohit/.pyenv/versions/3.6.0/lib/python3.6/site-packages/errbot/plugin_manager.py", line 82, in install_packages
    elif is_docker():
  File "/Users/rohit/.pyenv/versions/3.6.0/lib/python3.6/site-packages/errbot/plugin_manager.py", line 67, in is_docker
    with open('/proc/1/cgroup') as d:
```

The faulty code is here:

https://github.com/errbotio/errbot/blob/master/errbot/plugin_manager.py#L66-L68
```
def is_docker():
        with open('/proc/1/cgroup') as d:
            return 'docker' in d.read()
```
Can be modified as 
```
def is_docker():
[...]

PR #1389, opened 2019-10-02
errbot/plugin_manager: check for /proc/1/cgroup only if path exists

RO-29:
* errbot is mandatorily checking for this file for non-docker builds too
* Fixes https://github.com/errbotio/errbot/issues/1383


sijis:
Thanks!

PR #1400, opened 2019-12-16
Feature: Version included in website

sijis:
This simplifies our domain structure and includes a way for us to use the json in versionchecker plugin.

The version would be available in: https://errbot.io/versions.json

We'd be able to remove part of this change https://github.com/errbotio/errbot/commit/8f20371eccffeaa4030fd56ee8f11e0ca7f7f375.

<strike>The only thing left is to update vcheck plugin to include some of the logic from the [version server](https://github.com/errbotio/errbot-version-server/blob/master/errbot/version.go).</strike>


sijis:
@gbin These changes require that the RTD setup is able to build upon merges to master (or other branches).

sijis:
Change works!
```
curl -L https://errbot.io/versions.json
{
  "python2": "4.2.2",
  "python3": "6.1.2"
}
```

PR #1403, opened 2019-12-19
Feature: Sphinx serve repos.json

sijis:
Doing some cleanup and preparing to setup sphinx to auto-build the `repos.json` file.


sijis:
While testing the build process via RTD, the build takes too long to be useful as the build timeouts.
RTD does provide a generous 900 seconds limit.

Unfortunately, while we cannot use RTD to automatically build the `repos.json` file, we can still have the file hosted on the website.

I'll update this PR to reflect that.

darwinzavalaroque172-stack (approved):

PR #1404, opened 2020-01-11
A typo and name with .py does not work

fernand0:
* I think the dot at the end of `Name` is not needed (nor useful)
* If you put the `.py` in the `Module` in `[Core]` there is an error because the plugin loader adds the .py extension.


sijis:
Thanks!

PR #1405, opened 2020-01-11
description seems to be read from bootcmd definition

fernand0:
Since description seems to be read from `@bootcmd` definition and this is not easy to see in the documentation (even the examples don't use it, see https://github.com/errbotio/errbot/blob/master/docs/user_guide/flow_development/basics.rst ) maybe it is better to change this code to not produce an error in the command `flows list` when no description is available.


sijis:
Looks good. An example output is
```
[@sijis ➡ @errbot] >>> !flows list
• poll_setup No description
```

sijis:
Thank you for the fix.

Issue #1000, opened 2017-05-05
Command not found only overwritten sometimes


crimsonknave:
In order to let us help you better, please fill out the following fields as best you can:

### I am...

* [ ] Reporting a bug

### I am running...

* Errbot version: 5.0.0
* OS version: Linux
* Python version: 3.5.3
* Using a virtual environment: yes

### Issue description

I'm attempting to override the default command not found behavior. It works... sometimes... Without changing any config I can start errbot multiple times and get the default and my behavior. It does not seem to be swapping on a given run.

### Steps to reproduce

* Define a new cnf entry using ```@cmdfilter(catch_unprocessed=True)```
* Start errbot
* Enter a non existent command (you will get the new or the old)
* Restart errbot
* Enter a non existent command (you will get the new or the old)
* Repeat about 4-5 times and you should have gotten both.

crimsonknave:
This appears to be due to core.py pulling the first command filter with catch_unprocessed true and self.command_filters isn't in a consistent order run to run.

gbin:
Can you try #1005 and tell me if it is ok like this ? You'll have to exclude the CommandNotFoundFilter plugin from the CORE_PLUGINS setting in config.py if you don't want both behavior to happen at the same time.

crimsonknave:
@gbin For my specific purposes disabling the plugin is all I need. I just want to replace the default behavior. 

It would, however, be nice to have a blacklist setting instead of having to specifically list out all of the plugins I want. First, I've now got to figure out all the core plugins that exist and add them. Second, I'll never get any new core plugins unless I notice that they exist and then add them to the whitelist.

PR #1415, opened 2020-02-18
Document over-ride CommandNotFoundFilter.

nzlosh:
Document how to prevent CommandNotFoundFilter from executing before other command filters.


sijis:
Document issue #1000.

PR #1367, opened 2019-07-21
Basic support for slackclient v2

gbin:
It adds a LOT of noise in Errbot as the API is way lower level than the previous version (this is basically a pure socket to Slack + json parsing).

PR #1416, opened 2020-02-22
Feature: Slack RTM Backend

sijis:
This leverages the original work from https://github.com/errbotio/errbot/pull/1367.

I refactored the commits so the backend is registered under a different name `SlackRTM`.

The goal is to be able to release this, although not fully featured with the intent to replace the existing Slack backend (using the older api) once its feature parity complete.

### DIFF
--- a/errbot/backends/slack_rtm.plug
+++ b/errbot/backends/slack_rtm.plug
@@ -1,0 +1,6 @@
+[Core]
+Name = SlackRTM
+Module = slack_rtm
+
+[Documentation]
+Description = This is the Slack RTM backend for Errbot.
--- a/errbot/backends/slack_rtm.py
+++ b/errbot/backends/slack_rtm.py
@@ -7,6 +7,7 @@ import re
 import sys
 import pprint
 from functools import lru_cache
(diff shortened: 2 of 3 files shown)


nzlosh:
Sorry arrive late on this one, I have a question around the choice of `SlackRTM`.  I wonder how obvious this name will be for users compared to calling the new plugin `Slack2`.  What are your thoughts around this and is it too late for a name change?

sijis:
> Sorry for arriving late on this one. I have a question around the choice of `SlackRTM`. I wonder how obvious this name will be for users choosing which Slack backend to install compared to `Slack2`. I think Slack2 is more obvious to people that it's a newer version than `Slack`. What are your thoughts around this and is it too late for a name change?

I thought the name was representative of the change. I figured it would eventually be the default and the `Slack` one would be removed.

The name can be changed. I'm not necessary against a rename. Giving the user a better idea that its a newer version can make it more obvious.

nzlosh:
OK, no need to worry about a name change.  I didn't realise it would eventually be changed to Slack to replace the existing backend.

PR #1418, opened 2020-02-23
WIP - Markdown v3 support

nzlosh:
Replace deprecated methods and arguments with v3 methods and arguments.

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -68,7 +68,7 @@ def slack_markdown_converter(compact_output=False):
     """
     enable_format('imtext', IMTEXT_CHRS, borders=not compact_output)
     md = Markdown(output_format='imtext', extensions=[ExtraExtension(), AnsiExtension()])
-    md.preprocessors['LinkPreProcessor'] = LinkPreProcessor(md)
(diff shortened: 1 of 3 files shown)


nzlosh:
@sijis tox passes with this but I'm not entirely clear of the priority values given for the markdown extensions.

sijis:
> @sijis tox passes with this but I'm not entirely clear on the priority values given for the markdown extensions. They seem to work based on Slack and Discord backends.

I don't understand the priorities fully myself.

There is a `!render_test` command that can be used to test the markdown output in those backends.

nzlosh:
The latest change produces a render test that works for Slack but Mattermost and Discord continue to render `</br>` in the _Markdown extra (table multiline)_ section.

sijis:
Would you be able to post an image, so I could visualize the difference?

nzlosh:
Here's the rendering

**Discord** (Ignore the blue highlighted text, that was me copy/pasting)
discord_render

**Mattermost**
mattermost_render

**Slack**
slack_render

sijis:
I see what you mean now. I'm finding the markdown library confusing.

The options I keep thinking about are:
* using `md.X.add()` instead of `md.X.register()`
* Removing the actual `br` definition and see

PR #1423, opened 2020-04-12
Fixing Deprecation Warning

dvaerum:
When using pytest I get the following Deprecation Warning
```
DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
    from collections import MutableMapping
```
So thought I would fix it, hope you will accept it 😃


sijis:
Thank you

Issue #1431, opened 2020-06-18
Fix Slack backend's use of deprecated method IM.OPEN


gkspranger:
* [x ] Reporting a bug

* Errbot version: 6.1.4
* OS version: CentOS7
* Python version: 3.7.7
* Using a virtual environment: yes
* Backend: Slack

I am installing the Slack backend via `errbot[slack]` .. I am getting DEBUGs in my logs and sometimes outright errors about the used of a deprecated method `IM.OPEN`

https://github.com/errbotio/errbot/blob/master/errbot/backends/slack.py#L592-L603

https://api.slack.com/methods/im.open

```2020-06-18T13:27:31.66-0400 [APP/PROC/WEB/0] ERR 2020-06-18 17:27:31,662 DEBUG    errbot.backends.slack     Tried to build an identifier from '<@abc123>' but got exception: Slack API call to im.open failed: method_deprecated```

It appears we need to use this method, but not sure how to implement/test :: https://api.slack.com/methods/conversations.open

thanks,
greg

jcastro-sfdc:
#1432 Have a look at this

![image](https://user-images.githubusercontent.com/48135245/86162330-9abefa00-badc-11ea-9f44-520c17d4550f.png)

gkspranger:
thanks @jcastro-sfdc for looking into this -- i appreciate the work and the PR

cappetta:
Just hit this.  Thanks for fixing @jcastro-sfdc

olanystrom:
Works great, when is this release coming? 6.1.4 was a long time ago

sijis:
6.1.5 was release.

PR #1432, opened 2020-06-30
fixing the im.open deprecated call 

jcastro-sfdc:
using conversations.open, this will work for all apps/bots created after June 10, 2020
#1431

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -593,7 +593,7 @@ class SlackBackend(ErrBot):
     def get_im_channel(self, id_):
         """Open a direct message channel to a user"""
         try:
-            response = self.api_call('im.open', data={'user': id_})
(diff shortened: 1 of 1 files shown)


gkspranger:
👍 👍 👍

sijis:
Will this also work with older apps?

jcastro-sfdc:
@sijis Yes it will.  However,the deprecated messages will stop working in February 2021 all together.  Errbot also uses the group and channel methods which should be changed to "conversations" method too.

from the Slack website:

"This method is deprecated. It will stop functioning in February 2021 and will not work with newly created apps after June 10th, 2020."

jcastro-sfdc:
im.open and conversations.open will work with all apps created before June 10, 2020 but you should still change your im, group and channel methods to conversations.  :). my last reply was a bit messy

PR #1465, opened 2020-10-16
Chore: minor code cleanup

sijis:
Minor code cleanup

* Remove unused imports
* Fix lazy logging usages
* Remove old object inheritance usage


errbot/backends/hipchat.py
```diff
@@ -590,7 +590,7 @@ def _hipchat_upload(self, stream):
                 log.error('Request text: %s.', resp.text)
                 stream.error()
         except Exception:
-            log.exception(f'Upload of {stream.name} to {stream.identifier.channelname} failed.')
```
auvipy, line 593: why f string is removed?

auvipy (approved):
overall looks good to me

errbot/backends/hipchat.py
sijis: I don't recall the specific reason but i'm sure it was because the linter complained about lazy logging.

I really appreciate the review!

PR #1451, opened 2020-09-07
Support for slack events

AgarFu:
This PR is reusing most of the classes from the SlackRTM modules, there is no documentation nor tests in this PR and the main class can be refactored in order to be able to leverage most methods from `SlackRTMBackend` class.

Please let me know if this is something that you're interested in incorporate to the project I can add the missing bits, it is now in a "works for me" status.

PR #1472, opened 2020-10-26
Add Slack Events backend

duhow:
This is a copy of Slack RTM adapted to work with **Slack Events v2**.
It creates a Webserver webhook manually to listen to Events sent by Slack.
It is using latest library **slackclient>=2**.
There are few changes, such as username become user full name - Slack API deprecates `name` field and intends to use `display_name` instead, paging while loading all channels, caching some functions to avoid querying information at all times...
Not complete, but enough to interact with messages! :)

### DIFF
--- a/errbot/backends/slack_events.plug
+++ b/errbot/backends/slack_events.plug
@@ -1,0 +1,6 @@
+[Core]
+Name = SlackEvents
+Module = slack_events
+
+[Documentation]
+Description = This is the Slack Events backend for Errbot.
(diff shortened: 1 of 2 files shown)


nzlosh:
I think this duplicates the work here https://github.com/errbotio/errbot/pull/1451

duhow:
Oh, didn't know there was another PR 😮 
This backend was waiting for too long and I needed to write it recently for replacing my old bot.

PR #1475, opened 2020-11-01
docs: Fix typographical issues in setup guide

e4r7hbug:
Some minor text issues in the setup guide:

* Packaged examples looks like it wants to be a list for readability
* Use the `:code:` role to avoid double-dashes turning into en-dashes by Docutils Smart Quotes


sijis:
Thank you for the fixes.

PR #1477, opened 2020-11-19
Enable testing using Python 3.9

nzlosh:
Python 3.9 is stable and people are running errbot with it.  This PR is to enable testing for python 3.9 to have better coverage to catch any potential issues.


nzlosh:
I updated the `.travis.yaml` file, but it looks like all tests are run under github actions now, so I remove the file?

sijis:
Yes. Could you create a new pr that only removes travis, along with any other references to it, like the banner, etc?

sijis:
:+1:

PR #1478, opened 2020-11-19
Remove travis configuration

nzlosh:
Remove travis configuration from project and use github actions badge.


nzlosh:
This PR should be merged after #1477 to avoid conflicts.

sijis:
There is one reference to travis under https://errbot.readthedocs.io/en/latest/user_guide/plugin_development/testing.html#travis-and-coveralls.

I'm thinking we could just reference using tox and choosing travis or github actions.
In fact, i don't think we use coveralls at all either :thinking:.

nzlosh:
I saw that and thought it was more a suggestion for plugin developers test their plugins.  It maybe be useful to re-write that part of the documentation to suggesting using github actions, but I felt it was out of scope for travis removal.

sijis (approved):

PR #1487, opened 2020-12-08
Let send() and send_card() return the message that was created and sent

codablock:
This allows the backend implementation to attach meta information to the
message so that it can be used as parent message in threaded responses.


codablock:
Some additional information: Right now, you can only respond to messages that the bot received or to own messages when `send_message()` is used with a self crafted Message object. It's however not possible to do the same with message sent via `send()`or `send_card()`. This PR fixes this by returning the internally created Message objects.

For `send_card()`, this will only work when the underlying backend is sending something that is representable via a `Message` object. This is automatically true for all backends that use the Markdown templates for cards. Backends like Slack however would need to implement their own handling and creation of Message objects.

codablock:
Regarding build failures...I have absolutely no idea what's wrong.

sijis:
> Regarding build failures...I have absolutely no idea what's wrong.

Looks like a Github issue. I reran all the tests.

codablock:
@sijis Do you still plan to merge this? I'm cleaning up my open PRs atm and would like to close this one if there is no interest in it.

sijis:
I think so. I may have to tweak the PR a bit and what not.

I'd leave it open for now. I'll close it if it's not going to work out.

PR #1497, opened 2021-01-11
Remove Google+ References

fernand0:
Google + does not exist anymore. The same information is published in the PyPi page of the project.


sijis:
Thank you. I thought we've removed all references to it. i guess we missed one!
pypi should be updated upon the next release.

PR #1500, opened 2021-01-19
Add identifier for callbacks room join and room leave

duhow:
Enable to use **callback_room_joined** and **callback_room_left** with other identities (Persons), rather than the bot itself only.
Not sure if this is a breaking change, currently testing with Slack Events Backend (custom).

### DIFF
--- a/errbot/backends/slack.py
+++ b/errbot/backends/slack.py
@@ -601,7 +601,8 @@ class SlackBackend(ErrBot):
         """Event handler for the 'member_joined_channel' event"""
         user = SlackPerson(self.sc, event["user"])
         if user == self.bot_identifier:
-            self.callback_room_joined(SlackRoom(channelid=event["channel"], bot=self))
+            user = self.bot_identifier
+        self.callback_room_joined(SlackRoom(channelid=event["channel"], bot=self), user)
 
     def _reaction_event_handler(self, event):
         """Event handler for the 'reaction_added'
--- a/errbot/backends/slack_rtm.py
+++ b/errbot/backends/slack_rtm.py
@@ -572,9 +572,11 @@ class SlackRTMBackend(ErrBot):
         """Event handler for the 'member_joined_channel' event"""
         user = SlackPerson(webclient, event["user"])
         if user == self.bot_identifier:
-            self.callback_room_joined(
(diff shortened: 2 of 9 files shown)


duhow:
I'm thinking if I should just change Room to RoomOccupant, instead of adding new params.
@sijis let me know your concerns.

errbot/backends/slack.py
```diff
@@ -601,7 +601,8 @@ def _member_joined_channel_event_handler(self, event):
         """Event handler for the 'member_joined_channel' event"""
         user = SlackPerson(self.sc, event["user"])
         if user == self.bot_identifier:
-            self.callback_room_joined(SlackRoom(channelid=event["channel"], bot=self))
+            user = self.bot_identifier
```
sijis, line 604: I don't think this if condition is needed, if the callback is moved from it.
Setting user to self.bot_identifier does nothing from what i can tell.

I don't know offhand what the value of user is, if it does not match `self.bot_identifier`.

errbot/backends/slack_rtm.py
```diff
@@ -572,9 +572,11 @@ def _member_joined_channel_event_handler(self, webclient: WebClient, event):
...
         user = SlackPerson(webclient, event["user"])
         if user == self.bot_identifier:
-            self.callback_room_joined(
-                SlackRoom(webclient=webclient, channelid=event["channel"], bot=self)
-            )
+            user = self.bot_identifier
+        self.callback_room_joined(
+            SlackRoom(webclient=webclient, channelid=event["channel"], bot=self),
+            user
+        )
```
sijis, line 579: Same as earlier comment.

sijis (commented):
I like what is going on here.

I'll have to run through this and see if there's anything at a quick glance that i've missed.

You may need to rebase too.

errbot/backends/slack.py
duhow: Slack Backend may generate additional calls when loading Person info, so instead of returning a new SlackPerson, I'm giving the existing one.
Actually I may just compare against `bot_identifier.userid` to make it quicker.
But the main reason for this is to lower the amount of API calls done to Slack.

sijis:
@duhow Thank you for your work. I rebased and made some minor commits to get this mergeable.

@nzlosh - fyi callback could be useful for the v3 backend.

PR #1505, opened 2021-01-30
fix: circular dependencies error when there are none

### LINKED ISSUE #1397: Circular Dependencies when there are none
### I am...

* [ ] Reporting a bug
* [ ] Requesting help with running my bot
* [x] Here about something else

### I am running...

* Errbot version: 6.1.1
* OS version: Tried on Windows and MacOS
* Python version: 3.6.8
* Using a virtual environment: N/A

### Issue description

I have a situation where I have a total of 3 plugins:

- PluginA
- PluginB
- PluginC

Below are their plugin dependencies:

- PluginA depends on PluginB and PluginC
- PluginB depends on PluginC

I've roughly visualized the dependencies here:

deps

Trying to run with this configuration results in an error:

```
2019-12-05 20:21:59,078 INFO     errbot.core               Error: PluginA failed to activate: Circular dependency in the set of plugins (PluginC, PluginA, PluginB).
```

PluginA fails to load, due to "circular dependencies". However, I do not see circular dependencies from what I described above.

So is this intentional; not allowing PluginA to depend on PluginC, since PluginB already depends on PluginC, and PluginA does depend on PluginB?

There aren't circular dependencies, since `self` depends on the same plugins as a child plugin depends on.

A workaround is to do something like this:

PluginA does not depend on PluginC

Then, from within `PluginA` do teh following:
```python
self.get_plugin('PluginB').get_plugin('PluginC')
```

### Steps to reproduce

I've created a repo with the necessary code to reproduce this error [here](https://github.com/jakubclark/errbot-dependencies).
Simply run the [test](https://github.com/jakubclark/errbot-dependencies/blob/master/plugins_test.py): `pytest plugins_test.py`.
Or run `errbot` itself, it is configured for text mode.

browniebroke:
I have the same issue as the one reported in #1397. I'm not sure I can fix it, but I thought it would be helpful to at least write a test reproducing the issue. 

Fix #1397 

## Implementation note

I initially had:

- `chained1` depends on `chained2`, `chained3`
- `chained2` depends on `chained3`
- `chained3` depends on nothing

But this worked fine, there was no problem. It looks like the problem depends on the plugin names, I can reproduce with:

- `chained3` depends on `chained2` and `chained1`
- `chained2` depends on `chained1`
- `chained1` depends on nothing

## Checklist

- [x] Add a new test case reproducing the problem reported in #1397
- [x] Fix the problem

### DIFF
--- a/tests/commands_test.py
+++ b/tests/commands_test.py
@@ -169,6 +169,10 @@ def test_broken_plugin(testbot):
         )
         assert "import borken  # fails" in testbot.pop_message()
         assert "as it did not load correctly." in testbot.pop_message()
+        assert (
+            "Error: Broken failed to activate: "
+            "'NoneType' object has no attribute 'is_activated'"
+        ) in testbot.pop_message()
         assert "Plugins reloaded." in testbot.pop_message()
     finally:
         rmtree(tempd)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -34,6 +34,7 @@ fixes:
 - fix: removed deprecated argument reconnection_interval for irc v20.0 (#1568)
 - docs: Add Gentoo packages (#1567)
 - chore: bump actions/setup-python from 3.1.0 to 3.1.2 (#1564)
+- fix: circular dependencies error when there are none (#1505)
 
 v6.1.8 (2021-06-21)
 -------------------
--- a/errbot/plugin_manager.py
+++ b/errbot/plugin_manager.py
@@ -421,7 +421,7 @@ class BotPluginManager(StoreMixin):
         """
         Activates all plugins that are not activated, respecting its dependencies.
 
-        :return: Empty string if no problem occured or a string explaining what went wrong.
+        :return: Empty string if no problem occurred or a string explaining what went wrong.
         """
         log.info("Activate bot plugins...")
         errors = ""
--- a/errbot/plugin_manager.py
+++ b/errbot/plugin_manager.py
@@ -469,7 +469,7 @@ class BotPluginManager(StoreMixin):
             name: set(info.dependencies) for name, info in self.plugin_infos.items()
         }
         plugins_in_cycle = set()
-        while 1:
+        while True:
             plugins_sorter = TopologicalSorter(plugins_graph)
             try:
                 # Return plugins which are part of a circular dependency at the end,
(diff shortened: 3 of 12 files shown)


sijis:
I really appreciate you putting this together.

Could you elaborate what you mean by 

> It looks like the problem depends on the plugin names

Maybe its late but I'm missing but both scenarios looks similar if not identical to me.

browniebroke:
I agree, the 2 cases should be equivalent, but one passes the test I wrote while the other doesn't. Since the only difference is with the names I assumed that it's playing a role into the problem. 

I guess the plugin discovery system order plugins somehow related to their names. If the name ordering matches the dependency ordering, they can start, otherwise they can't.

tests/commands_test.py
```diff
@@ -169,6 +169,10 @@ def test_broken_plugin(testbot):
         )
         assert "import borken  # fails" in testbot.pop_message()
         assert "as it did not load correctly." in testbot.pop_message()
+        assert (
+            "Error: Broken failed to activate: "
+            "'NoneType' object has no attribute 'is_activated'"
+        ) in testbot.pop_message()
```
browniebroke, line 175: This new error happens because this line returns `None`:

https://github.com/browniebroke/errbot/blob/c81c78ec29592cb17d29f3aaa4b954f23f9a77ce/errbot/plugin_manager.py#L431

Which cause an attribute error down here:

https://github.com/browniebroke/errbot/blob/c81c78ec29592cb17d29f3aaa4b954f23f9a77ce/errbot/plugin_manager.py#L439

Open to suggestions on how to fix it better.

tmcevoy14:
Hi @browniebroke any idea when this fix will be merged?

browniebroke:
No idea, no, it's not really up to me at this point. I _think_ the next step would be for a maintainer to assess whether they are happy with my code of if they have any suggestions. This is probably not the only thing on their plate, though...

In the meantime, if you want to help, I'm sure they would appreciate if you could try my version out and see if it maybe has a bug in it. You can install from source with pip:

`pip install -e git+https://github.com/browniebroke/errbot.git@36920efa7818d71d4d9dc839100385f384d4c665#egg=errbot`

You can put the part after `pip install` in your requirements file too. If you find any problem or if it works like a charm, please report back here.

sijis:
@browniebroke @tmcevoy14 Would you folks be able to see if this issue persist with the latest changes on `master` branch?

Issue #1510, opened 2021-03-07
`errbot --init` is failure failed when passed relative path as bot directory 


attakei:
In order to let us help you better, please fill out the following fields as best you can:

### I am...

* [x] Reporting a bug
* [ ] Suggesting a new feature
* [ ] Requesting help with running my bot
* [ ] Requesting help writing plugins
* [ ] Here about something else

### I am running...

* Errbot version: 6.1.7
* OS version: Arch
* Python version: 3.9
* Using a virtual environment: no

### Issue description

If I pass sub directory when run `errbot --init`, command does not complete correctly and example plugin is not created.

### Steps to reproduce

1. Create sub directory in current working directory(anywhere).
1. Run `errbot --init` passing sub directory as relative path.

Example:

```shell-session
$ mkdir errbot-env
$ errbot --init errbot-env 
The initialization of your errbot directory failed: [Errno 2] No such file or directory: 'errbot-env/errbot-env/plugins/err-example'.
```

Patterns of reproduce: 

- Reproduce:
  - Pass relative path of children directory (any hierarchy)
  - Pass relative path of 2-step upper hierarchy (example: `../../errbot-env`)
- No reproduce:
  - Pass relative path of parent directory (example: `../errbot-env`)
  - Pass absolute path

### Additional info

I will create PR to fix it.

sijis:
PR merged.

sijis:
Thank you for your contribution.

PR #1511, opened 2021-03-07
Make example_plugin_dir using not directly base_dir

attakei:
This is for fix to #1510.

Base thinking for fix (python repl):

```
>>> from pathlib import Path
>>> base_dir = Path("foo")
>>> extra_plugin_dir = base_dir / "plugins"

>>> example_plugin_dir = base_dir / extra_plugin_dir / "err-example"
>>> example_plugin_dir
PosixPath('foo/foo/plugins/err-example')

>>> example_plugin_dir = extra_plugin_dir / "err-example"
>>> example_plugin_dir
PosixPath('foo/plugins/err-example')
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,7 @@ features:
 
 fixes:
 
+- core/cli: failure when passing relative directory during --init (#1511)
 - backend/xmpp: include message delayed for send/recieved messages (#1270)
 - backend/xmpp: "unexpected keyword argument 'wait'" when connecting (#1507)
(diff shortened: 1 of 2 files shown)


attakei:
I think `base_dir` should be created absolute path even if it passed any values of arguments.
(not changed in commit)

example:

```diff
-base_dir = pathlib.Path.cwd() if args["init"] == "." else Path(args["init"])
+base_dir = pathlib.Path.cwd() if args["init"] == "." else Path(args["init"]).resolve()
```

PR #1513, opened 2021-05-05
Several logging fixes

zaughon:
Several small fixes for logging-related bugs.

* Fix: Setting a custom BOT_LOG_FORMATTER no longer stops logging to a file 
* Fix: Check that the BOT_LOG_FILE attribute exists in config before trying to use it (don't break bot-startup if not set)


sijis:
@zaughon Thank you!

Issue #1514, opened 2021-05-10
KeyError exception in slack _reaction_event_handler


maninthewall:
### I am...

* [x] Reporting a bug
* [ ] Suggesting a new feature
* [ ] Requesting help with running my bot
* [ ] Requesting help writing plugins
* [ ] Here about something else

### I am running...

* Errbot version: 6.1.X
* OS version: Ubuntu 2020
* Python version: 3.8
* Using a virtual environment: yes

### Issue description

In the legacy slack backend, under `_reaction_event_handler`, there is a check for `event["item_user"]`.  The `item_user` attribute of the event won't always be available for some events. Slack says the following:

> Some messages aren't authored by "users," like those created by incoming webhooks. `reaction_added` events related to these messages will not include an `item_user`.

In its current state, Errbot will throw a KeyError exception on the check `if event["item_user"]` on line 613 of slack.py

### Steps to reproduce
I believe the steps to reproduce would be to send a message to Slack via a webhook, then add a reaction to it.  

### Additional info

I will submit a PR fixing this issue.

PR #1515, opened 2021-05-10
Fixes issue KeyError exception in slack _reaction_event_handler #1514

maninthewall:
Fixes the issue described in issue #1514
Adds unit tests for the reaction handler


sijis:
@nzlosh fyi.

PR #1520, opened 2021-07-10
Fix success handling for update_repos

bboe:
As part of the switch to python git, the success case handling was reversed even though `0` was still meant to indicate success (formerly used to represent process exited without error).

https://github.com/errbotio/errbot/commit/89bce02a836fdbf14f184ad54e3f9eb1a7becbb1#diff-b279f20d61152a2a99689009a59d70609c92039568c203da8c84bb56ceacd94fR287

This pull request fixes that issue and makes the code a bit more clear by using booleans rather than integers to represent the success state.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,10 @@
+v6.1.9 (unreleased)
+-------------------
+
+fixes:
+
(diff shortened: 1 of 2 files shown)


sijis:
Thank you!

PR #1521, opened 2021-07-10
Reload all repo plugins when updating a repo

bboe:
`get_plugin_by_path` would only return the first plugin encountered in a
repository. This change adds the method `get_plugins_by_path` which yields
plugins as they are found.

For backwards compatibility, I have not deleted the original
`get_plugin_by_path` method, lest anyone use that method directly outside of
this package.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,7 @@ fixes:
 
 - core: success handling for update_repos (#1520)
 - core/plugins: cascade dependency plugins (#1519)
+- core/plugins: reload all repo plugins when updating a repo (#1521)
 
 v6.1.8 (2021-06-21)
 -------------------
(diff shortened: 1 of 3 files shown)


sijis:
Is this reloading all the plugins found in a specific repo?

bboe:
Yes. Errbot already handles reading just not reloading.

bboe:
Just the ones in a specific repository

sijis:
So in this scenario, a repo has multiple plugins in it, correct?

bboe:
> So in this scenario, a repo has multiple plugins in it, correct?

Yes, that's correct. In my case, multiple `.plug` files in the root of the repository.

sijis (approved):

sijis:
Thank you for these fixes.

PR #1523, opened 2021-07-12
Add repos_branch command

bboe:
The method for doing the checkout is likely not the most efficient as it deletes everything in the working tree, and relies upon `reset_index` to restore all the files. Nevertheless, it appears to get the job done.

Usage:

    !repos branch REPO BRANCHNAME

Note: Just like `repos update`, which also relies upon the same `pull` command, it does not support pulling in history-rewriting updates.

### DIFF
--- a/errbot/core_plugins/plugins.py
+++ b/errbot/core_plugins/plugins.py
@@ -4,7 +4,7 @@ import shutil
 from ast import literal_eval
 from pprint import pformat
 
-from errbot import BotPlugin, botcmd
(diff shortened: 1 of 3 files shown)


bboe:
After testing some more, this isn't quite ready as `repos update` doesn't work after changing to a non-primary branch.

I can work around that by reloading the relevant plugins on branch switch, but switching branches shouldn't break any of the other commands.

I'd also like to handle the force-push case, which I think I'll solve by simply deleting the local branch before switching.

PR #1525, opened 2021-07-16
chore: remove HipChat backend

sijis:
Removing HipChat backend as the service [no longer exists](https://www.atlassian.com/partnerships/slack).


nzlosh:
:+1: Good initiative.  I don't think keeping the backend around is a good look for errbot when the chat service itself has been shutdown.

PR #1530, opened 2021-08-27
Don't use str(room) for acl matches

webpigeon:
While fixing the documentation, I noticed that rooms are using str(room)
for acl matches. `__str__` is meant to be a human-readable object
representation which could be something other than the room ID -
although as no one spotted this I guess that in most cases it is.

This commit makes it so that rooms have an aclattr, just like people do.
This can then be used to store a suitable version of the room for ACL
matches.

I've tried to maintain backwards compatibility in two ways:

  1. aclattr defaults to the old behaviour str(self)
  2. the getter can detect if the room (Occupant) is a room or a str,
     and if it's a str use that instead. I know for the Matrix backend
     I'd incorrectly been returning a string and it still worked - if
     other backends are doing that it'll throw an exception without this
     mitigation.

I've checked using tox, and it passed the tests :).

```
class MatrixRoom(MatrixIdentifier, backend.Room):
    """Representation of a matrix room.

    Provides a way to do room-related things. We need to be careful because this is one of the main places
    that the async and sync code mixes.
    """

    def __init__(self, mxid: str):
        super().__init__(mxid)
        self.mxid = mxid

   @property
    def aclattr(self) -> str:
        return self.mxid
  
    def __str__(self):
        return "{} ({})".format(self.display_name, self.machine_name)
```

in terms of bot usage differences, there shouldn't be any vs the original room-string based approach.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,7 @@ features:
 - core: add type hints to core and backend functions (#1542)
 - docs: add ACL and numerous backends to official documentation (#1552)
 - core: add Python 3.10 to automated tests (#1539)
+- core: add room acl attribute (#1530)
 
 fixes:
 
--- a/errbot/backends/base.py
+++ b/errbot/backends/base.py
@@ -140,6 +140,13 @@ class Room(Identifier):
         )
 
     @property
+    def aclattr(self) -> str:
+        """
+        :return: returns the unique identifier that will be used for ACL matches.
+        """
+        return str(self)
+
+    @property
     def exists(self) -> bool:
         """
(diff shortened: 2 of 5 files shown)


sijis:
I would still appreciate an example for this.

webpigeon:
Sorry - got sidetracked, I've added two tests and implementation of the acl attribute for testing purposes. The existing tests should be evidence that the old behaviour is kept if there is no change.

duhow (approved):

sijis (approved):

sijis:
@webpigeon Thank you for your contribution!

PR #1534, opened 2021-10-02
Add configurable aliases for commands and sub-commands

### LINKED ISSUE #1279: Add support for command aliases
It'd be nice to have support for command aliases that would allow use of shortcuts for long commands (e.g. `!ls`, `!list`, or `!metrics`, `'!performance_metrics`) or use synonyms for commands with the same meaning (e.g. `!stop`, `!cancel`)

LemmaEOF:
Closes #1279.

This pull request adds a new field to the config for defining aliases to commands and sub-commands. These are helpful for users to add shortcuts for often-used actions. Aliases are checked after standard commands, to avoid overriding commands added from plugins. Correction suggestions incorporate aliases as well.

The `COMMAND_ALIASES` field in the config takes a dictionary of aliases to standard commands. `{"h": "help", "pinfo": "plugin info"}` would make it so running `!h` would return the result for `!help`, and running `!pinfo Help` would return the result for `!plugin info Help`.

Tests for Python 3.9 are currently failing on my machine, but this is likely an issue with my Python configuration as the Python 3.8 tests pass without issue. I'm going to verify the tests tomorrow on another computer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.pyc
 __pycache__
 .venv
+venv
 /.coafile
 /.idea
 /atlassian-ide-plugin.xml
--- a/errbot/bootstrap.py
+++ b/errbot/bootstrap.py
@@ -72,6 +72,8 @@ def bot_config_defaults(config: object) -> None:
         config.TEXT_COLOR_THEME = "light"
     if not hasattr(config, "BOT_ADMINS_NOTIFICATIONS"):
         config.BOT_ADMINS_NOTIFICATIONS = config.BOT_ADMINS
+    if not hasattr(config, "COMMAND_ALIASES"):
+        config.COMMAND_ALIASES = {}
 
 
 def setup_bot(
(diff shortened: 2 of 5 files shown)


sijis:
@LemmaEOF Thank you for this. Would you be able to include an example and include some tests around this?

LemmaEOF:
Sounds good! I'll add them in as soon as I can.

LemmaEOF:
Alright, test added! I put an example in the PR description, since I think that's what you meant. If it's not, please let me know and I can put it elsewhere instead! All the tests pass on Python 3.9 on my other machine (except for the ones that involve downloading the Hello World plugin due to annoying windows file lock stuff)

sijis:
I like the idea of including this into core. I wondering if it makes better sense to keep this functionality as a plugin (https://github.com/rroemhild/err-alias), which does the same thing and can be dynamically edited without reloading the bot.

LemmaEOF:
Oh right yeah, bot config is only loaded at launch. I'll see if I've got time to look into other ways to load aliases soon; I'm out of action for a little while due to some IRL stuff.

PR #1546, opened 2021-11-15
Update README to replace deprecated add-on

ciaranjordan:
Cisco Spark has been rebranded to Webex and @marksull has deprecated the Cisco Spark add-on. 

This change would update the README to remove the reference to the deprecated add-on and replace it with a link to the replacement.


sijis (approved):
It's just missing an entry in CHANGES file

PR #1553, opened 2022-01-23
chore: pin all package dependencies

sijis:
Pinning all package dependencies.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,6 +25,7 @@ fixes:
 - backends: deprecate built-in Slack and SlackRTM (#1526)
 - chore: remove python 3.6 checks and test environment (#1540)
 - chore: add/update issue templates (#1554)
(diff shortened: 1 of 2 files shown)


nzlosh (approved):
This looks good to me.  Is there a security bot in-place that'll track dependencies (dependabot or whatever) that will prompt us for stream updates?

sijis:
> This looks good to me. Is there a security bot in-place that'll track dependencies (dependabot or whatever) that will prompt us for stream updates?

I had the same thought too. There is [dependabot](https://docs.github.com/en/code-security/supply-chain-security/managing-vulnerabilities-in-your-projects-dependencies/configuring-dependabot-security-updates) but based on our current project layout I dont think it will detect the packages, so i'd have to swtich things up a bit to make that work. I did enable it though, so we'll see.

PR #1556, opened 2022-02-09
Use errbot log level with webserver for consistent logging

### LINKED ISSUE #1555: Incorrect log levels showing for werkzeug
**Describe the bug**
When I have configured the Log Level to "WARNING", werkzeug always outputs the Webserver logs as in "INFO" to the console.
`BOT_LOG_LEVEL = logging.WARNING`

**To Reproduce**
```
@webhook('/health_check/')
def health_check(self, request):
	self.log.debug(repr(request))
	pass
```

Enable the Webserver in Errbot and build a webhook using the above code.

**Expected behavior**
Zero console logging under the INFO level when config level is set to WARNING

**Screenshots**
image

**Environment (please complete the following information):**
- Errbot version: 6.1.8
- OS version: Ubuntu 20.04
- Python version: Python 3.8.10
- Using a virtual environment: yes
- Using Docker: yes

nzlosh:
This patch will take errbot's log level configuration and pass it to werkzeug for consistent logging.

fixes #1555

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,6 +26,7 @@ fixes:
 - chore: remove python 3.6 checks and test environment (#1540)
 - chore: add/update issue templates (#1554)
 - chore: pin all package dependencies (#1553)
+- core/webserver: use errbot loglevel for consistent logging. (#1555)
 
 v6.1.8 (2021-06-21)
 -------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,7 +26,7 @@ fixes:
 - chore: remove python 3.6 checks and test environment (#1540)
 - chore: add/update issue templates (#1554)
 - chore: pin all package dependencies (#1553)
-- core/webserver: use errbot loglevel for consistent logging. (#1555)
+- core/webserver: use errbot loglevel for consistent logging. (#1556)
 
 v6.1.8 (2021-06-21)
 -------------------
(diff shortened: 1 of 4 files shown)


sijis:
@nzlosh Do you think this is still necessary after the workaround given in #1555 ?

nzlosh:
Either we handle logging consistently in the core or we clearly document this workaround in the official documentation.  I'm inclined to go with this patch since it's more obvious how logging is controlled.

I did wonder if people might want access logs (INFO level for werkzeug) but less verbose logging for errbot. If someone does create a request for such a feature, I think it'd make sense to split out werkzeug log level and log entries into a separate log file and create news variables `WEBSERVER_LOG_FILE` and `WEBSERVER_LOG_LEVEL`.

This would also encapsulate the webserver implementation details and avoid breaking configuration if we switch to a different webserver in the future.

AshleyJackson:
In my implementation I use other modules that have their own logging (outside of werkzeug) that still send to console, even if the configuration is set correctly.

If there is an option for the Root configuration to take precedence over sub-modules, I believe that that would be beneficial for nearly everyone who relies on the log outputs for Errbot.

If logging is required, developers can create their own log output where necessary using self.log.*(msg) IMO.

nzlosh:
Errbot should not be attempting to override logging in other peoples code as there is no reasonable to way to know if it is required for the use case.  User code will need to handle logging based on their needs.  They have access to `self.bot_config.BOT_LOG_LEVEL`  if they want to set the log level for their loggers to be the same as errbot's configuration.

sijis (approved):

PR #1557, opened 2022-02-16
Prevent infinite loop when only BOT_PREFIX passed

davidamin:
I have hit an error in which a message that contains only my bot prefix causes the bot to become unresponsive to any further messages. After a bit of debugging, I believe the problem is that `text` here becomes an empty string, which makes the value of `i` start as 0, then decrement (because the empty string is not in `commands`), leading to an infinite loop where `i` decreases forever.

I have confirmed that changing the comparison operation to `<=` fixes this case, without breaking other commands as far as I can tell.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,6 +27,7 @@ fixes:
 - chore: add/update issue templates (#1554)
 - chore: pin all package dependencies (#1553)
 - core/webserver: use errbot loglevel for consistent logging. (#1556)
+- fix/core: prevent infinite loop when only BOT_PREFIX is passed (#1557)
 
 v6.1.8 (2021-06-21)
 -------------------
--- a/errbot/core.py
+++ b/errbot/core.py
@@ -350,7 +350,7 @@ class ErrBot(Backend, StoreMixin):
                         args = " ".join(text_split[i:])
                     else:
                         i -= 1
-                if i == 0:
+                if i <= 0:
                     break
 
             if (


nzlosh:
As I understand the code, the proposed change will stop processing the first element of the command.  I'm interested in reproducing the infinite loop issue.

Can you provide the following information so I can setup errbot as close to your setup please:
 - Execution environment
 - Errbot version
 - Chat backend
 - Bot Configuration
 - Command that triggers the loop

davidamin:
Sure thing! I've been able to reproduce this on version `6.1.8` using the text backend, with the default configuration defined by `errbot --init`.

To reproduce, I fire up the bot with `errbot`, then run a simple test command like `!help`. That returns as expected, but sending just `!` as the command stops the bot from ever returning. I have been able to reproduce the same behavior with the Slack backend, and by running the bot in Docker.

davidamin:
Perhaps a better way to resolve the issue here would be by checking if the `text` string is empty? Happy to update the PR if that would be a cleaner approach

nzlosh:
Sorry for the delay, I was able to reproduce the infinite loop as you described.   This a really great catch, thank you for reporting this.

Your initial PR is a good solution.  As a small bonus, if the test is moved from the bottom of the loop to the top and test for any value lower than 1, it will avoid the infinite loop as well as not spend time and resource to acquire the thread lock when the test_split list is empty. 

```
            while cmd is None:
                if i < 1:
                    break
```

sijis (approved):
Nice catch. Thanks!

PR #1562, opened 2022-03-22
Import ABC from collections.abc for Python 3.10 compatibility

tirkarthi:


sijis:
@tirkarthi thanks for the PR. Were you seeing an error / warning on python 3.10? Also, do you if there are any backwards version issues merging this?

tirkarthi:
I don't see an error in running the tests. I guess the code path is not tested as they seem to be not hit or probably mocked. In terms of compatibility the deprecation warning is from Python 3.4 and above so the patch is compatible with Python 3.4+ and I don't see any issues.

jonesnc:
@tirkarthi Are you testing with the slack plugin enabled? That's where I'm seeing this collections import error:

```
   File "/usr/lib/python3.10/site-packages/errbot/backends/base.py", line 868, in serve_forever
     if self.serve_once():
   File "/usr/lib/python3.10/site-packages/errbot/backends/slack.py", line 427, in serve_once
     self.auth = self.api_call("auth.test", raise_errors=False)
   File "/usr/lib/python3.10/site-packages/errbot/backends/slack.py", line 378, in api_call
     if not isinstance(response, collections.Mapping):
 AttributeError: module 'collections' has no attribute 'Mapping'
```

sijis:
Fix in more recent PR.

PR #1585, opened 2022-06-23
Update exception if installing on Python <3.7

browniebroke:
Python 3.6 was dropped in https://github.com/errbotio/errbot/pull/1540 but the exception was missed, leaving a misleading error message.


sijis:
@browniebroke Thank you!

PR #1587, opened 2022-06-29
Create codeql-analysis.yml

cyber-pi:


sijis:
@cyber-pi could you elaborate what this is for?

sijis:
I see. I was curious as I see multiple PRs for the same thing and it looks like spam.

We really appreciate the the effort put here. Im just trying to understand what does codeql do/offer that would be useful to the project.

PR #1590, opened 2022-07-03
Feature: plugins detection using entrypoints

sijis:
This feature should allow plugins to be detected using entrypoints when installed via pip.

The entrypoints are
* `errbot.plugins` - normal botcmd plugins
* `errbot.backend_plugins` - plugins related to communications systems (Slack, Gitter, etc)

A package would need to include this snippet in their `setup.py` to work.
Using, https://github.com/errbotio/err-helloworld as an example, we need to add

```
...
    entry_points = {
        "errbot.plugins": [
            "helloworld = helloWorld:HelloWorld",
        ]
    }
...
```

I would suggest, plugins use the naming convention of `errbot-plugin-$pluginName`

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,10 @@
 v9.9.9 (unreleased)
 -------------------
 
+features:
+
+- core/plugins: detect plugins using entrypoints (#1590)
+
 fixes:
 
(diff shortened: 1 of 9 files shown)


nzlosh (approved):
LGTM

PR #1597, opened 2022-09-02
feat: add new config `SENTRY_OPTIONS` for more flexible Sentry configuration

browniebroke:
The Sentry SDK has several options, but right now we are limited by errbot as to which one can be configured.

This adds a new `SENTRY_OPTIONS` config, which enable users to specify any option supported by the SDK in the `sentry_sdk.init()` call.

Our use case is that we have 2 instances of a given bot running, one for testing changes to our plugins and one which we use (like staging & prod) and we use the [`environment` option](https://docs.sentry.io/platforms/python/configuration/options/#environment) to tell which bot caused the error. I assume some people might find the `release` option useful too.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,7 @@ breaking:
 features:
 
 - core/plugins: detect plugins using entrypoints (#1590)
+- core/logging: add new SENTRY_OPTIONS config (#1597)
 
 fixes:
 
--- a/docs/user_guide/sentry.rst
+++ b/docs/user_guide/sentry.rst
@@ -50,8 +50,9 @@ To setup Errbot with Sentry:
 * Set **BOT_LOG_SENTRY** to *True* and fill in **SENTRY_DSN** with the DSN value obtained previously
 * Optionally adjust **SENTRY_LOGLEVEL** to the desired level
 * Optionally adjust **SENTRY_TRANSPORT** to the desired transport
+* Optionally adjust **SENTRY_OPTIONS** to customise the rest of the initialization.
 * Restart Errbot
 
-You can find a list of `Sentry transport classes <https://docs.sentry.io/clients/python/transports/>`_.
+You can find a list of `Sentry options <https://docs.sentry.io/platforms/python/configuration/options/>`_.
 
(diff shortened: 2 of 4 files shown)


sijis:
Is there a way to consolidate all the sentry options into this new option, basically removing SENTRY_TRANSPORT?

browniebroke:
Sure, can do that. That would be a breaking change, is it something that we should deprecate first?

sijis:
@browniebroke that's a fair concern. There is some planned work that may cause other breaks too.

Would you be ok with creating a separate PR with the separate breaking changes?

browniebroke:
Yes sure :+1:

sijis:
@browniebroke With the single `SENTRY_OPTIONS` option now, how could someone specify a different transport?

browniebroke:
> @browniebroke With the single `SENTRY_OPTIONS` option now, how could someone specify a different transport?

They would specify it as per [the Sentry docs](https://docs.sentry.io/platforms/python/configuration/options/#transport), but to be honest, that's not super clear over there. I looked into their tests and I found an example here: 

https://github.com/getsentry/sentry-python/blob/6db44a95825245b1f7c9baa54957d044f7be18eb/tests/tracing/test_integration_tests.py#L245-L254

browniebroke:
I think the transport has been simplified compared to the old Sentry client (Raven) which was providing multiple transport classes: https://docs.sentry.io/clients/python/transports/

The newer client sentry-sdk is lighter on the topic: https://docs.sentry.io/platforms/python/guides/airflow/configuration/options/#transport-options

browniebroke:
Let me know if I missed anything. I'm not sure it's worth expanding more on this in the docs?

sijis:
@browniebroke Does the PR https://github.com/errbotio/errbot/pull/1605 superseed this one or does this one need to be merged first before #1605?

browniebroke:
It builds on top of it, so it's a bit of both. 

If it was my project, I would probably merge this first and then merge the other, this way the 2 PR would appear as 2 distinct changes, but it's up to you. Depends how you like to separate things.

sijis (approved):

PR #1602, opened 2022-09-26
Update slackv3 configuration/installation and development documentation

nzlosh:
To ease the burden of maintaining documentation for the backends, the slackv3 README will no longer have explicit documentation for configuration/installation but provide links to the official errbot documentation.

### DIFF
--- a/docs/user_guide/configuration/slackv3.rst
+++ b/docs/user_guide/configuration/slackv3.rst
@@ -19,7 +19,7 @@ installed in a python virtual environment (adjust the command to your errbot's i
 
     git clone https://github.com/errbotio/err-backend-slackv3.git
     source /opt/errbot/bin/activate
-    /opt/errbot/bin/pip install -r /opt/errbot/backends/err-backend-slackv3/requirements.txt
+    /opt/errbot/bin/pip install .
 
 Connection Methods
 ------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


docs/user_guide/configuration/slackv3.rst
```diff
@@ -22,15 +22,18 @@ installed in a python virtual environment (adjust the command to your errbot's i
     /opt/errbot/bin/pip install -r /opt/errbot/backends/err-backend-slackv3/requirements.txt
```
tejaswivinodg, line 22: 👋 @nzlosh since you are working on this documentation PR, I found that `requirements.txt` has been removed from the repo in this PR 
https://github.com/errbotio/err-backend-slackv3/pull/82/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552.

nzlosh:
Hi @tejaswivinodg thank you for taking the time to pointing this out.

The removal of `requirements.txt` is intentional because the dependencies have been moved to `setup.py`.  This is part of making the slackv3 backend available on pypi.  We had expected to publish the package immediately afterwards, but encountered an unexpected problem with module imports when installing via pypi.

You can install the backend from a git clone using the following command in the root of the clone directory:
`pip install .`

Sorry for the inconvenience caused by this.

nzlosh:
OK, I realised I'd misunderstood what you were asking.  I'll update the documentation, thank you for pointing that out! :+1:

PR #1605, opened 2022-10-04
Deprecate the `SENTRY_TRANSPORT` option

browniebroke:
This PR builds on top of #1597. Due to the addition of the `SENTRY_OPTIONS` setting, we no longer need a custom one for the transport option.

In the newer client `sentry-sdk`, [the transport option](https://docs.sentry.io/platforms/python/guides/airflow/configuration/options/#transport-options) has been simplified compared to the old Sentry client (Raven) which was [providing multiple transport classes](https://docs.sentry.io/clients/python/transports/). 

The SDK only ships with the base abstract `Transport` class and the default `HttpTransport` class as can be seen [here](https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/transport.py).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,7 @@ v9.9.9 (unreleased)
 breaking:
 
 - backend/slack: remove slack and slack_rtm built-in backends (#1581)
+- core/logging: deprecate SENTRY_TRANSPORT config (#1604)
 
 features:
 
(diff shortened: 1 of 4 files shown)


sijis:
@browniebroke You'll have to rebase. Also, could you add a note in `CHANGES.rst`?

browniebroke:
Sure, will do!

browniebroke:
All done I think.

PR #1608, opened 2022-10-10
fix: docker build use local tree

sijis:
refactor docker build
* use new errbot-backend-slackv3 package
* ensure build uses local tree

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,17 +2,16 @@ FROM python:3.9-slim as build
 WORKDIR /wheel
 COPY . .
 RUN apt update && apt install -y build-essential git
-RUN pip3 wheel --wheel-dir=/wheel . \
-    wheel \
-    errbot errbot[irc] errbot[XMPP] errbot[telegram] errbot-backend-slackv3
+RUN pip3 wheel --wheel-dir=/wheel \
+    wheel . .[irc] .[XMPP] .[telegram] errbot-backend-slackv3
 
 FROM python:3.9-slim as base
 COPY --from=build /wheel /wheel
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -2,21 +2,16 @@ FROM python:3.9-slim as build
...
 FROM python:3.9-slim as base
 COPY --from=build /wheel /wheel
 RUN apt update && \
     apt install -y git && \
     cd /wheel && \
-    pip3 -vv install --no-cache-dir --no-index --find-links /wheel . \
-    -r /wheel/slackv3-requirements.txt \
-    errbot errbot[irc] errbot[XMPP] errbot[telegram] && \
+    pip3 -vv install --no-cache-dir --no-index --find-links /wheel \
+    . .[irc] .[XMPP] .[telegram] errbot-backend-slackv3 && \
```
e4r7hbug, line 14: You probably want to keep the package name `errbot` for the installation. Using the relative `.` makes sense for building the package automatically named by `setup.py`, but wouldn't give `pip` a concrete package name to install.
```suggestion
    errbot errbot[irc] errbot[XMPP] errbot[telegram] errbot-backend-slackv3 && \
```
You can also collapse the list of extras down which might help with legibility.
```suggestion
    errbot[irc,XMPP,telegram] errbot-backend-slackv3 && \
```

sijis: @e4r7hbug You make some valid points, however, the goal here is to create the master releases published to Dockerhub and I'd like to ensure we are just publishing the local tree changes.

I did, however, include your suggestion on simplifying the extras install.

PR #1612, opened 2022-10-24
Update Dockerfile to have correct requirements file name

trungatdoshii:


sijis:
Thank you for your suggestion, however, this should be fixed via https://github.com/errbotio/errbot/pull/1608.

PR #1627, opened 2023-01-22
fix: update module versions and build

sijis:
Update versions to address broken builds
* dulwich - fixing deprecation warning of `HTTPResponse.getheader()`
* pyOpenSSL- fixing error `module 'lib' has no attribute 'OpenSSL_add_all_algorithms'`

Fix build by addressing path changes in `dist-check`


adamnfish:
Any chance of a release with this change included?

Thank you for the fix!

PR #1631, opened 2023-02-10
Log filtered command name

jcfrt:
If a command is blocked by a cmdfilter, the name of the command should be logged instead of being logged as "None".


sijis:
@jcfrt Thank you!

sijis:
As an example, this is what the before and after look like

```
## Before
  INFO:errbot.core:Command None blocked or deferred.
    returned tuple: (None, None, None)

## After
  INFO:errbot.core:Command "restart" blocked or deferred.
    returned tuple: (None, 'restart', ('--force',))
```

PR #1651, opened 2023-08-27
chore: update core dependencies

sijis:
Updating all dependencies.


nzlosh:
Tested installation from `chore/update_all_dependencies` + `errbot --init; errbot --text` for py3.8, py3.9, py3.10 and py3.11.  LGTM :+1:

PR #1655, opened 2023-09-01
chore: update backend dependencies

sijis:
Update backend dependencies for
* irc
* telegram
* XMPP


sijis:
@nzlosh Would you be able to try and spot check that these continue to work?

nzlosh:
OK, I'll take a look.

Issue #1665, opened 2023-10-26
Automate release process


kayman-mk:
As discussed in #1646 it would be helpful to automate the release process of the Errbot to free up some of the maintainers' resources and make the whole process more comfortable and faster.

Let me briefly outline my ideas how to do that:
1. Enforce semantic commits. This enables the release process to automatically set the new release version number. Tool: amannn/action-semantic-pull-request. Enforces the correct title of the PR. The workflow fails in case the title is incorrect.
2. Ensure to have the PR title and description used for the commits in the default branch. Otherwise it might be possible to create a commit which does not follow the semantic commit rules. Tool: GitHub project settings (General > PR).
3. Ensure squashing the PR to not add non semantic commits accidentally. Tool: GitHub settings (General > PR)
1. Use [release-please](https://github.com/googleapis/release-please) to create a release, i.e. to tag the default branch with a new version number. Release Please creates a PR to do a release. So we can simply modify the `errbot/version.py` file and add the changelog. This way multiple changes can be integrated into one release too. So there is no need to release every single commit to PyPi.
2. Automate setting the release version in the `errbot/version.py`. Should be possible within the release workflow.
4. Let Release Please maintain the changelog. This, of course, requires good PR titles/descriptions. Might be enabled by default. Not sure.
5. Add a workflow to automatically release to PyPi. Tool: GitHub workflow.

If you agree I will start working from top to bottom and create the necessary PRs. What do you think?

nzlosh:
@sijis handles errbot releases, so I'd expect he'll be able to confirm with the process he uses.  I'm all for having an automated (or as much is possible) release process.

kayman-mk:
Got everything up an running and released version 7.0.0 including everything up to commit 0080eff1324930207f021f55674d5240abae172d (2023-10-30).

Docker image: https://github.com/Hapag-Lloyd/errbot/pkgs/container/errbot-hl/149130296?tag=7.0.0
PyPi package: https://pypi.org/manage/project/errbot-hl/release/7.0.0/

GitHub project: https://github.com/Hapag-Lloyd/errbot

Release process is 100% automatic now.

PR #1666, opened 2023-10-26
ci: ensure PR titles follow the conventional commit rules

kayman-mk:
Adds workflow to lint the PR title and enforce the conventional commit rules. By enforcing them we can later automatize the generation of release versions.

Solves part of #1665


kayman-mk:
` validateSingleCommit*` can be set to `false` as soon as the commits are squashed and the PR is used as commit message. Not sure if this is already the case.

PR #1673, opened 2023-12-04
migrate from external mock package to stdlib unittest.mock

eli-schwartz:
It was moved to the stdlib in python 3.3, and the PyPI version is a backport for python versions less than 3.3.


nzlosh:
The CHANGELOG should also be updated.

eli-schwartz:
Done.

sijis (approved):

sijis:
@eli-schwartz thank you!

PR #1675, opened 2023-12-18
correctly import Mapping from collections.abc

tgrondier:
Since python 3.10, Mapping is only available through `collections.abc.Mapping` and not through `collections.Mapping`

https://github.com/errbotio/errbot/pull/1562 was proposed including this change but it has been untouched for over a year.


nzlosh (approved):

sijis:
@tgrondier Added note on your PR to `CHANGES.rst`.

Thank you!

sijis (approved):

PR #1698, opened 2024-07-10
Type hints Fix

ZhipengXue97:
## Description
Fix type check warning reported by Pyre@Google, which was outdated after code modification.
## Detail
update the return type of  function `_read_dict()` from `dict` to `Mapping`, since the condition check of the return `new_dict` and type `Mapping` after commit 9193b1fd3d6a629da707fce4344a10606c0404ba


sijis:
@ZhipengXue97 thank you!

PR #1724, opened 2025-04-28
Close and join thread pools between tests

cjwatson:
https://bugs.debian.org/1103066 reported test failures on Debian's i386 architecture.  After some digging I found that tests were hitting `ENOMEM` depending on which other tests had been run previously, suggesting test isolation problems; i386 has a more limited address space than most other architectures so is more likely to run into this sort of thing.  Forcing thread pools to be shut down between tests seems to avoid this problem.


nzlosh (approved):

sijis:
@cjwatson thank you!

Issue #1721, opened 2025-04-23
Status of Python 3.13 support


lbillinghamwrk:
Hello Folks, 

Can you please help me understand the support status for the latest python?

It looks like we depend on `cgi`; apparently via `webob` and then via `webtest` [which is used in production to validate incoming webhooks](https://github.com/errbotio/errbot/blob/master/errbot/core_plugins/webserver.py#L183)
However,  [`cgi` is no longer included in python 3.13](https://docs.python.org/3/deprecations/pending-removal-in-3.13.html).

While do have a test-suite running on 3.13 already [python-package.yml](https://github.com/errbotio/errbot/blob/653656d8dd96e84c3b7974bf2cbec30806bd6cdf/.github/workflows/python-package.yml#L22); I'm not sure that we exercise the full webserving/webob path there.

The [latest PR checks run] (https://github.com/errbotio/errbot/actions/runs/14037314539/job/39298415722#step:5:256) does still show warnings about `cgi`

```sh
 /home/runner/work/errbot/errbot/.tox/py/lib/python3.12/site-packages/webob/compat.py:5: DeprecationWarning: 'cgi' is deprecated and slated for removal in Python 3.13
    from cgi import parse_header
```

Are we confident that CI exercises the  `cgi`-ish code paths? Can we run fully on python 3.13

lbillinghamwrk:
If the webserver doesn't work, the fix may be to depend on  `legacy-cgi; python_version >= '3.13'`

lbillinghamwrk:
Thank you. Do you think it could be worth adding `legacy-cgi` to our dependencies as a stopgap?

I can do that, or change things to flag max version is 3.12.

I wouldn't be capable of doing the refactor/migration in the foreseeable future.

nzlosh:
If `legacy-cgi` get 3.13 support, it seems like a reasonable stopgap until we hit a hard blocker.

sijis:
@lbillinghamwrk if you create a PR to add that dependency for 3.12, I'll merge that in.

Long term, as @nzlosh suggested, a switch to ASGI/WSGI may be the best approach.

lbillinghamwrk:
Thanks @sijis. I have opened https://github.com/errbotio/errbot/pull/1729 to address this. 
I've tested that `pip install .` works on python 3.9 and 3.13.

sijis:
merged to master.

PR #1729, opened 2025-05-27
fix(deps): allow webtest to work on python 3.13

lbillinghamwrk:
Should fix issue [#1721](https://github.com/errbotio/errbot/issues/1721).

`webtest` is used to validate incoming webhooks in the production `Webserver()`.

Via `webtest` and `webob`, we have a transitive dependency on `cgi`.
As of python 3.13 [`cgi` has been dropped from stdlib](https://docs.python.org/3/deprecations/pending-removal-in-3.13.html).

The recommended stopgap is to add an explicit dependency on `legacy-cgi`, 
long term `webob` should be replaced with WSGI or ASGI

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -33,6 +33,7 @@ fixes:
 - docs: update tox chat server URLs (#1739)
 - chore: update Dockerfile python version (#1744)
 - chore: update errbot-backend-slackv3 version to 0.3.2 (#1743)
+- fix: allow webtest to work on python 3.13 (#1729)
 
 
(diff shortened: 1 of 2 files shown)


lbillinghamwrk:
I'm not 100% sure that `setup.py` is compatible with PEP 508 environment markers.
Pending a full test-suite workflow approval, I've tested that `python -m pip install .` works in both python 3.9 and 3.13. 
Only the latter, as expected, has `legacy-cgi` present.

sijis:
@lbillinghamwrk thank you!

PR #1766, opened 2026-06-13
feature: support editable entrypoints

sijis:
This adds support for editable installed entrypoint plugins.

We also updated tests cases and consolidated into a single file.

### DIFF
--- a/errbot/utils.py
+++ b/errbot/utils.py
@@ -1,9 +1,11 @@
 import collections
 import fnmatch
 import importlib.metadata
+import importlib.util
 import inspect
 import logging
 import os
+import pathlib
 import re
 import sys
 import time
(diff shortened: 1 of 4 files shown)


errbot/utils.py
```diff
@@ -200,13 +202,21 @@ def collect_roots(base_paths: List, file_sig: str = "*.plug") -> List:
...
-                    paths.add(str(parent))
+    for ep in importlib.metadata.entry_points(group=group):
+        # 1. Spec-based discovery (Handles editable installs)
+        try:
+            spec = ep.module and importlib.util.find_spec(ep.module)
+            if spec and spec.origin:
+                paths.add(str(pathlib.Path(spec.origin).resolve().parent))
+                continue
+        except Exception:
+            pass
```
nzlosh: Is it possible to tighten the exception range to avoid debugging frustration due to relevant errors being silently drop?  Alternatively add a debug log to hint an error occurred in that part of the code.

sijis: Good point. added debug log.

nzlosh (approved):
LGTM!

